PR #169, opened 2023-07-19
refactor(all): submit transaction to chain as it is received in the handler

technicallyty:
 
- Refactors the game shard transaction storage system to emit signed payloads as they are received in the handler.
- implements a psuedo-tx mempool system that waits for transactions to come in from certain ticks. once a transaction from a new tick has arrived, the previous pool of txs in that tick get sent to an outbox, which gets drained by EndBlocker.
- refactors cardinal to more sophisticatedly handle recovery. before, it would skip ticks inbetween transaction batches, which could rebuild an entirely different game state (i.e. if a game has a health regen system per tick, skipping ticks would mean different health outcome). 

## Brief Changelog

- see above

## Testing and Verifying

unit tests in each affected package

## Documentation and Release Note

- Does this pull request introduce a new feature or user-facing behavior changes? (yes / no)
- Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? (yes / no)
- How is the feature or change documented? (not applicable / specification (`x/<module>/spec/`) / [Osmosis docs repo](https://github.com/osmosis-labs/docs) / not documented)


cardinal/chain/adapter.go
```diff
@@ -17,8 +18,18 @@ import (
...
+}
+
+type Writer interface {
+	Submit(ctx context.Context, p *sign.SignedPayload, txID, tick uint64) error
+}
+
+type Reader interface {
+	QueryTransactions(
+		context.Context,
+		*shardtypes.QueryTransactionsRequest) (*shardtypes.QueryTransactionsResponse, error)
```
namrapatel, line 32: I trust that this is correct, but wondering: why have we chosen to use pointers in these vars?

cardinal/server/server.go
```diff
@@ -198,14 +203,33 @@ func (t *Handler) makeTxHandler(tx transaction.ITransaction) http.HandlerFunc {
...
+			} else {
+				tick := submitTx()
+				err = t.adapter.Submit(context.Background(), sp, uint64(tx.ID()), tick)
+				if err != nil {
+					writeError(writer, "error submitting transaction to blockchain", err)
+				}
+			}
+		} else {
+			submitTx()
 		}
```
namrapatel, line 232: 1. Can you quickly explan why the lack of an adapter in the handler means that the world should be in recovery mode?

2. The code seems to be checking if we have an adapter, and if we do, checks if the world `IsRecovering`, but I thought if we have an adapter, the world is not in recovery mode?

3. In which case should we call `submitTx()`, in recovery mode, or outside of recovery mode?

cardinal/chain/adapter.go
technicallyty: gRPC client/server generated code operates on pointer requests/response types. since this is mostly an alias to the underlying gRPC call, i just kept the params the same. they could be value types instead of pointers, but don't think it matters too much

cardinal/ecs/tests/chain_recover_test.go
```diff
@@ -15,23 +20,57 @@ import (
...
+	bz, err := proto.Marshal(sp)
+	if err != nil {
+		return err
+	}
+	if d.txs[tick] == nil {
+		d.txs[tick] = &types.Transactions{Txs: make([]*types.Transaction, 0)}
+	}
+	d.txs[tick].Txs = append(d.txs[tick].Txs, &types.Transaction{
+		TxId:          txID,
+		SignedPayload: bz,
```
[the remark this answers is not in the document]
technicallyty: keep in mind this is a test, the adapter implementation above is just a dummy one. i store them in the dummy implementation to simulate having them stored on chain

2 - yes, when there's a new tick, the mapping at that tick will return `nil` so we have to instantiate a new `*types.Transactions` first

cardinal/server/server.go
technicallyty: the recovery mechanism reads transactions from the chain, which an adapter is needed to support. the recovery function won't run, and in turn won't set the `IsRecovering` value, unless an adapter is present in the world

if a transaction is submitted during recovery, we should not process it, as we need to get the game state back to where it was before, before we continue ticking

submit should only happen if we're not recovering

technicallyty:
good questions here @namrapatel, i appreciate the thorough review 🕵🏻‍♂️

chain/x/shard/types/genesis.go
```diff
@@ -5,10 +5,5 @@ func DefaultGenesis() *GenesisState {
 }
 
 func (g *GenesisState) Validate() error {
```
namrapatel, line 7: Is this meant to still exist?

cardinal/server/server.go
technicallyty: ah, are you referring to this `// as long as we have an adapter, and the current world is not in recovery mode`? i can see how that is confusing. i can update it

chain/x/shard/types/genesis.go
technicallyty: forgot to add that [feat: add genesis validation](https://github.com/Argus-Labs/world-engine/pull/169/commits/2132db263d68ef9b54e4218620f28422e4710076)

chain/shard/tx_queue.go
```diff
@@ -0,0 +1,93 @@
...
+			txs:       make(map[uint64]*types.SubmitCardinalTxRequest),
+		}
+		tc.ntx[namespace].tickQueue.Enqueue(tick)
+		tc.ntx[namespace].txs[tick] = &types.SubmitCardinalTxRequest{
+			Sender:    tc.moduleAddr,
+			Namespace: namespace,
+			Tick:      tick,
+			Txs:       &types.Transactions{Txs: []*types.Transaction{{TxId: txID, SignedPayload: payload}}},
+		}
+		return
```
jerargus, line 50: I think this function should be reorganized a bit. Remove this return, and limit this if branch to only set up the namespace. Use the tickQueue.Empty() branch and nil checks below to actually add the transactions. This way, the logic for adding stuff is only implemented in one place.

namrapatel (approved):

chain/x/shard/keeper/keeper_test.go
```diff
@@ -41,137 +43,117 @@ func (s *TestSuite) SetupTest() {
...
-	_, err := s.keeper.SubmitBatch(s.ctx, &types.SubmitBatchRequest{
-		Sender:           s.auth,
-		TransactionBatch: batch,
-	})
-	s.Require().NoError(err)
+	gen := s.keeper.ExportGenesis(s.ctx)
+	// there should only be 2 namespaced txs, because we only submitted 2 diff ones.
+	s.Require().Len(gen.Txs, 2)
+	s.Require().Len(gen.Txs[1].Txs, 2)            // we submitted 2 ticks for namespace "foo"
+	s.Require().Len(gen.Txs[1].Txs[0].Txs.Txs, 3) // the first tick had 3 txs
```
jerargus, line 147: Holy cow, this is confusing. I think these "Txs" should be renamed to better reflect what they actually are. e.g.: `gen.Namespace[1].Tick[0].Txs`. Can the second field in the TickedTransaction struct be a `repeated Transaction` instead of a struct that only contains the repeated transaction?

cardinal/server/server.go
namrapatel: yeah the comment seemed contradictory

jerargus (changes requested):

cardinal/ecs/tests/chain_recover_test.go
```diff
@@ -40,104 +79,68 @@ type SendEnergyTransaction struct {
...
+	sysRuns := 0
+	timesSendEnergyRan := 0
+	// send energy system
+	w.AddSystem(func(world *ecs.World, queue *ecs.TransactionQueue) error {
+		sysRuns++
 		txs := SendEnergyTx.In(queue)
-		for i, tx := range txs {
-			assert.DeepEqual(t, tx, sendEnergyTxs[i])
+		if len(txs) > 0 {
+			timesSendEnergyRan++
```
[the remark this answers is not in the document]
technicallyty: i just wanted to test how many times the system itself ran, not how many txs were run in the system

cardinal/server/server.go
```diff
@@ -198,14 +203,33 @@ func (t *Handler) makeTxHandler(tx transaction.ITransaction) http.HandlerFunc {
...
+				writeError(writer, "unable to marshal response", err)
+				return 0
+			}
+			writeResult(writer, res)
+			return uint64(tick)
+		}
+
+		// as long as we have an adapter, and the current world is not in recovery mode
+		if t.adapter != nil {
+			if t.w.IsRecovering() {
```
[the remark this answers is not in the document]
technicallyty: yeah im not entirely sure the correct flow for dealing with recoveries. it should ideally happen before the handler is even set up. 

and handlers are going to be internally setup soon, so maybe we can wait until then to decide how to deal with this concern?

cardinal/chain/adapter.go
```diff
@@ -17,8 +18,18 @@ import (
 
 // Adapter is a type that helps facilitate communication with the EVM base shard.
 type Adapter interface {
```
smsunarto, line 20: nit: can we call this ShardAdapter

cardinal/chain/adapter.go
```diff
@@ -17,8 +18,18 @@ import (
...
-	QueryBatches(ctx context.Context, req *shardtypes.QueryBatchesRequest) (*shardtypes.QueryBatchesResponse, error)
+	Writer
+	Reader
+}
+
+type Writer interface {
+	Submit(ctx context.Context, p *sign.SignedPayload, txID, tick uint64) error
+}
+
+type Reader interface {
```
smsunarto, line 29: nit: can we call this ShardAdapterReader

cardinal/chain/adapter.go
```diff
@@ -17,8 +18,18 @@ import (
...
 // Adapter is a type that helps facilitate communication with the EVM base shard.
 type Adapter interface {
-	Submit(ctx context.Context, namespace string, tick uint64, txs []byte) error
-	QueryBatches(ctx context.Context, req *shardtypes.QueryBatchesRequest) (*shardtypes.QueryBatchesResponse, error)
+	Writer
+	Reader
+}
+
+type Writer interface {
+	Submit(ctx context.Context, p *sign.SignedPayload, txID, tick uint64) error
```
smsunarto, line 26: can we comment on what this is needed for/does

cardinal/chain/adapter.go
```diff
@@ -17,8 +18,18 @@ import (
...
+}
+
+type Reader interface {
+	QueryTransactions(
```
smsunarto, line 30: can we comment on what this is needed for/does

cardinal/chain/adapter.go
```diff
@@ -17,8 +18,18 @@ import (
...
+		context.Context,
+		*shardtypes.QueryTransactionsRequest) (*shardtypes.QueryTransactionsResponse, error)
 }
 
 type Config struct {
```
smsunarto, line 35: nit: can we call this ShardAdapterConfig

chain/x/shard/keeper/keeper_test.go
technicallyty: okay simplified this a bunch - the storage of the transactions were wrapped so i could store them like this:

`store(namespace) -> key = tick number : value = transactions`

wrapping the transactions in their own proto message allows you to proto.Marshal them entirely self contained (i can't proto.Marshal a slice of proto messages). but, i don't think this really gives us any real benefit, and ended up just being confusing.

i switched it to just marshal the tick. this does add redundancy, so now you have 

`store(namespace) -> key = tick number : value = types.Tick(tick, []txs)`

so tick number gets stored twice, but i dont think it should be much of a problem

cardinal/ecs/tests/chain_recover_test.go
```diff
@@ -15,23 +20,57 @@ import (
...
+	}
+	sort.Slice(tickedTxs, func(i, j int) bool {
+		return tickedTxs[i].Tick < tickedTxs[j].Tick
+	})
+	var pr *types.PageResponse
+	if d.calls == 0 {
+		// return the first half
+		tickedTxs = tickedTxs[0 : len(tickedTxs)/2]
+		d.calls++
+		pr = &types.PageResponse{Key: []byte("this doesnt matter")}
```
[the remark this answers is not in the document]
technicallyty: updated

cardinal/evm/server.go
smsunarto, line 1: would it make sense to just have this as a part of `chain`? it seems to be doing similar things (i.e. communicating with base shards)

chain/shard/tx_queue.go
technicallyty: updated, lmk how it looks now

chain/x/shard/keeper/tx_storage.go
```diff
@@ -0,0 +1,67 @@
+package keeper
+
+import (
+	"cosmossdk.io/store/prefix"
```
smsunarto, line 4: why is this not github.com :/

chain/x/shard/keeper/msg_server.go
```diff
@@ -11,15 +10,18 @@ import (
 
 var _ types.MsgServer = &Keeper{}
 
-func (k *Keeper) SubmitBatch(ctx context.Context, msg *types.SubmitBatchRequest) (*types.SubmitBatchResponse, error) {
+func (k *Keeper) SubmitCardinalTx(ctx context.Context, msg *types.SubmitCardinalTxRequest,
```
smsunarto, line 13: nit: SubmitShardTx. We want this to be generalizable to other shards that are not cardinal.

chain/x/shard/keeper/query_server.go
```diff
@@ -2,39 +2,40 @@ package keeper
...
+	k.iterateTransactions(sdk.UnwrapSDKContext(ctx), key, nil,
+		req.Namespace, func(tick uint64, txs *types.Transactions) bool {
+			// we keep the check here so that if we hit the limit,
+			// we return the NEXT key in the iteration, not the one before it.
+			if count == limit {
+				res.Page.Key = k.getTransactionKey(tick)
+				return false
+			}
+			res.Txs = append(res.Txs, &types.TickedTransactions{
+				Tick: tick,
```
smsunarto: i would generalize this to "epoch". in the future, there might be other shards that doesn't have the concept of tick.

cardinal/chain/adapter.go
technicallyty: if we want to go that direction of naming, i would suggest to rather rename the `chain` package to `shard`

usage then looks like:
```go
import "github.com/argus-labs/world-engine/cardinal/shard"

func Foo() {
    a := shard.NewAdapter()
    ...
}
```

keeps things slimmed down and achieves the same goal

cardinal/chain/adapter.go
```diff
@@ -17,8 +18,18 @@ import (
 
 // Adapter is a type that helps facilitate communication with the EVM base shard.
 type Adapter interface {
-	Submit(ctx context.Context, namespace string, tick uint64, txs []byte) error
-	QueryBatches(ctx context.Context, req *shardtypes.QueryBatchesRequest) (*shardtypes.QueryBatchesResponse, error)
+	Writer
+	Reader
+}
+
+type Writer interface {
```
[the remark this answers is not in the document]
technicallyty: if we go with renaming the package to `shard` i think we can forego the shard prefix there and possibly just do `WriteAdapter`

cardinal/chain/adapter.go
technicallyty: if we go with renaming the package to `shard` i think we can forego the shard prefix there and possibly just do `ReadAdapter`

cardinal/evm/server.go
technicallyty: hmm.. maybe? `evm` package is meant to deal with receiving and executing transactions that arrive from evm/smart contracts. where as the `chain` or `shard` (whichever we end up naming it) deals with processing internal transactions and storage. 

although this would be a bit out of scope for this PR, so if we have compelling reasons to merge the package, id prefer to have a linear ticket opened for it

chain/x/shard/keeper/tx_storage.go
technicallyty: much of the sdk is moving to aesthetic package names https://cosmossdk.io/

chain/x/shard/keeper/query_server.go
technicallyty: [refactor: tick -> epoch](https://github.com/Argus-Labs/world-engine/pull/169/commits/4c17cdd25f6260b3f3d5752495fbc08a4de76392) done

chain/x/shard/keeper/msg_server.go
technicallyty: done

cardinal/chain/adapter.go
technicallyty: leaning towards AdapterConfig

cardinal/chain/adapter.go
technicallyty: done

cardinal/chain/adapter.go
technicallyty: done

cardinal/chain/adapter.go
smsunarto: sounds good

cardinal/chain/adapter.go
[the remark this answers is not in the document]
smsunarto: sounds good

chain/proto/shard/v1/genesis.proto
```diff
@@ -15,6 +15,6 @@ message NamespaceTransactions {
   // namespace is the namespace the transactions occurred in.
   string namespace = 1;
 
-  // txs contains a tick number, and the transactions that occurred within that tick.
-  repeated Tick ticks = 2;
+  // epochs contains an epoch number, and the transactions that occurred within that epoch.
+  repeated Epoch epochs = 2;
```
jerargus, line 19: Oh nice.

cardinal/chain/adapter.go
smsunarto: sounds good

cardinal/evm/server.go
smsunarto: hm, yeah i think this should become a part of the `shard` package. i agree we should do this in a separate PR

chain/x/shard/keeper/tx_storage.go
smsunarto: not very aesthetic sad!

smsunarto (approved):

chain/shard/tx_queue.go
jerargus: 👍

chain/x/shard/keeper/keeper_test.go
jerargus: 👍

jerargus (approved):

PR #186, opened 2023-08-01
feat(chain)!: game shard query precompile

technicallyty:

adds a new precompile method to `router` called `query`, which allows EVM smart contracts to make query's against Cardinal.

## Brief Changelog

- updates `Read` interface with methods that add EVM support
- updates `NewReadType` function - adds parameter `supportEVM bool` which, when true, will dynamically generate the ABI equivalent types of the request and reply types.
- adds `query` method to `router` precompile contract
- extends `Router` proto to have `query` rpc method
- implements query method client and server implementations

## Testing and Verifying
pending e2e tests
`cardinal/ecs/tests/read_test.go`


cardinal/ecs/abi.go
```diff
@@ -0,0 +1,81 @@
...
+	"uint":           "uint64",
+	"int":            "int64",
+	"common.Address": "address",
+	"uint8":          "uint8",
+	"uint16":         "uint16",
+	"uint32":         "uint32",
+	"uint64":         "uint64",
+}
+
+func GenerateABIType(goStruct any) (*abi.Type, error) {
```
jerargus, line 23: This seems like a good candidate for a table driven test.

jerargus (changes requested):

cardinal/ecs/abi.go
technicallyty: i added more test coverage, but seemed more straightforward to just have a massive struct with all supported types to make sure we can pack & unpack, and then individual tests for edge cases. 

if you still think its valuable to have these be table driven, lmk, i can look into this more

jerargus (approved):

cardinal/evm/server.go
```diff
@@ -107,6 +110,26 @@ func (s *srv) SendMsg(ctx context.Context, msg *routerv1.MsgSend) (*routerv1.Msg
 		return nil, err
 	}
 	// add transaction to the world queue
-	s.txh.AddTransaction(itx.ID(), tx, nil)
+	s.world.AddTransaction(itx.ID(), tx, nil)
 	return &routerv1.MsgSendResponse{}, nil
 }
+
+func (s *srv) QueryShard(ctx context.Context, req *routerv1.QueryShardRequest) (*routerv1.QueryShardResponse, error) {
```
smsunarto, line 117: Im cautious of using the word "query" here since it means something in ECS. Maybe we should rename this to read for consistency?

chain/precompile/router/router.go
```diff
@@ -39,3 +39,11 @@ func (c *Contract) Send(
 	_, err := c.rtr.Send(ctx, namespace, pCtx.MsgSender().String(), messageID, message)
 	return nil, err
 }
+
+func (c *Contract) Query(
```
smsunarto, line 43: Read?

chain/precompile/router/router.go
```diff
@@ -39,3 +39,11 @@ func (c *Contract) Send(
 	_, err := c.rtr.Send(ctx, namespace, pCtx.MsgSender().String(), messageID, message)
 	return nil, err
 }
```
smsunarto, line 41: We should probably make it clearer here `sendMessage` sounds about right

chain/contracts/src/cosmos/precompile/router.sol
```diff
@@ -1,5 +1,10 @@
 pragma solidity ^0.8.4;
 
 interface IRouter {
-    function send(bytes calldata message, uint64 messageID, string calldata namespace) external returns (bytes memory);
+    function send(bytes calldata message, uint64 messageID, string calldata namespace)
+        external
+        returns (bytes memory);
+    function query(bytes calldata request, string calldata resource, string calldata namespace)
```
namrapatel, line 7: if we change from query -> read prob also gotta change here

cardinal/ecs/read.go
```diff
@@ -57,3 +118,73 @@ func (r *ReadType[req, rep]) HandleRead(w *World, bz []byte) ([]byte, error) {
 	}
 	return bz, nil
 }
+
+func (r *ReadType[req, rep]) DecodeEVMRequest(bz []byte) (any, error) {
+	if r.requestABI == nil {
+		return nil, errors.New("cannot call DecodeEVMRequest without using the WithEVMSupport option when " +
+			"creating the read")
```
namrapatel, line 125: just wondering: are we using this + in the error for readability?

chain/precompile/router/router.go
technicallyty: not entirely sure whats being said here

cardinal/ecs/read.go
technicallyty: yeah, linter has line length rules

cardinal/evm/server.go
technicallyty: yeah, i was veering either way, cause `Read` is quite specific naming to Cardinal, and the precompile is eventually generalized to many different game shards, where i imagine query will be the default word for these types of things. 

the more i think about this, the more i want to rename our read system to query honestly lol

cardinal/evm/server.go
```diff
@@ -21,33 +20,37 @@ var (
...
-
-// TxHandler is a type that gives access to transaction data in the ecs.World, as well as access to queue transactions.
-type TxHandler interface {
-	AddTransaction(transaction.TypeID, any, *sign.SignedPayload) uint64
-	ListTransactions() ([]transaction.ITransaction, error)
-}
+// IReadTypes maps read resource names to the underlying IRead
+type IReadTypes map[string]ecs.IRead
 
 type srv struct {
```
namrapatel, line 29: not to be a troll, but feels like someone reading this code would be confused by what this is for cuz of the name? feel free to resolve/ignore if you think it doesn't matter much or there isn't a better name

technicallyty: glad you called this out, cause these types shouldn't be exported either, since they're only used within this package. 

I've renamed them to describe what they're actually doing, i.e. mapping some identifier to some type - txById and readByName

chain/precompile/router/router.go
smsunarto: yeah

cardinal/evm/server.go
smsunarto: grrr, yeah i guess query makes more sense

cardinal/evm/server.go
technicallyty: re: `srv` updated to be `msgServerImpl`

namrapatel (approved):

chain/precompile/router/router.go
technicallyty: updooted [refactor: send -> SendMessage](https://github.com/Argus-Labs/world-engine/pull/186/commits/a2c7e08701ffed26bbfec8453262e40655a3b9df)

PR #187, opened 2023-08-01
feat: abstract game loop into world.StartTicks()

namrapatel:
Fixes: CAR-109

Cardinal users currently have to manually spin out a go routine that runs the game loop like so:

```go
go func() {
  	for range time.Tick(time.Second) {
		if err := world.Tick(context.Background()); err != nil {
			panic(err)
		}
	}
}
```

We should abstract this away and let the user pass in their preferred tick time, to something like this:

```go
world.StartTicks(ctx, time.Millisecond * 100)
```

## Brief Changelog

- _Add a function called `StartTicks(ctx, loopInterval)` in `world.go` that would be called in the user's `main.go`_

## Testing and Verifying

### DIFF
--- a/cardinal/ecs/world.go
+++ b/cardinal/ecs/world.go
@@ -7,9 +7,11 @@ import (
 	"fmt"
 	"sort"
 	"sync"
+	"time"
 
 	"github.com/argus-labs/world-engine/chain/x/shard/types"
 	"github.com/argus-labs/world-engine/sign"
+	"github.com/rs/zerolog/log"
 
 	"github.com/argus-labs/world-engine/cardinal/chain"
 	"github.com/argus-labs/world-engine/cardinal/ecs/component"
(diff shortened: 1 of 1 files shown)


cardinal/ecs/world.go
```diff
@@ -459,6 +461,17 @@ func (w *World) Tick(ctx context.Context) error {
 	return nil
 }
 
+func (w *World) StartTicks(ctx context.Context, loopInterval time.Duration) {
```
smsunarto, line 464: nit: i would call this `StartGameLoop`

cardinal/ecs/world.go
```diff
@@ -459,6 +461,17 @@ func (w *World) Tick(ctx context.Context) error {
...
+	log.Info().Msg("Game loop started")
+	go func() {
+		for range time.Tick(loopInterval) {
+			if err := w.Tick(ctx); err != nil {
+				panic(err)
```
smsunarto, line 469: We should zerolog this as well

namrapatel: sg, lmk if the new change is fine

cardinal/ecs/world.go
namrapatel: done

technicallyty (approved):
bigly

PR #192, opened 2023-08-03
DO NOT SUBMIT: Transaction receipt example

jerargus:
This test shows an example of how transaction receipts/results/errors will be used in ECS. 

Some API changes to note:
- Each transaction will now be associated with 2 types: an input and an output.
- The "In" function (used in a System to get a list of transactions of a particular type) will be replaced with "TxsIn" which will return a list of transactions along with transaction IDs, signatures, and the actual transaction data. Alternatively, the "In"'s return type can be changed to be this triplet of information.
- TxID is added to the ecs package. It contains a unique identifier for a specific transaction. It's implemented here as a tuple of Tick, ID (the transaciton.TypeID), and Index, with "Index" being the offset into the world.txQueue[transaction.TypeID] map.

### DIFF
--- a/cardinal/ecs/tests/transaction_test.go
+++ b/cardinal/ecs/tests/transaction_test.go
@@ -2,10 +2,12 @@ package tests
 
 import (
 	"context"
-	"github.com/ethereum/go-ethereum/accounts/abi"
+	"errors"
 	"reflect"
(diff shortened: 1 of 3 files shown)


technicallyty:
can this be closed?

PR #198, opened 2023-08-09
feat(cardinal): Implement an endpoint to read transaction receipts fr… CAR-130 CAR-135

jerargus:
…om cardinal; remove leading slahes from cardinal http endpoints

Closes: #CAR-135, #CAR-130

- Add an HTTP endpoint that will return a list of transactions that have been processed over a range of ticks. 
- Update the HTTP endpoints to NOT have a leading slash.
- Code refactor: read and tx handler definitions have been moved to the handler.go file. 
- Some functions have been moved to the util.go file.
- NOTE: handler.go's code has mainly been moved from server.go
- NOTE: receipt.go's code is new

## Testing and Verifying

- Unit tests added to verify ranges of ticks are returned correctly
- Uint tests added to verify transaction errors and results are returned from the receipt endpoint
- Calling http://localhost:3333/list/tx-endpoints should now list tx endpoints without leading slash
- Calling http://localhost:3333/list/read-endpoints should now list read endpoints without leading slash

### DIFF
--- a/cardinal/server/server.go
+++ b/cardinal/server/server.go
@@ -30,7 +30,9 @@ var (
 const (
 	listTxEndpoint   = "list/tx-endpoints"
 	listReadEndpoint = "list/read-endpoints"
-	listTxReceipts   = "list/tx-receipts"
+
+	// Don't name this tx-receipts to ensure it doesn't match the prefix for normal transactions
+	txReceiptsEndpoint = "transaction-receipts"
 
 	schemaEndpointPrefix = "schema/"
 
--- a/cardinal/server/server.go
+++ b/cardinal/server/server.go
@@ -1,20 +1,13 @@
 package server
 
 import (
-	"context"
-	"encoding/json"
 	"errors"
 	"fmt"
-	"io"
 	"log"
 	"net/http"
 
-	"github.com/argus-labs/world-engine/cardinal/shard"
-	"github.com/invopop/jsonschema"
-
 	"github.com/argus-labs/world-engine/cardinal/ecs"
-	"github.com/argus-labs/world-engine/cardinal/ecs/transaction"
-	"github.com/argus-labs/world-engine/sign"
+	"github.com/argus-labs/world-engine/cardinal/shard"
 )
 
 // Handler is a type that contains endpoints for transactions and queries in a given ecs world.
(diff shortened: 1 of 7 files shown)


technicallyty (approved):

cardinal/server/server.go
```diff
@@ -35,10 +28,11 @@ var (
 )
 
 const (
-	listTxEndpoint   = "/list/tx-endpoints"
-	listReadEndpoint = "/list/read-endpoints"
+	listTxEndpoint   = "list/tx-endpoints"
+	listReadEndpoint = "list/read-endpoints"
+	listTxReceipts   = "list/tx-receipts"
```
smsunarto, line 33: hm, i think we shouldn't use `list/tx-receipt`. we should be consistent of using `list/` prefix only to list endpoints with a certain category. 

i think it would be great to consolidate the endpoints convention to something like this

`list/tx` -> list tx endpoints
`list/read` -> list read endpoints
`/tx/module/abc` -> a tx endpoint (i.e. `/tx/persona/abc`)
`/read/module/abc` -> a read endpoint (i.e. `/read/receipt/abc`)

jerargus: I renamed the tx receipts endpoint so it isn't prefixed with list, nor is it prefixed with "tx-" (making ti look like a normal, user-defined transaction). 

I don't know about having the module as part of the path. I agree in principle, however I think getting the module name is going to be messy.

smsunarto: i dont mean `module` as go module. it's basically just arbitrarily set categories

jerargus: Ah, in that case, there's nothing stopping a user from making whatever groupings they want. The only piece we need to do would be to replace tx- with tx/. Are you suggesting that for this PR?

smsunarto: what i want to have

`tx/persona/abc` -> all persona tx is like this
`tx/game/abc` -> all tx that is defined by game implementation goes here

same with read

`read/persona/abc` -> all persona read is like this
`read/receipt/abc` -> receipts read
`read/game/abc` -> all read that is defined by game implementation goes here

i want it to be very explicit from the endpoint on what functionality/feature/category/etc this `tx` or `read` belongs to

PR #201, opened 2023-08-10
feat(cardinal): AddSystems function

technicallyty:

we have variadic functions for all our other register type methods, so adding one for systems.

## Brief Changelog

- adds variadic AddSystems function
- changes the original AddSystem function to just call into AddSystems.

## Testing and Verifying
`cardinal/ecs/tests/world_test.go`

### DIFF
--- a/cardinal/ecs/tests/world_test.go
+++ b/cardinal/ecs/tests/world_test.go
@@ -1,0 +1,27 @@
+package tests
+
+import (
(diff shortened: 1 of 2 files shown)


technicallyty:
theres a lot of changes that need to be made to do the mega function. im not going to do it in this PR.

jerargus (approved):

PR #205, opened 2023-08-16
feat(cardinal): return a transaction id and the current tick when sub…

jerargus:
…mitting transactoins

- Improve the transaction receipt system by returning a receipt hash to the client as well as the tick that the transaction was added. This will give Nakama the ability to match up transaction receipts with the clients that originally sent them.
- Remove the CreatePersonaResponse struct. Instead, a TransactionReply will be returned (just like every other transaction).

## Testing and Verifying

- Added unit test that validates receipt IDs are returned for normal transaction endpoint was well as the slightly custom create-persona endpoint.


cardinal/server/handler.go
```diff
@@ -116,9 +116,12 @@ func (t *Handler) makeTxHandler(tx transaction.ITransaction) http.HandlerFunc {
 		}
 
 		submitTx := func() uint64 {
-			tick, _ := t.w.AddTransaction(tx.ID(), txVal, sp)
+			tick, id := t.w.AddTransaction(tx.ID(), txVal, sp)
 
-			res, err := json.Marshal("ok")
+			res, err := json.Marshal(ReceiptID{
+				ID:   id.String(),
```
smsunarto, line 122: Can we make the transaction hash to be keccak256(signedPayload) instead?

smsunarto: https://ethereum.stackexchange.com/questions/45648/how-to-calculate-the-assigned-txhash-of-a-transaction

smsunarto: I would prefer to not use tick as a part of the receipt ID here because that can't be calculated in advance in Nakama (you don't know which tick your transaction is going to be executed at)

cardinal/ecs/transaction/transaction.go
```diff
@@ -18,6 +20,10 @@ type TxID struct {
 	Index      uint64
 }
 
+func (id TxID) String() string {
+	return fmt.Sprintf("%s-%d", id.PersonaTag, id.Index)
```
smsunarto, line 24: What happens if the transaction doesn't involve the persona tag of the sender? i.e. register persona transaction

jerargus: PersonaTag should always be defined (https://github.com/Argus-Labs/world-engine/blob/main/cardinal/server/utils.go#L98-L100).

That being said, two different clients (say Nakama and some random smart contract) could both request the same persona tag with (luckily) the same nonce, and generate identical IDs. So maybe <personatag>-<nonce> isn't going to work.

cardinal/server/handler.go
jerargus: I included the tick because it's an input for reading the signer of a persona tag. Returning the tick would make integrating this change into Nakama a little easier.

However, I think with the addition of transaction receipts we should be able to get by without returning the tick.

RE the keccak256, I'm not familiar with that. Would this be the geth implementation: https://pkg.go.dev/github.com/ethereum/go-ethereum/crypto#Keccak256Hash ?

smsunarto: Yeap, that's the geth implementation.

We should return the tick as a part of the receipt but not as a part of the id/hash.

smsunarto: To be clear, what I'm advocating for here is that we remove the idea of a "Receipt ID" there should only be a "transaction hash" and the receipt should attach the transaction hash.

cardinal/ecs/transaction/transaction.go
smsunarto: What is the persona tag used when I call register persona?

jerargus: A signed payload is:
```
type SignedPayload struct {
	PersonaTag string
	Namespace  string
	Nonce      uint64
	Signature  string          // hex encoded string
	Body       json.RawMessage // json string
}
```
Which includes a PersonaTag. The linked code above verified that the PersonaTag is not empty. 

Digging more into the code, it looks like we never actually verify that the PersonaTag in the signature matches the PersonaTag in the message body. 

That's wrong (i'll fix in this PR), but still, it shouldn't be empty.

jerargus: This PR doesn't include the refactoring of SignedPayload to RawTransaction/SignedTransaction: https://linear.app/arguslabs/issue/CAR-159/refactor-signsignedpayload-into-rawtransaction-and-signedtransaction

sign/sign.go
```diff
@@ -16,13 +16,19 @@ import (
 var (
 	// ErrorSignatureValidationFailed is returned when a signature is not valid.
 	ErrorSignatureValidationFailed = errors.New("signature validation failed")
+	ErrorCannotSignEmptyBody       = errors.New("cannot sign empty body")
+	ErrorInvalidPersonaTag         = errors.New("invalid persona tag")
+	ErrorInvalidNamespace          = errors.New("invalid namespace")
 )
 
+const SystemPersonaTag = "SystemPersonaTag"
```
technicallyty, line 24: should comment some stuff about how this is a reserved tag for internal/sys txs

jerargus: done

cardinal/server/utils.go
```diff
@@ -84,7 +89,7 @@ func getSignerAddressFromPayload(sp sign.SignedPayload) (string, error) {
 	return createPersonaTx.SignerAddress, nil
 }
 
-func (t *Handler) verifySignature(request *http.Request, getSignedAddressFromWorld bool) (payload []byte, sig *sign.SignedPayload, err error) {
+func (t *Handler) verifySignature(request *http.Request, requireSystemTransaction bool) (payload []byte, sig *sign.SignedPayload, err error) {
```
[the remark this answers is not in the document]
jerargus: done

smsunarto (approved):
lgtm ship it

PR #206, opened 2023-08-16
fix: update miniredis version

pyrofolium:
Closes: CAR-9

miniredis had a bug. The code involved a work around that would get around it in unit tests. Miniredis is now fixed and the workaround it no longer needed. 

Removes custom copying code and utilizes the client version of the copy directly. 

## Brief Changelog

-removed workaround custom copy function.
-replaced all function calls to the custom copy function with direct client calls to redis copy. 
-added unit test to test redis client copy for this bug. 

## Testing and Verifying

Added unit test to catch miniredis bug, in case miniredis is ever rolled back.


namrapatel (commented):
@jerargus can you confirm that this fix does not solve [this ticket ](https://linear.app/arguslabs/issue/CAR-152/fix-makesnapshot-function-slowness)as well?

jerargus (approved):

PR #207, opened 2023-08-16
refactor: remove hoststr default port

pyrofolium:
Closes: #CAR-111

(Note some commits related to redis are in the history related to another PR, those have been reverted.)

Change function signature of Serve in server.go to remove host and port

Changes the function signature of Serve to be simplified. However this involves the Serve to default to a specific port or there will be an error. I used 4040 because it's what was used in the unit test. 

## Brief Changelog

- removes entry for host as it's not needed at all, go libraries have a default
- removes port as a parameter and sets it as a property of handler. 
- If port is never set, Server runs with a default of 4040
- Added WithPort("some port string") as an "Option" for the test factory. Users can use this to set a port if they don't want the default. Though this pattern is a little complex, tbh. We can discuss. 
- Changed unit tests to work with new changes. 

## Testing and Verifying

Adjusted unit tests to account for new changes. 
Ran tests to see it still works.


namrapatel (commented):
@jerargus Should we make the change in the submoduled starter-game-template too? Not sure if thats the practice we're going for now

jerargus:
> @jerargus Should we make the change in the submoduled starter-game-template too? Not sure if thats the practice we're going for now

Once this is merged, the next time the starter-game-template increases its cardinal version, it will get this code and the call to txh.Serve() will fail. So whoever ends up doing that version increase will have to fix it (i suspect it will be me in the next few days). There's nothing to do right away.

jerargus (approved):

smsunarto (changes requested):
I think we should allow changing the port using `CARDINAL_PORT` environment variable.

namrapatel (approved):

cardinal/server/server_test.go
```diff
@@ -61,6 +62,33 @@ func makeTestTransactionHandler(t *testing.T, world *ecs.World, opts ...Option)
 	}
 }
 
+func TestIfServeSetEnvVarForPort(t *testing.T) {
+	world := inmem.NewECSWorldForTest(t)
+	alphaTx := ecs.NewTransactionType[SendEnergyTx, SendEnergyTxResult]("alpha")
+	betaTx := ecs.NewTransactionType[SendEnergyTx, SendEnergyTxResult]("beta")
+	gammaTx := ecs.NewTransactionType[SendEnergyTx, SendEnergyTxResult]("gamma")
+	assert.NilError(t, world.RegisterTransactions(alphaTx, betaTx, gammaTx))
```
technicallyty, line 70:
```suggestion
	assert.NilError(t, world.RegisterTransactions(alphaTx))
```

will be fine with one tx registered

cardinal/server/server.go
```diff
@@ -72,14 +74,26 @@ func NewHandler(w *ecs.World, opts ...Option) (*Handler, error) {
...
-		Addr:    fmt.Sprintf("%s:%s", host, port),
+		Addr:    fmt.Sprintf(":%s", t.port),
 		Handler: t.mux,
 	}
+}
+
+// Serve sets up the endpoints passed in by the user, as well as a special "/list" endpoint, that informs consumers
+// what endpoints the user set up in the Handler. Then, it serves the application, blocking the main thread.
+// Please us `go txh.Serve(host,port)` if you do not want to block execution after calling this function.
+func (t *Handler) Serve() {
```
technicallyty, line 95:
```suggestion

// Serve sets up the endpoints passed in by the user, as well as a special "/list" endpoint, that informs consumers
// what endpoints the user set up in the Handler. Then, it serves the application, blocking the main thread.
// Please us `go txh.Serve()` if you do not want to block execution after calling this function.
func (t *Handler) Serve() {
```

technicallyty (approved):
lgtm, just a few minor things

pyrofolium:
> I think we should allow changing the port using `CARDINAL_PORT` environment variable.

@smsunarto made those changes. Let me know if it works.

smsunarto (approved):
lgtm

PR #210, opened 2023-08-17
fix(cardinal): update miniredis version

pyrofolium:
A merge conflict in go.mod reintroduced the old error. My mistake. You will see the failure on main. This was the original PR: https://github.com/Argus-Labs/world-engine/pull/206

Everything should be the same as that old PR. 

This was the change that got reintroduced by the merge conflict, I reverted it back to .5: 

https://github.com/Argus-Labs/world-engine/commit/fb69fbb3d20885d8083f92ba1c0b0e6417333110

I think?? there's some loop hole that allowed this merge into mainline to go through even though the tests weren't passing? When code has been approved and previously passed tests and you sync the code back to master, that sync can go through without the tests passing.


technicallyty (approved):
LGTM. can u add the scope the PR title? i.e. `fix(cardinal): update miniredis version`

PR #213, opened 2023-08-18
feat(Cardinal): Brian/car-160/car-149/improve logging

pyrofolium:
Closes: #CAR-160, this is a sub issue of CAR-149 and partially completes it 

Adds the following methods to world:
- LogEntity - outputs entity info when given an entity id
- LogWorldState - outputs info on total components and systems in the world.

Changes splits the definition of system into two components:

```
type RegisteredSystem func(*World, *TransactionQueue) error
type System func(*World, *TransactionQueue, *zerolog.Logger) error
```

System is what the user deals with when defining a system themselves. 
World then curries System and injects a custom logger into System creating RegisteredSystem then places it into the internal system array. The rest of the framework only interacts with RegisteredSystem. 

More methods can/will be added in various other sub-issues created under car-160. 

Also this is a very opinionated take on a logging module. Feel free to suggest different ways of doing it here. 

## Testing and Verifying
See world_logger_test.go. Much of the functionality of how this will be used and what the output looks like is demonstrated in the testing there. 

Feel free to request any additional output/changes to be added to this PR. Or if you have some big ones to add I can create another sub-issue.


cardinal/ecs/world_logger.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+func NewWorldLogger(logger *zerolog.Logger, world *World) WorldLogger {
+	return WorldLogger{
+		logger: logger,
+		world:  world,
+	}
+}
+
+func (wl *WorldLogger) Log(traceId string, logLevel zerolog.Level, message string) {
```
jerargus, line 24: It looks like the message parameter is unused here. That seems important.

cardinal/ecs/world_logger_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+	traceId := "test_trace_id"
+	var buf bytes.Buffer
+	bufLogger := zerolog.New(&buf)
+	//opts is used to hijack the Logger placed in the world object and replace it with something that writes to buf(above)
+	//so it's more easily unit tested. All logging methods called by w.Logger are now written to the buf variable.
+	opt := func(w *ecs.World) {
+		w.Logger = ecs.NewWorldLogger(&bufLogger, w)
+	}
+	w := inmem.NewECSWorldForTest(t, opt)
```
jerargus, line 39: Can you just put `w.Logger = ecs.NewWorldLogger(&bufLogger, w)` here without bothering with the opt parameter?

cardinal/ecs/world_logger.go
```diff
@@ -0,0 +1,110 @@
...
+	zeroLoggerEvent = zeroLoggerEvent.Str("trace_id", traceId)
+	zeroLoggerEvent, err = wl.loadEntityInfoIntoLogger(entityId, zeroLoggerEvent)
+	if err != nil {
+		return err
+	}
+	zeroLoggerEvent.Msg(message)
+	return nil
+}
+
+func (wl *WorldLogger) LogWorldState(traceId string, logLevel zerolog.Level, message string) {
```
jerargus, line 104: it looks like wl.world is only used in this LogWorldState method. How about making it a parameter you pass in, instead of something the WorldLogger object always has a reference to.

cardinal/ecs/world_logger.go
```diff
@@ -0,0 +1,110 @@
...
+
+func (wl *WorldLogger) LogError(traceId string, message string) {
+	wl.Log(traceId, zerolog.ErrorLevel, message)
+}
+
+func loadComponentIntoArrayLogger(component component.IComponentType, arrayLogger *zerolog.Array) *zerolog.Array {
+	dictLogger := zerolog.Dict()
+	dictLogger = dictLogger.Int("component_id", int(component.ID()))
+	dictLogger = dictLogger.Str("component_name", component.Name())
+	arrayLogger = arrayLogger.Dict(dictLogger)
```
jerargus, line 45: nit: This line can be: return arrayLogger.Dict(dictLogger)

cardinal/ecs/world_logger.go
```diff
@@ -0,0 +1,110 @@
...
+		logger: logger,
+		world:  world,
+	}
+}
+
+func (wl *WorldLogger) Log(traceId string, logLevel zerolog.Level, message string) {
+	wl.logger.WithLevel(logLevel).Str("trace_id", traceId)
+}
+
+func (wl *WorldLogger) LogInfo(traceId string, message string) {
```
jerargus, line 28: What will traceId be used for?

Did you consider using a context.Context to keep track of trace ids (and/or span ids)?

cardinal/ecs/world_logger.go
pyrofolium: Yeah that's a bug will add unit tests for these.

cardinal/ecs/world_logger.go
pyrofolium: It's used in LogEntity, see one of the private functions. I will be adding more related methods to this logger that will be utilizing methods on world.

cardinal/ecs/world_logger.go
pyrofolium: No. Did not kno about this. I will look into it and utilize.

cardinal/ecs/world_logger_test.go
pyrofolium: yeah. I was going to eventually use this for other test functions, but even that doesn't really improve anything. Will remove.

cardinal/ecs/world_logger.go
pyrofolium: I researched go contexts. 

My intention with trace ids are indeed used similarly to a unique id for a go context.

However I think it's better to have a string here. 

1. The reason is these functions can be used in places where there is no context (which in turn requires the user to instantiate the context and set a key, value pair) 

2. context may not have the specific key for the trace_id that the logger is looking for. 

What happens is the user just passes the context to the logger, the logger can't find the trace_id and now the resulting log does not have a trace. The user doesn't have to think about trace ids he just fulfills the function signature. 

By keeping it a string here it forces the user to explicitly choose to either have a trace_id by filling it in with something, or if he doesn't care he can pass a useless or empty string. This choice is abstracted away from the user if we pass a context object. He may not understand even the purpose of the context object without consulting the function definition. 

Better to do this if you want to grab it off the context:

```
Log(ctx.Value('trace_id').....)
```

That's just my opinion. I realize it is a bit wordy to do that on every log What do you think?

pyrofolium:
@jerargus @smsunarto I refactored and changed the code completely in response to Scotts comments on Slack. Sorry but it sort of has to be re-evaluated. 

Relevant files are:
system.go
world.go
world_test.go

Changes in other places simply involve changes to the System function signature. This can get confusing so I listed the relevant files above.

cardinal/ecs/world.go
```diff
@@ -38,7 +42,8 @@ type StorageAccessor struct {
 type World struct {
 	namespace                Namespace
 	store                    storage.WorldStorage
-	systems                  []System
+	systems                  []RegisteredSystem
+	systemNames              []string
```
pyrofolium, line 46: systemNames is required because once a System is curried into RegisteredSystem, the function name is lost and reflection can no longer retrieve it. This is required for printing system names short of a full refactoring of the concept of a System. 

Of course if the user used an anonymous function, the name will not be known no matter what.

PR #217, opened 2023-08-21
feat(cardinal): improve logging

pyrofolium:
Closes: #CAR-149

Closed the other PR as there are massive changes. This is a new one. 

This issue injects a logger into system function that already has the system name as a log parameter. 
Additionally added some utility functions to help log additional information about the world, components, systems and entities.

These utility functions do not actually trigger a log. Instead they just augment a zerolog.Event with additional information. The user is suppose to continue to use zerolog methods and patterns to do logging. 

System now has a new type signature that accepts a logger as a parameter. 
There is a new type called RegisteredSystem which is basically the type signature of the old system. 

World curries a new logger into System to form RegisteredSystem then saves it within world. 

## Brief Changelog
Added a new type called RegisteredSystem
relevant files:
     system.go
     world.go 
     world_test.go

Other files were changed simply because changing because the signature of System changed so instantiations of System also needed to be changed the most relevant files with the core changes are listed above.

## Testing and Verifying

Changed all unit tests referencing system to use new function signature
added unit test called TestWorldLogger
test world logger tests:
- a system to see whether it can log correctly and have the system name in the log
- utility functions to print out certain states of various states of certain things.


technicallyty (approved):
LGTM

cardinal/ecs/world.go
```diff
@@ -109,7 +116,15 @@ func (w *World) AddSystems(s ...System) {
 	if w.stateIsLoaded {
 		panic("cannot register systems after loading game state")
 	}
-	w.systems = append(w.systems, s...)
+	for _, system := range s {
```
smsunarto, line 119: lets add a comment here on what this is doing and what it is for

cardinal/ecs/world.go
```diff
@@ -757,3 +773,74 @@ func (w *World) GetTransactionReceipt(id transaction.TxID) (any, []error, bool)
...
 }
+
+func (w *World) loadComponentIntoArrayLogger(component component.IComponentType, arrayLogger *zerolog.Array) *zerolog.Array {
+	dictLogger := zerolog.Dict()
+	dictLogger = dictLogger.Int("component_id", int(component.ID()))
+	dictLogger = dictLogger.Str("component_name", component.Name())
+	return arrayLogger.Dict(dictLogger)
+}
+
+func (w *World) LoadComponentInfoToLogEvent(zeroLoggerEvent *zerolog.Event) *zerolog.Event {
```
smsunarto, line 784: can we add comments to all public methods

cardinal/ecs/system.go
```diff
@@ -1,3 +1,6 @@
 package ecs
 
-type System func(*World, *TransactionQueue) error
+import "github.com/rs/zerolog"
+
+type RegisteredSystem func(*World, *TransactionQueue) error
```
smsunarto, line 5: what is registered system for?

technicallyty, line 5: i don't think we need to export this type

pyrofolium:
@smsunarto this is approved but because of you requested changes I want to get your approval before I merge. If you approve I will merge, otherwise lmk if there are any additional changes you need. Thanks!

smsunarto (approved):
lgtm, but eventually we should revisit where we are putting world.go and log.go, i think we are really cramming everything in the ecs package right now even things that are not really ecs

PR #218, opened 2023-08-23
feat(cardinal): Create cardinal package that only exposes the portions of the world e…

jerargus:
…ngine that game developers need to use

Closes: CAR-110, CAR-140

- Reduce the methods that are exposed by the ecs, and sub packages.
- Reduce the number of packages that are imported when creating a Cardinal based game.
- Put all godoc information into a single package to make it easier to discover/process the Cardinal package.
- Create example_transactiontype_test.go, which demonstrates how to write sample code that can be associated with specific Cardinal methods.

## Testing and Verifying

The best way to verify this PR is to look at the generated godoc for this cardinal package. 

- `git checkout` this branch
- run `godoc` from the `world-engine/cardinal` directory
- navigate to http://localhost:6060/pkg/pkg.world.dev/world-engine/cardinal/

Take note of the Example section of `type TransactionType`


technicallyty:
for those that might not have `godoc` already installed run this command:

`go install golang.org/x/tools/cmd/godoc`

cardinal/query.go
```diff
@@ -0,0 +1,59 @@
...
+
+// QueryCallBackFn represents a function that can operate on a single EntityID.
+type QueryCallBackFn func(EntityID) bool
+
+// Each executes the given callback function on every EntityID that matches this query.
+func (q *Query) Each(w *World, callback QueryCallBackFn) {
+	q.impl.Each(w.impl, func(eid storage.EntityID) bool {
+		return callback(eid)
+	})
+}
```
technicallyty, line 27: we should mention the behavior of returning true/false in the callback function in the comment

cardinal/world.go
```diff
@@ -0,0 +1,168 @@
...
+	return ecs.WithAdapter(adapter)
+}
+
+func WithReceiptHistorySize(size int) WorldOption {
+	return ecs.WithReceiptHistorySize(size)
+}
+
+func WithNamespace(namespace string) WorldOption {
+	return ecs.WithNamespace(namespace)
+}
```
technicallyty, line 168: i know the underlying opts don't have comments, but maybe in this PR we could add some here.

cardinal/example_transactiontype_test.go
smsunarto, line 1: i think we should move this into tests folder, since it's an "external use" test and uses a separate package anyway. it might not be idiomatic to have different package in the same folder. thoughts @technicallyty?

cardinal/world.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+	var opts []server.Option
+	if disableSigVerification {
+		opts = append(opts, server.DisableSignatureVerification())
+	}
+	opts = append(opts, server.WithPort(port))
+	txh, err := server.NewHandler(w.impl, opts...)
+	if err != nil {
+		return err
+	}
```
technicallyty, line 131: the behavior of port setting was to have it be either explicitly set via option, then fallback to an environment variable, then finally defaulting to a default hard coded port. i don't think we want to require it as an explicit parameter here

cardinal/world.go
```diff
@@ -0,0 +1,168 @@
...
+
+	return &World{
+		impl: world,
+	}, nil
+}
+
+// NewWorldInMemory creates a World that uses an in-memory redis DB as the storage
+// layer. This is only suitable for local development.
+func NewWorldInMemory(opts ...WorldOption) (*World, error) {
```
smsunarto, line 61: thoughts of renaming this `NewMockWorld`? I think it's unclear of what NewWorldInMemory means because Redis is also in-memory and it's unclear that this should not be used in production

cardinal/world.go
```diff
@@ -0,0 +1,168 @@
...
+
+type (
+	// EntityID represents a single entity in the World. An EntityID is tied to
+	// one or more components.
+	EntityID = storage.EntityID
+	TxHash   = transaction.TxHash
+)
+
+// NewWorldUsingRedis creates a new World object using Redis as the storage layer.
+func NewWorldUsingRedis(addr, password string, opts ...WorldOption) (*World, error) {
```
smsunarto, line 34: See comments below, following that I think we can just call this `NewWorld`

cardinal/example_transactiontype_test.go
jerargus: This is a standard way to include example code in documentation. For example, see the `sort` package: https://cs.opensource.google/go/go/+/refs/tags/go1.21.0:src/sort/

more reading: https://go.dev/blog/examples

cardinal/world.go
```diff
@@ -0,0 +1,168 @@
...
+// Remove removes the given entity id from the world.
+func (w *World) Remove(id EntityID) error {
+	return w.impl.Remove(id)
+}
+
+// StartGame starts running the world game loop. After loopInterval time passes, a world tick is attempted.
+// In addition, an HTTP server (listening on the given port) is created so that game transactions can be sent
+// to this world. After StartGame is called, RegisterComponents, RegisterTransactions, RegisterReads, and AddSystem(s) may
+// not be called. If StartGame doesn't encounter any errors, it will block on the given done channel.
+func (w *World) StartGame(loopInterval time.Duration, disableSigVerification bool, port string, done chan struct{}) error {
```
technicallyty, line 119: wondering if we need the channel, we could just block on `txh.Serve()`? does the channel give us more benefits?

cardinal/world.go
```diff
@@ -0,0 +1,168 @@
...
+// AddSystem adds the given system to the world object so that the system will be executed
+// at every game tick.
+func (w *World) AddSystem(system System) {
+	w.impl.AddSystem(func(world *ecs.World, queue *ecs.TransactionQueue) error {
+		return system(&World{world}, &TransactionQueue{queue})
+	})
+}
+
+// AddSystems allows for the adding of multiple systems in a single call. See AddSystem for more details.
+func (w *World) AddSystems(systems ...System) {
```
smsunarto, line 76: I think @technicallyty wanted to rename this to RegisterSystem to be consistent with RegisterComponent and RegisterReadds

technicallyty (commented):
just a few comments

cardinal/world.go
jerargus: Yes, I'll add comments. query.go needs some more comments at well.

cardinal/world.go
jerargus: My thinking was that this would allow a user to intercept a SIGINT to stop the game loop... that being said, the done channel just returns and doesn't clean anything up. I'll remove the done channel.

I still want this to block forever, and relying on the behavior of txh.Serve didn't seem right. I could just block forever with `select{}`. Thoughts?

cardinal/world.go
jerargus: Renamed this to NewMockWorld.

cardinal/world.go
jerargus: Renamed to RegisterSystem and RegsiterSystems. Do we need both? How about just RegisterSystems (the plural) that takes in a variable number of systems. RegisterSystems is a little different from the other Register methods because you're allowed to call it multiple times.

cardinal/query.go
jerargus: Added comment.

cardinal/world.go
jerargus: Ok, I used the server.Option as a guide to match the existing behavior. I also re-did the options, which I feel a little weird about. Let me know if it looks ok.

cardinal/world.go
jerargus: I adjusted this to not take in the done channel, and to block forever.

cardinal/world.go
jerargus: done

cardinal/world.go
```diff
@@ -0,0 +1,235 @@
...
+// will be used, and if that is unset, port 4040 will be used.
+func WithHTTPServerPort(port string) WorldOption {
+	return WorldOption{
+		serverOption: server.WithPort(port),
+	}
+}
+
+// WithNoSignatureVerification disables signature verification for the HTTP server. This should only be
+// used for local development.
+func WithNoSignatureVerification() WorldOption {
```
smsunarto, line 206: Thoughts on this being `WithDisableSignatureVerification` for consistency?

cardinal/world.go
```diff
@@ -0,0 +1,235 @@
...
+// signing process.
+func WithNamespace(namespace string) WorldOption {
+	return WorldOption{
+		ecsOption: ecs.WithNamespace(namespace),
+	}
+}
+
+// WithHTTPServerPort specifies the port for the World's HTTP server. If omitted, the environment variable CARDINAL_PORT
+// will be used, and if that is unset, port 4040 will be used.
+func WithHTTPServerPort(port string) WorldOption {
```
smsunarto, line 198: Thoughts on this being `WithPort` for consistency?

cardinal/example_transactiontype_test.go
smsunarto, line 1: thoughts of moving this into `/tests`? cc @technicallyty for go thoughts

cardinal/world.go
```diff
@@ -0,0 +1,235 @@
...
+		impl:          inmem.NewECSWorld(ecsOptions...),
+		serverOptions: serverOptions,
+	}
+	for _, opt := range cardinalOptions {
+		opt(world)
+	}
+	return world, nil
+}
+
+func toITransactionType(ins []AnyTransaction) []transaction.ITransaction {
```
smsunarto, line 82: this might be able to go to utils.go instead?

cardinal/world.go
jerargus: I figured the longer name was more explicit in its behavior, but the function comment should be fine. done.

cardinal/world.go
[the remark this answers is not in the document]
jerargus: I move the options code into a separate options.go file, and I took out some private helpers into util.go. I feel like various source file names chunk up the code into sections pretty well. 

If you still think there need to be section names with the additional files, I can add them.

cardinal/world.go
jerargus: done

cardinal/world.go
```diff
@@ -0,0 +1,235 @@
...
+type World struct {
+	impl          *ecs.World
+	loopInterval  time.Duration
+	serverOptions []server.Option
+}
+
+// System is a function that process the transaction in the given transaction queue.
+// Systems are automatically called during a world tick, and they must be registered
+// with a world using AddSystem or AddSystems.
+type System func(*World, *TransactionQueue) error
```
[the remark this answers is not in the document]
jerargus: I merged main and added a wrapper for the logger.

technicallyty (approved):
👀 looks ✅ good ✌🏻 to 🧍🏻‍♂️ me

PR #219, opened 2023-08-23
fix: add fallbacks to insecure transport for adapter/router

technicallyty:

adds fallbacks to insecure transport for adapter and router services. this helps with testing, so you don't need to have SSL certs on hand when you're testing locally.

## Brief Changelog

- adds insecure transport fallback to adapter and router servers

## Testing and Verifying

spin up a router/adapter without credential options

### DIFF
--- a/cardinal/shard/adapter.go
+++ b/cardinal/shard/adapter.go
@@ -6,6 +6,7 @@ import (
 	"crypto/x509"
 	"fmt"
 	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/credentials/insecure"
 	"os"
 	"pkg.world.dev/world-engine/sign"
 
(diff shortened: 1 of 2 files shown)


cardinal/shard/adapter.go
```diff
@@ -24,21 +25,25 @@ type Adapter interface {
...
 	QueryTransactions(
 		context.Context,
 		*shardtypes.QueryTransactionsRequest) (*shardtypes.QueryTransactionsResponse, error)
 }
 
 type AdapterConfig struct {
-	// ShardReceiverAddr is the address to communicate with the secure shard submission channel.
-	ShardReceiverAddr string `json:"shard_receiver_addr,omitempty"`
+	// ShardSequencerAddr is the address of the base shard's game tx sequencer.
+	ShardSequencerAddr string `json:"shard_receiver_addr,omitempty"`
```
jerargus, line 44: ShardSequencerAddr does not match shard_receiver_addr

technicallyty: got it - thanks for catching that

PR #222, opened 2023-08-24
feat(cardinal): add mechanism to authorize other addresses to control…

technicallyty:
… a persona

extends the persona system to allow a list of addresses to control the persona. this is useful specifically for requests coming from the EVM. A user can now add their EVM address to their persona, letting them make actions from smart contracts for that persona.

also removes RegisteredSystem and fixes a bug wrt systems.

## Brief Changelog

- add new transaction `AuthorizePersonaAddressTx`
- add new system `AuthorizePersonaAddressSystem`
- updates the EVM server to check that a given evm address is an authorized signer of some Persona, and sets the Persona when submitting the transaction to the World's tx queue.
- updated EVM server tests

## Testing and Verifying
- updated existing `persona_test.go` to verify transaction updates SignerComp as intended.
- updated `evm/server_test.go` test to add an AuthorizedAddress before submitting transactions
- added a new test in `evm/server_test.go` that shows transactions fail from EVM if the sender is not an authorized address of some persona.
- added test to verify ordering of systems is happening correctly


cardinal/ecs/persona_test.go
```diff
@@ -114,3 +115,34 @@ func TestDuplicatePersonaTagsInTickAreOnlyRegisteredOnce(t *testing.T) {
...
+
+	wantAddr := "0xfoobar"
+	ecs.AuthorizePersonaAddressTx.AddToQueue(world, ecs.AuthorizePersonaAddress{
+		PersonaTag: wantTag,
+		Address:    wantAddr,
+	}, &sign.SignedPayload{PersonaTag: wantTag})
+	// PersonaTag registration doesn't take place until the relevant system is run during a game tick.
+	assert.NilError(t, world.Tick(context.Background()))
+
+	ecs.NewQuery(filter.Exact(ecs.SignerComp)).Each(world, func(id storage.EntityID) bool {
```
jerargus, line 139: nit: add a count variable before this ecs.NewQuery line, increment the count by 1 in the Each function, and verify the count was incremented. If this NewQuery line fails to find any matches, nonce of the asserts will be hit and it will look like a passing test.

cardinal/ecs/world.go
```diff
@@ -469,8 +469,9 @@ func (w *World) Tick(ctx context.Context) error {
 		queue: txs,
 	}
 
-	for _, sys := range w.systems {
-		if err := sys(w, txQueue); err != nil {
+	for i, sys := range w.systems {
+		systemLogger := w.logger.CreateSystemLogger(w.systemNames[i])
```
jerargus, line 473: An advantage of the previous pattern was that this systemLogger was only being created once for each registered system. In this case we're creating new instances for each tick. Can a systemLogger slice replace the systemNames slice so we allocate them just once?

cardinal/ecs/world_test.go
```diff
@@ -26,6 +26,28 @@ func TestAddSystems(t *testing.T) {
 	assert.Equal(t, count, 3)
 }
 
+func TestSystemExecutionOrder(t *testing.T) {
```
jerargus, line 29: nice

cardinal/evm/server_test.go
```diff
@@ -46,22 +48,45 @@ func TestServer_SendMessage(t *testing.T) {
...
-		inFooTxs := FooTx.In(queue)
-		inBarTxs := BarTx.In(queue)
-		assert.Equal(t, len(inFooTxs), len(fooTxs))
-		assert.Equal(t, len(inBarTxs), len(barTxs))
-		for i, tx := range inFooTxs {
-			assert.DeepEqual(t, tx.Value, fooTxs[i])
-		}
-		for i, tx := range inBarTxs {
-			assert.DeepEqual(t, tx.Value, barTxs[i])
+		if enabled {
```
jerargus, line 57: nit: if !enabled{ return }

cardinal/ecs/world.go
technicallyty: good idea

cardinal/ecs/persona_test.go
technicallyty: ah right. good catch. fixed

cardinal/ecs/world.go
technicallyty: added

cardinal/evm/server_test.go
technicallyty: added

jerargus (approved):

namrapatel (approved):

pyrofolium:
lgtm

pyrofolium (commented):
lgtm

PR #223, opened 2023-08-24
feat(cardinal): add logging to specific areas

pyrofolium:
Closes: #CAR-174

Completes CAR-149, CAR-160 was about making logging utilities, 

this issue is for using the log utilities around certain parts of world

## Brief Changelog

- logs when tick starts
- logs when tick ends (and logs difference in transaction queue)
- logs when entity is created 
- logs when entity is destroyed
- logs when entity is updated.
- logs when archetype is created
- there was a task to log when error occurs, errors tend bubble up to to StartGameLoop where an error is logged already I just updated this to use the logger embedded in world rather then the global zerologger logger. 
- 

## Testing and Verifying
See TestWorldLogger. Also good to see this for helpful visualization of what's being printed. 

I might've put things in the wrong place or there may not be enough information in the log. Please let me know


cardinal/ecs/world.go
```diff
@@ -458,6 +465,8 @@ func (w *World) AddTransaction(id transaction.TypeID, v any, sig *sign.SignedPay
 // Tick performs one game tick. This consists of taking a snapshot of all pending transactions, then calling
 // each System in turn with the snapshot of transactions.
 func (w *World) Tick(ctx context.Context) error {
+	tickAsString := strconv.FormatUint(w.tick, 10)
+	w.logger.Info().Str("tick", tickAsString).Msg("Tick started")
```
pyrofolium, line 469: Might be slowish here due to conversion to string. zerolog doesn't support uint64, but I can convert to int.

technicallyty: could also just do `fmt.Sprintf("%d", w.tick)`

technicallyty: not sure if one is faster than the other though

technicallyty:
lets keep the PR scope the same casing as what we're scoping to - i.e. since we're only editing the `cardinal` (no capital c) directory here, PR title should be `feat(cardinal)` 😁

cardinal/ecs/log_test.go
```diff
@@ -88,18 +112,79 @@ func TestWorldLogger(t *testing.T) {
...
+	require.JSONEq(t, `
+			{
+				"level":"info",
+				"tick":"0",
+				"message":"Tick started"
+			}`, logStrings[0])
+	// test if system name recorded in log
+	require.JSONEq(t, `
+			{
+				"system":"ecs_test.testSystem",
```
smsunarto, line 138: nice

cardinal/ecs/world.go
```diff
@@ -277,7 +279,10 @@ func (w *World) createEntity(archetypeID storage.ArchetypeID) (storage.EntityID,
 		return 0, err
 	}
 	archetype.PushEntity(nextEntityID)
-
+	err = w.logger.LogEntity(w, zerolog.DebugLevel, nextEntityID)
```
smsunarto, line 282: I would prefer if LogEntity doesn't return an error and logs its own error instead, but cc @technicallyty for his thoughts here

cardinal/ecs/world.go
```diff
@@ -486,7 +494,9 @@ func (w *World) Tick(ctx context.Context) error {
 	}
 	w.tick++
 	w.receiptHistory.NextTick()
-
+	w.logger.Info().
+		Str("tick", tickAsString).
+		Msg("Tick ended")
```
smsunarto, line 499: Can we also log how long it took for the tick to finish executing?

cardinal/ecs/world.go
smsunarto, line 1: Can we also log:
- [INFO] Log how many and the name of the registered system, component, read, and tx
- [INFO] During start, log cardinal instance data (what port is it running, what address is redis running on)
- [WARNING] Warn if using miniredis (embedded redis instance meant only for testing)
- [WARNING] Warn if using redis without password
- [WARNING] During start, log when no system is registered
- [WARNING] During start, log when no component is registered
- [WARNING] During start, log when no read is registered
- [WARNING] During start, log when no tx is registered
- [WARNING] During start, log when DisableSignatureVerification option is enabled (this should not be enabled in production)

cardinal/ecs/world.go
pyrofolium: Yeah, you're right. I agree your suggestion better.

cardinal/ecs/world.go
pyrofolium: Our logger lives in world. Some of the data required here lives in nested parts of world which necessitates passing the logger to those parts which ultimately changes the signature for many functions and interfaces. It's a rather large change if we go this route. 

I'm thinking we can just use the global zeroLogger or instantiating a new log for logging in private places (like storage for example) that the user tends not to touch and just have the "world logger" be a public logger for user space or things in scope. 

I can go whatever path. But I think this is better to be scoped into another issue which I can take on as it's rather large. What do you think if we lock in the changes in this PR first? @smsunarto

cardinal/ecs/world.go
pyrofolium: This is done.

cardinal/ecs/world.go
pyrofolium: This is done.

cardinal/ecs/world.go
pyrofolium: I did all the above but I believe these three:

[INFO] During start, log cardinal instance data (what port is it running, what address is redis running on)
[WARNING] Warn if using miniredis (embedded redis instance meant only for testing)
[WARNING] Warn if using redis without password

Worth moving to another issue.

smsunarto (approved):

PR #234, opened 2023-08-29
fix(cardinal): fix flaky http tests

jerargus:
Closes: #CAR-115 (hopefully)

Fix the flaky tests in cardinal/server. Two issues were identified:

1) http's ListenAndServe returns an error when the server is shut down, and Handler.Serve would log a fatal error when ListenAndServe returns an error. During tests, it is desirable to close the http sever to free up the port for other tests. This was causing log.Fatal to be called during tests, which would not run all future tests. *The fix:* Return an error from Handler.Serve and let the caller decide how to handle the error.

2) While startup up an http server is fast, it's not instantaneous. In tests, we start the server on a non-main goroutine, but there's no guarantee that the server would be ready to handle traffic once the test code on the main threat was reached. *The fix:* Create a health check endpoint for tests that we query before any test code is run. Errors from this health check endpoint are ignored (as the server is still starting up). Once the health check returns 200, allow the test logic to run.  

## Testing and Verifying

This PR verifies very little non-test code. Existing unit tests should verify the change.

### DIFF
--- a/cardinal/server/server.go
+++ b/cardinal/server/server.go
@@ -3,12 +3,12 @@ package server
 import (
 	"errors"
 	"fmt"
-	"log"
 	"net/http"
 	"os"
+	"strconv"
+
 	"pkg.world.dev/world-engine/cardinal/ecs"
 	"pkg.world.dev/world-engine/cardinal/shard"
-	"strconv"
 )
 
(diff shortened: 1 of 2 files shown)


pyrofolium (approved):
lgtm

PR #238, opened 2023-08-30
fix(cardinal): Add the AddTo and RemoveFrom functions to components

jerargus:

Add the RemoveFrom and AddTo function to the cardinal package. This allows users to change the components that exist on entities.

## Testing and Verifying

### DIFF
--- a/cardinal/component.go
+++ b/cardinal/component.go
@@ -46,6 +46,16 @@ func (c *ComponentType[T]) Name() string {
 	return c.impl.Name()
 }
 
+// RemoveFrom removes this component form the given entity.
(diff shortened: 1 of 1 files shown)


cardinal/component.go
```diff
@@ -46,6 +46,16 @@ func (c *ComponentType[T]) Name() string {
 	return c.impl.Name()
 }
 
+// RemoveFrom removes this component form the given entity.
```
technicallyty, line 49:
```suggestion
// RemoveFrom removes this component from the given entity.
```

technicallyty (approved):
LGTM, 1 typo

PR #239, opened 2023-08-30
feat(cardinal): add new swagger server with one implimented route /query/http/endpoints #1

pyrofolium:
Closes: #CAR-183

First PR for creating a server that will validate at runtime the swagger spec. It has a bunch of bs routes made to satisfy go-swagger and one route implimented for /query/http/endpoints

All old code still should work this replaces nothing and adds swagger as an additional feature. 

This is a sub-issue of CAR-155 which is to fully flesh out the swagger server. 

## Brief Changelog

created `swagger.yml`
added new functions to `server.go`: `NewSwaggerHandler`, `registerReadHandlerSwagger`,  `registerTxHandlerSwagger`, `getValueFromParams`

## Testing and Verifying

Ran all unit tests. still works added a new unit test that validates the /query/http/endpoints


cardinal/server/server.go
```diff
@@ -3,12 +3,20 @@ package server
...
 	"os"
+	"pkg.world.dev/world-engine/cardinal/ecs/transaction"
 	"strconv"
 
 	"pkg.world.dev/world-engine/cardinal/ecs"
 	"pkg.world.dev/world-engine/cardinal/shard"
+
+	//"github.com/go-openapi/errors"
+	"github.com/go-openapi/loads"
+	swagger_runtime "github.com/go-openapi/runtime"
```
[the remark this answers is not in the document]
pyrofolium: There was a collision here earlier. As I coded, that collision was discarded somehow? in a merge I believe? will remove.

cardinal/server/server.go
```diff
@@ -46,6 +54,161 @@ const (
...
+	coreHandler := swagger_runtime.OperationHandlerFunc(func(params interface{}) (interface{}, error) {
+		txType, ok := getValueFromParams(&params, "txType")
+		if !ok {
+			return nil, errors.New("txType not found in params")
+		}
+		txTypeString, ok := txType.(string)
+		if !ok {
+			return nil, errors.New("txType was not a string")
+		}
+		fmt.Print(txTypeString)
```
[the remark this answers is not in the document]
pyrofolium: no this is to be removed. just temporary stuff for me personally. Will remove.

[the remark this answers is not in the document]
pyrofolium: Also this is a stub route. Unimplemented. I only implemented one route. Will remove all the extra stuff to prevent confusion.

cardinal/server/server.go
```diff
@@ -46,6 +53,152 @@ const (
...
+		}{test: "test"}, errors.New("not implemented")
+	})
+	listHandler := runtime.OperationHandlerFunc(func(params interface{}) (interface{}, error) {
+		txs, err := world.ListTransactions()
+		txEndpoints := make([]string, 0, len(txs))
+		if err != nil {
+			return nil, err
+		}
+		for _, tx := range txs {
+			if tx.Name() != "create-persona" && tx.Name() != "authorize-persona-address" {
```
jerargus, line 155: This line should be replaced with:

`if tx.Name() != ecs.CreatePersonaTx.Name() && tx.Name() != ecs.AuthorizePersonaAddressTx.Name() {...`

cardinal/server/server.go
```diff
@@ -46,6 +53,152 @@ const (
...
+		}
+		queryEndpoints = append(queryEndpoints, "/query/http/endpoints")
+		queryEndpoints = append(queryEndpoints, "/query/persona/signer")
+		queryEndpoints = append(queryEndpoints, "/query/receipt/submit")
+		return EndpointsResult{
+			TxEndpoints:    txEndpoints,
+			QueryEndpoints: queryEndpoints,
+		}, nil
+	})
+	personaHandler := runtime.OperationHandlerFunc(func(i interface{}) (interface{}, error) {
```
jerargus, line 174: This should be defined in ECS (just like create-persona endpoint). I think that change is outside of the scope of this PR, but I added a ticket: https://linear.app/arguslabs/issue/CAR-190/convert-the-read-persona-signer-endpoint-to-a-ecs-based-read-endpoint -- would you mind adding a comment about how this should get moved to the ECS pattern and add a link to that linear ticket?

cardinal/server/server.go
```diff
@@ -46,6 +53,152 @@ const (
...
+		return ReadPersonaSignerResponse{Status: "", SignerAddress: ""}, errors.New("not implemented")
+	})
+	receiptsHandler := runtime.OperationHandlerFunc(func(i interface{}) (interface{}, error) {
+		return struct {
+			start_tick uint64
```
jerargus, line 179: What's the purpose of returning these empty structs? Will these function eventually return actual structs, and this code is just serving as samples for what it will eventually look like?

cardinal/server/server.go
```diff
@@ -46,6 +53,152 @@ const (
...
+			tx_hash string
+			tick    uint64
+			result  interface{}
+			errors  []string
+		}{}}, errors.New("not implemented")
+	})
+	api.RegisterOperation("GET", "/query/core/{readType}", coreHandler)
+	api.RegisterOperation("GET", "/query/http/endpoints", listHandler)
+	api.RegisterOperation("GET", "/query/persona/signer", personaHandler)
+	api.RegisterOperation("GET", "/query/receipts/submit", receiptsHandler)
```
jerargus, line 197: I don't like this name. This endpoint takes in some world tick and returns a list of transactions between the given tick and the current tick. how about...
`/quer/receipt/list`
or
`/query/receipt/since`

'submit' seems to generic.

cardinal/server/server.go
```diff
@@ -46,6 +53,152 @@ const (
 	getSignerForPersonaStatusAssigned  = "assigned"
 )
 
+// NewSwaggerHandler instantiates handler function for creating a swagger server that validates itself based on a swagger spec.
+func NewSwaggerHandler(w *ecs.World, pathToSwaggerSpec string, opts ...Option) (*Handler, error) {
```
jerargus, line 57: Did you consider making the pathToSwaggerSpec parameter into an Option?

cardinal/server/swagger.yml
```diff
@@ -0,0 +1,300 @@
...
+                      $ref: '#/definitions/TxRequest'
+            responses:
+                '200':
+                    description: successful operation
+                    schema:
+                        $ref: '#/definitions/TxReply'
+                '400':
+                    description: Invalid transaction request
+    /query/core/{readType}:
+        get:
```
jerargus, line 86: Based on that slack conversation, do these need to be changed to `post:`

cardinal/server/swagger.yml
```diff
@@ -0,0 +1,300 @@
...
+        type: object
+        properties:
+            tx_hash:
+                type: string
+            tick:
+                type: integer
+                format: int64
+    TxRequest:
+        required:
+            - personaTag
```
jerargus, line 239: Not 100% sure, but this should probably be "persona_tag"

cardinal/server/server_test.go
```diff
@@ -203,6 +210,95 @@ func TestHandleTransactionWithNoSignatureVerification(t *testing.T) {
 	assert.Equal(t, 1, count)
 }
 
+func TestHandleSwaggerServer(t *testing.T) {
+	count := 0
+	endpoint := "move"
```
jerargus, line 215: since this is only used in one place, inline to line 217. 

nit: make it "send-energy".

cardinal/server/server_test.go
```diff
@@ -203,6 +210,95 @@ func TestHandleTransactionWithNoSignatureVerification(t *testing.T) {
...
+		Age  uint64
+	}
+
+	expectedReply := FooReply{
+		Name: "Chad",
+		Age:  22,
+	}
+	fooRead := ecs.NewReadType[FooRequest, FooReply]("foo", func(world *ecs.World, req FooRequest) (FooReply, error) {
+		return expectedReply, nil
+	}, ecs.WithReadEVMSupport[FooRequest, FooReply])
```
jerargus, line 257: Is ecs.WithReadEVMSupport required for this test?

cardinal/server/server_test.go
```diff
@@ -203,6 +210,95 @@ func TestHandleTransactionWithNoSignatureVerification(t *testing.T) {
...
+	err := w.RegisterReads(fooRead)
```
jerargus, line 258: this line should be: `assert.NilError(t, w.RegisterReads(fooReaD))

cardinal/server/server_test.go
```diff
@@ -203,6 +210,95 @@ func TestHandleTransactionWithNoSignatureVerification(t *testing.T) {
...
+	signedTx := sign.SignedPayload{
+		PersonaTag: "some_persona",
+		Namespace:  "some_namespace",
+		Nonce:      100,
+		// this bogus signature is OK because DisableSignatureVerification was used
+		Signature: common.Bytes2Hex([]byte{1, 2, 3, 4}),
+		Body:      bz,
+	}
+
+	bz, err = json.Marshal(&signedTx)
```
jerargus, line 278: nit: signedTx.Marshal() is a thing.

cardinal/server/server.go
```diff
@@ -46,6 +53,152 @@ const (
...
+	})
+	listHandler := runtime.OperationHandlerFunc(func(params interface{}) (interface{}, error) {
+		txs, err := world.ListTransactions()
+		txEndpoints := make([]string, 0, len(txs))
+		if err != nil {
+			return nil, err
+		}
+		for _, tx := range txs {
+			if tx.Name() != "create-persona" && tx.Name() != "authorize-persona-address" {
+				txEndpoints = append(txEndpoints, "/tx/core/"+tx.Name())
```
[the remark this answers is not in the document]
pyrofolium: I mentioned that a "game" could be a collection of plugins with no central concept so we don't need to force a section to be called "game". Then Scott suggested we call it "core" on the white board (it's still there written above game) which sort of defeats my suggestion. I don't care too much either way. I will change it back.

cardinal/server/server.go
pyrofolium: sure I'm defining this endpoint in the next PR in a sub issue for this one. I will link your issue with the sub issue and move it accordingly.

cardinal/server/swagger.yml
pyrofolium: Actually it's PersonaTag. All the names should be capitalized. I will correct it now. When I get to implementing and testing each route, I will automatically have to make these correct so for now if it's rough i think it's fine.

cardinal/server/server_test.go
pyrofolium: It probably will be for the other routes I will be implementing.

cardinal/server/server.go
pyrofolium: Yeah, it's just stubs. The anonymous structs are patterned after the swagger spec file.

cardinal/server/server.go
pyrofolium: Yeah I don't know, Seems more straightforward as a string in my opinion. I'm not to tied to either way.

cardinal/server/swagger.yml
```diff
@@ -0,0 +1,300 @@
...
+                '200':
+                    description: successful operation
+                    schema:
+                        $ref: '#/definitions/TxReply'
+                '400':
+                    description: Invalid transaction request
+    /tx/persona/create-persona:
+        post:
+            summary: Create a Persona transaction to Cardinal
+            description: Submit a Persona transaction to Cardinal
```
smsunarto, line 44: nit: let's just make this the same?

cardinal/server/swagger.yml
```diff
@@ -0,0 +1,300 @@
+consumes:
+    - application/json
+info:
+    description: Backend server for world engine
```
smsunarto, line 4: capitalize World Engine

jerargus (approved):

PR #250, opened 2023-09-04
fix(chain): world binary help text, script fixes

technicallyty:

fixes the help text for the chain CLI

fixes the script to remove all `world` chain data before running again.

### DIFF
--- a/chain/cmd/world/cmd/root.go
+++ b/chain/cmd/world/cmd/root.go
@@ -104,8 +104,8 @@ func NewRootCmd() *cobra.Command {
 		WithKeyringOptions(keyring.EthSecp256k1Option())
 
 	rootCmd := &cobra.Command{
-		Use:   "polard",
-		Short: "polaris sample app",
+		Use:   "world",
+		Short: "world engine application",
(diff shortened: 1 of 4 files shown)


heronimus (commented):
If we don't specify `--build` docker will try to pull from GCP images instead, maybe we can remove the `image` part from [docker-compose.yml:L12](https://github.com/Argus-Labs/world-engine/blob/38a7dc52a4f7432c2f73b088c5a984d9fd57a57a/chain/docker-compose.yml#L12C5-L12C66) to avoid confusion

technicallyty:
> If we don't specify `--build` docker will try to pull from GCP images instead, maybe we can remove the `image` part from [docker-compose.yml:L12](https://github.com/Argus-Labs/world-engine/blob/38a7dc52a4f7432c2f73b088c5a984d9fd57a57a/chain/docker-compose.yml#L12C5-L12C66) to avoid confusion

ahh i see. i removed `--build` as it takes quite a while to build (mostly gets hung on `go mod download` step). maybe i can also fix that in this PR by using `vendor`. 

but in any case, can just remove the `image` part, since it seems it wasn't getting used with the `--build` thing there anyway

heronimus (approved):

PR #253, opened 2023-09-05
pretty logging option for cardinal.

pyrofolium:
Closes: #CAR-197

Option to add pretty logging to cardinal

## Brief Changelog

Just a new function in options.go

## Testing and Verifying

Visually. Changes aren't human readable directly you can only see the results if it's written to os.Stderr.

### DIFF
--- a/cardinal/option.go
+++ b/cardinal/option.go
@@ -1,8 +1,11 @@
 package cardinal
 
 import (
+	"os"
 	"time"
 
+	"github.com/rs/zerolog"
+	"github.com/rs/zerolog/log"
 	"pkg.world.dev/world-engine/cardinal/ecs"
 	"pkg.world.dev/world-engine/cardinal/server"
 	"pkg.world.dev/world-engine/cardinal/shard"
@@ -62,4 +65,9 @@ func WithLoopInterval(interval time.Duration) WorldOption {
 			world.loopInterval = interval
 		},
 	}
+}
+
+func WithPrettyLogOption(world *ecs.World) {
+	prettyLogger := log.Output(zerolog.ConsoleWriter{Out: os.Stderr})
+	world.Logger.Logger = &prettyLogger
 }


cardinal/option.go
```diff
@@ -63,3 +66,8 @@ func WithLoopInterval(interval time.Duration) WorldOption {
 		},
 	}
 }
+
+func WithPrettyLogOption(world *ecs.World) {
```
jerargus, line 70: This should be `func WithPrettyLog(world *ecs.World) WorldOption`. It's functionally the same thing, but by returning a WorldOption this function will get grouped with the other options in generated godocs.

technicallyty:
i think this needs to be implemented more like this option:

```go
// WithLoopInterval sets the time between ticks. If left unset, 1 second is used as a default.
func WithLoopInterval(interval time.Duration) WorldOption {
	return WorldOption{
		cardinalOption: func(world *World) {
			world.loopInterval = interval
		},
	}
}
```

pyrofolium:
> i think this needs to be implemented more like this option:
> 
> ```go
> // WithLoopInterval sets the time between ticks. If left unset, 1 second is used as a default.
> func WithLoopInterval(interval time.Duration) WorldOption {
> 	return WorldOption{
> 		cardinalOption: func(world *World) {
> 			world.loopInterval = interval
> 		},
> 	}
> }
> ```

Hm. This world is not ecs.World. I need a function that that impliments this for ecs.World. It think this is the wrong file.

pyrofolium (commented):
@jerargus @technicallyty Thanks for reviewing I think I made a big mistake here. There's two worlds and two option files in world-engine. I was using the wrong options.go file and wrong world. It needs to be ecs.World as this is the function signature needed in the starter-template to replace the logger.

jerargus (approved):

PR #261, opened 2023-09-11
feat(cardinal): create parser for component query language.

pyrofolium:
Closes: #WORLD-109

Adds a parser for a dynamic component query language. The intention of this language is to be parsed into a LayoutFilter type. 

Also branch name is wrong. Please ignore this is referencing WORLD 300, 109 is the EPIC. 

`
!(EXACT(a, b) & EXACT(a)) | CONTAINS(b)
`

CQL stands for component query language. 

## Brief Changelog
Added two files. cql.go and cql_test.go

cql.go generates a parser:  string -> AST

## Testing and Verifying

This expression was tested: 

`!(EXACT(a, b) & EXACT(a)) | CONTAINS(b)` it covers multiple functions contains and exact. It covers & and |. It also covers higher precedence with parenthesis and the NOT (!) operator.

### DIFF
--- a/cardinal/ecs/cql/cql.go
+++ b/cardinal/ecs/cql/cql.go
@@ -7,70 +7,70 @@ import (
 	"github.com/alecthomas/participle/v2"
 )
 
-type CQLOperator int
+type cqlOperator int
 
 const (
-	OpAnd CQLOperator = iota
-	OpOr
+	opAnd cqlOperator = iota
+	opOr
 )
 
-var operatorMap = map[string]CQLOperator{"&": OpAnd, "|": OpOr}
+var operatorMap = map[string]cqlOperator{"&": opAnd, "|": opOr}
 
-func (o *CQLOperator) Capture(s []string) error {
+func (o *cqlOperator) Capture(s []string) error {
 	*o = operatorMap[s[0]]
 	return nil
 }
 
-type CQLComponent struct {
+type cqlComponent struct {
 	Name string `@Ident`
 }
 
-type CQLNot struct {
-	SubExpression *CQLValue `"!" @@`
+type cqlNot struct {
+	SubExpression *cqlValue `"!" @@`
 }
 
-type CQLExact struct {
-	Components []*CQLComponent `"EXACT""(" (@@",")* @@ ")"`
+type cqlExact struct {
+	Components []*cqlComponent `"EXACT""(" (@@",")* @@ ")"`
 }
 
-type CQLContains struct {
-	Components []*CQLComponent `"CONTAINS" "(" (@@",")* @@ ")"`
+type cqlContains struct {
+	Components []*cqlComponent `"CONTAINS" "(" (@@",")* @@ ")"`
 }
 
-type CQLValue struct {
-	Exact         *CQLExact    `@@`
-	Contains      *CQLContains `| @@`
-	Not           *CQLNot      `| @@`
-	Subexpression *CQLTerm     `| "(" @@ ")"`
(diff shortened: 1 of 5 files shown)


cardinal/ecs/cql/cql.go
```diff
@@ -0,0 +1,135 @@
...
+
+const (
+	opAnd cqlOperator = iota
+	opOr
+)
+
+var operatorMap = map[string]cqlOperator{"&": opAnd, "|": opOr}
+
+func (o *cqlOperator) Capture(s []string) error {
+	*o = operatorMap[s[0]]
```
technicallyty, line 20: do we need to check that s will have non-zero length?

cardinal/ecs/cql/cql.go
```diff
@@ -0,0 +1,135 @@
...
+type cqlOperator int
+
+const (
+	opAnd cqlOperator = iota
+	opOr
+)
+
+var operatorMap = map[string]cqlOperator{"&": opAnd, "|": opOr}
+
+func (o *cqlOperator) Capture(s []string) error {
```
technicallyty, line 19: might be nice to comment that this implements an interface from participle. the `string` methods are obvious enough, but since this comes from an outside lib, could be useful here

cardinal/ecs/storage/legacy_storage.go
```diff
@@ -10,7 +10,7 @@ import (
 var _ ComponentStorageManager = &ComponentsSliceStorage{}
 
 // ComponentsSliceStorage is a structure that contains component data in slices.
-// Component data is indexed by component type ID, archetype Index, and finally component Index.
+// CQLComponent data is indexed by component type ID, archetype Index, and finally component Index.
```
technicallyty, line 13: looks like a victim of Goland's auto-refactor tool

cardinal/ecs/cql/cql.go
pyrofolium: I think the parser lib does not even execute Capture unless it sees a "&" or a "|". But you're right might as well plug this hole.

cardinal/ecs/cql/cql.go
```diff
@@ -0,0 +1,144 @@
...
+	for i, comp := range e.Components {
+		parameters += comp.Name
+		if i < len(e.Components)-1 {
+			parameters += ", "
+		}
+	}
+	return "CONTAINS(" + parameters + ")"
+}
+
+func (v *cqlValue) String() string {
```
technicallyty, line 105: for the other String methods you had accounted for the extra ", " at the end. should we also do that here? 

putting `EXACT(a, b, c)` produces `EXACT(a, b, c, )`

cardinal/ecs/cql/cql.go
```diff
@@ -0,0 +1,144 @@
...
+		for _, comp := range v.Contains.Components {
+			parameters += comp.Name + ", "
+		}
+		return "CONTAINS(" + parameters + ")"
+	} else if v.Not != nil {
+		return "!(" + v.Not.SubExpression.String() + ")"
+	} else if v.Subexpression != nil {
+		return "(" + v.Subexpression.String() + ")"
+	} else {
+		panic("blah")
```
technicallyty, line 123: need 2 update panic msg here

cardinal/ecs/cql/cql.go
pyrofolium: This was just code for debugging doesn't touch anything. It can reuse the sub-methods. I'll change.

PR #262, opened 2023-09-11
feat(cardinal): convert ast to filter. 

pyrofolium:
Closes: #WORLD-301

adds to https://github.com/Argus-Labs/world-engine/pull/261

By converting the generated AST to a LayoutFilter type which our framework can use to query things. 

## Brief Changelog

additional functions to cql.go and new tests

## Testing and Verifying

Added tests to cql_test.go


cardinal/ecs/cql/cql.go
```diff
@@ -133,3 +136,76 @@ func (t *cqlTerm) String() string {
...
+			if err != nil {
+				return nil, err
+			}
+			if *operator == opAnd {
+				acc = filter.And(acc, resultFilter)
+			} else if *operator == opOr {
+				acc = filter.Or(acc, resultFilter)
+			} else {
+				return nil, errors.New("invalid operator")
+			}
```
technicallyty, line 207: smol nit: this might look nicer as a switch case 

also, looks like we can just do return the filter instead of setting it to `acc` here

pyrofolium: that's fine, I'm used to python so i like if-else. 

I'm not sure what you mean by not returning the acc? The acc is all the filters combined?

technicallyty (approved):

technicallyty:
can you update the PR title before merging?

PR #263, opened 2023-09-11
feat(cardinal): fetch component by name.

pyrofolium:
Closes: #WORLD-299

Allow fetching component by name. 

Currently many tests don't utilize the component name, additionally names now have to be unique. 

This issue is worth a conversation about usage of component names. Right now component names live in code as types and don't really exist in run time. There is a property called "Names" but I don't know how often that's used as uniqueness and explicit naming isn't enforced. There is a bit of reflection to sometimes match the type name but this isn't enforced strictly either and the results aren't always unique. 

I'm putting this up as a PR for now. It could change depending on the conversation.

I think this is worth a bit of a conversation then writing a secondary issue on top of this. Will bring it up on next stand up then update this. 

## Brief Changelog

Adds a new map to World that associates name with IComponent. 

Added some tests to verify fetching of the component, and changed a lot of tests to make sure all components had unique names. 

Relevant files:

cardinal/ecs/world.go
cardinal/component.go

All the other changes are just fixing the tests because of the signature change.

### DIFF
--- a/cardinal/component.go
+++ b/cardinal/component.go
@@ -27,17 +27,17 @@ func toIComponentType(ins []AnyComponentType) []component.IComponentType {
 
 // NewComponentType creates a new instance of a ComponentType. When this ComponentType is added to an EntityID,
 // the zero value of the T struct will be saved with the entity.
-func NewComponentType[T any]() *ComponentType[T] {
+func NewComponentType[T any](name string) *ComponentType[T] {
 	return &ComponentType[T]{
-		impl: ecs.NewComponentType[T](),
+		impl: ecs.NewComponentType[T](name),
 	}
 }
 
 // NewComponentTypeWithDefault creates a new instance of a ComponentType. When this ComponentType is added to
 // an EntityID, the defaultVal will be saved with the entity.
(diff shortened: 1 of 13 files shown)


technicallyty (commented):
i think id prefer to have the name be set via options (or new constructor), rather than new method.  wdyt? cc @jerargus

technicallyty:
also can you fix the PR title?

pyrofolium:
> i think id prefer to have the name be set via options (or new constructor), rather than new method.  wdyt? cc @jerargus 

I think the constructor option is better imo. If people are going to query this by name it should be a required typed checked param.

pyrofolium:
> also can you fix the PR title?

fixed

technicallyty:
from stand up today - change here to make component name a required constructor argument

pyrofolium:
@technicallyty and whoever reviews:

cardinal/ecs/world.go
cardinal/component.go

These are the relevant files, the rest are just changes to test to account for the fact that components need names now.

technicallyty (approved):
lgtm

PR #264, opened 2023-09-13
feat(cardinal): add route to swagger to return parsed query results. 

pyrofolium:
Closes: #WORLD-302

add a route where you can POST your cql query and get results in the form of []{id: entityid, data: []Any}. data stores component information. 

## Brief Changelog

Lots of changes. 

- Added reflection to dynamically call the Get method on the IComponentType interface. This will panic if it's done on a mock. 
- Added a new route in swagger and in code. 
- Retrieving the component by string name was a function that had a bool in the return value. By using this I had to update cql parser functions to account for this. You'll see some of this. 

This one is pretty big unfortunately I don't see an easy way to split it into smaller PRs. 

## Testing and Verifying

Added tons and tons of tests.


cardinal/server/server.go
```diff
@@ -355,6 +358,78 @@ func registerReadHandlerSwagger(world *ecs.World, api *untyped.API, handler *Han
 		getListTxReceiptsReplyFromRequest(world),
 	)
 
+	cqlHandler := runtime.OperationHandlerFunc(func(params interface{}) (interface{}, error) {
```
technicallyty, line 361: optional: `any` can be used as an alias to `interface{}`. makes it a tiny bit less verbose, but up to you if you wanna use it!

cardinal/server/server.go
```diff
@@ -355,6 +358,78 @@ func registerReadHandlerSwagger(world *ecs.World, api *untyped.API, handler *Han
...
+				return false
+			}
+			components := entity.GetComponents(world)
+			resultElement := cql.QueryResponse{
+				id,
+				make([]json.RawMessage, 0),
+			}
+
+			//The way our framework is set up it's not designed to retrieve components dynamically at runtime.
+			//As a result we have to use reflection which is generally bad and expensive.
```
technicallyty, line 402: i wonder if we can just make a new method on the interface IComponentType that returns `interface{} / any` for component data, which is JSON marshal-able. 

would require passing in `world` which in turn would cause circular dep, but maybe we don't need `IComponentType` in its own package? maybe we can make a new ticket and follow up afterwards if that adds too much overhead to this PR

cardinal/server/server_test.go
```diff
@@ -381,6 +394,79 @@ func TestHandleSwaggerServer(t *testing.T) {
...
 	assert.Equal(t, resp6.StatusCode, 404)
 
+	//test query/game/cql
+	jsonQuery := struct{ CQL string }{"blah"}
+	jsonQueryBytes, err := json.Marshal(jsonQuery)
+	assert.NilError(t, err)
+	resp7, err := http.Post(txh.makeURL("query/game/cql"), "application/json", bytes.NewBuffer(jsonQueryBytes))
+	assert.NilError(t, err)
+	assert.Equal(t, resp7.StatusCode, 422)
+
```
technicallyty, line 404: would it be possible to use a table driven test here? might be a bit nicer to deal w/

pyrofolium: hm my honest thoughts on this is that it's just tests. It doesn't matter too much as long as the tests are readable. copy + pasta is just more efficient here imo. 

But I don't really care, I can refactor this into a table.

cardinal/server/server.go
pyrofolium: Yeah this makes sense. I see what you're saying. I also agree I'll put this in another ticket.

cardinal/server/server.go
pyrofolium: eh. Next time.

technicallyty (approved):
LGTM, one final nit

PR #267, opened 2023-09-13
feat(cardinal): create additional context for panics in logging. 

pyrofolium:
Closes: #WORLD-308

Additional logging during a system panic

## Brief Changelog

Added a recovery function that will log current system name and tick number, before forwarding the panic. 

## Testing and Verifying

Added a test to tick_test.go

### DIFF
--- a/cardinal/ecs/tick_test.go
+++ b/cardinal/ecs/tick_test.go
@@ -1,10 +1,15 @@
 package ecs_test
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
 	"errors"
+	"fmt"
+	"io"
 	"testing"
 
+	"github.com/rs/zerolog"
 	"pkg.world.dev/world-engine/cardinal/ecs/internal/testutil"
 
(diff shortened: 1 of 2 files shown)


cardinal/ecs/tick_test.go
```diff
@@ -35,6 +40,59 @@ func TestTickHappyPath(t *testing.T) {
...
+	// In this test, our "buggy" system fails once Power reaches 3
+	errorTxt := "BIG ERROR OH NO"
+	w.AddSystem(func(world *ecs.World, queue *ecs.TransactionQueue, _ *ecs.Logger) error {
+		panic(errorTxt)
+	})
+	assert.NilError(t, w.LoadGameState())
+	ctx := context.Background()
+
+	defer func() {
+		if panicValue := recover(); panicValue != nil {
```
jerargus, line 62: assert that panicValue is equal to errorTxt, so that if a panic shows up from any other source the test will fail.

jerargus (approved):

PR #268, opened 2023-09-13
feat(cardinal): prevent entity update and create if comp is not registered.

pyrofolium:
Closes: #WORLD-294

Prevents update and create if component isn't registered. 

## Brief Changelog

Just see the changes below. It's small. 

## Testing and Verifying

this is tested. Both with registration and without.

### DIFF
--- a/cardinal/ecs/component.go
+++ b/cardinal/ecs/component.go
@@ -81,6 +81,9 @@ func (c *ComponentType[T]) Update(w *World, id storage.EntityID, fn func(T) T) e
 
 // Set sets component data to the entity.
 func (c *ComponentType[T]) Set(w *World, id storage.EntityID, component T) error {
+	if _, ok := w.nameToComponent[c.Name()]; !ok {
+		return fmt.Errorf("%s is not registered, please register it before updating", c.Name())
+	}
 	entity, err := w.Entity(id)
 	if err != nil {
 		return err
--- a/cardinal/ecs/component.go
+++ b/cardinal/ecs/component.go
@@ -52,6 +52,10 @@ func (c *ComponentType[T]) SetID(id component.TypeID) error {
 	return nil
 }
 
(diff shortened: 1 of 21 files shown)


pyrofolium:
Oh wow this has big failures. Closing this PR.

cardinal/ecs/component.go
technicallyty, line 1: we should also guard calls to `Set` here

technicallyty (commented):
componen

cardinal/ecs/component.go
pyrofolium: This caused another issue in tests involving Set and two worlds. In order to use a component in another world with Set we now have to register it, upon registering it it tries to SetID for that component which will trigger an error because it was already set with the registration on the previous world. @technicallyty

technicallyty: the error is only thrown by SetID if its set to a different ID than before, so we can fix it by registering `onePower` first:

L139: `assert.NilError(t, twoWorld.RegisterComponents(onePower, twoPower))`

pyrofolium: yeah, but what about for testing, usage of components across worlds in general?

pyrofolium: @technicallyty what do you think if I make it so that ID creation accounts for ids already created. So basically it scans all the comps for ids and just skips over the ones that are already used. I'll do that.

pyrofolium: changed.

technicallyty: the pattern of re-using ComponentType's across worlds is a mistake imo - it won't ever happen in practice. i think we should avoid changing code to support it

pyrofolium: yeah will revert

technicallyty (approved):
LGTM

PR #273, opened 2023-09-14
refactor(cardinal): Remove StorageAccessor

jerargus:
Closes: #WORLD-319 and #WORLD-129

Remove StorageAccessor. It's just a wrapper around a subset of the world.store field. 

As we continue with refactoring the ECS/storage packages it will be important to know when we require the ecs.World object when when we require a storage object. This PR makes that requirement more clear.

## Testing and Verifying

Existing tests will verify no functionality has changed.

### DIFF
--- a/cardinal/ecs/query.go
+++ b/cardinal/ecs/query.go
@@ -35,9 +35,8 @@ type QueryCallBackFn func(storage.EntityID) bool
 // Each iterates over all entities that match the query.
 // If you would like to stop the iteration, return false to the callback. To continue iterating, return true.
 func (q *Query) Each(w *World, callback QueryCallBackFn) {
-	accessor := w.StorageAccessor()
-	result := q.evaluateQuery(w, &accessor)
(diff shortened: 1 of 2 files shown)


namrapatel (approved):
To make sure I understand, the point here is that if we decrease the amount of places where we require the `world` but rather require some specific object in the world specifically, we can narrow in on locations where the entire `world` object is required?

jerargus:
> To make sure I understand, the point here is that if we decrease the amount of places where we require the `world` but rather require some specific object in the world specifically, we can narrow in on locations where the entire `world` object is required?

@namrapatel Yes. Ultimately, if we have our global "world" object, I do not want that to be passed around. Instead, "world.store" should be passed to things that are going to read/write to storage.

This "StorageAccessor" felt weird because all the places we were actually using it we were already passing in the entire "world". 

Eventually, i want (as an example) this signature:
```
Each(w *World, callback QueryCallBackFn)
```
To look like this:
```
Each(s *StorageReference, callback QueryCallBackFn)
```

PR #274, opened 2023-09-14
refactor(cardinal): move tx queue out of ecs

jerargus:
Closes: #WORLD-318

Move the TransactionQueue struct out of the ecs package. We want to reduce the number of places that need to depend on the ecs package.

This touches a lot of files, but most of them are just `s/ecs.TransactionQueue/transaction.TxQueue`. 

## Testing and Verifying

Existing tests verify there are no functional changes.

### DIFF
--- a/cardinal/ecs/transaction/transaction.go
+++ b/cardinal/ecs/transaction/transaction.go
@@ -1,11 +1,51 @@
 package transaction
 
 import (
+	"sync"
+
 	"github.com/invopop/jsonschema"
 	"pkg.world.dev/world-engine/sign"
 )
 
-type TxMap map[TypeID][]TxAny
+type TxQueue struct {
+	m   txMap
+	mux *sync.Mutex
+}
+
+func NewTxQueue() *TxQueue {
+	return &TxQueue{
+		m:   txMap{},
+		mux: &sync.Mutex{},
+	}
+}
+
+func (t *TxQueue) AddTransaction(id TypeID, v any, sig *sign.SignedPayload) TxHash {
+	t.mux.Lock()
+	defer t.mux.Unlock()
+	txHash := TxHash(sig.HashHex())
+	t.m[id] = append(t.m[id], TxAny{
+		TxHash: txHash,
+		Value:  v,
+		Sig:    sig,
+	})
+	return txHash
+}
+
+func (t *TxQueue) CopyTransaction() *TxQueue {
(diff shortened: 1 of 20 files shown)


cardinal/ecs/transaction/transaction.go
```diff
@@ -1,11 +1,51 @@
 package transaction
```
namrapatel, line 1: I'm not all that familiar with this part of the codebase, but is it weird that we have `cardinal/ecs/transaction.go`, `cardinal/ecs/transaction/transaction.go`, and also `cardinal/transaction.go`?

technicallyty (approved):
LGTM! can you update the PR title?

cardinal/ecs/transaction/transaction.go
```diff
@@ -1,11 +1,51 @@
...
 
 import (
+	"sync"
+
 	"github.com/invopop/jsonschema"
 	"pkg.world.dev/world-engine/sign"
 )
 
-type TxMap map[TypeID][]TxAny
+type TxQueue struct {
```
namrapatel, line 10: The changes themselves look OK, but not sure I understand this PR:

1. Why is moving the TxQueue from the `ecs` root dir, into an `ecs/transaction` dir cleaner? 
2. Do we have a problem with the `World` object or both the `World` object and the `ecs` package too? Apologies if I missed something but I don't recall what the conflict we found with the `ecs` package was

namrapatel (approved):

pyrofolium (approved):
LGTM

cardinal/ecs/transaction/transaction.go
jerargus: I see problems with both the World object and the ecs package. They are both doing too much. Creating a new package to implement some feature is difficult because it 1) needs to import the ECS package to do anything interesting and 2) needs to be imported BY the ECS package to be used. To satisfy both these requirements we're forced to just implement it in the ECS package. Ideally, ECS will import some new feature package, and that new feature package will import a handful of sub-package so it can do its job.

cardinal/ecs/transaction/transaction.go
jerargus: I think that's ok. Each package (`cardinal`, `cardina/ecs` and `cardinal/ecs/transaction`) have some transaction related code.

PR #276, opened 2023-09-15
refactor(cardinal): Create store manager that will handle the modification of components …

jerargus:
…in a world

Currently, code for component management (Get/Set/AddToEntity/RemoveFromEntity) was split between the *World object and storage.Entity. 

storage.Entity is a very simple struct that doesn't need business logic attached to it. In addition, the Entity was dispatching some of the logic *World object (hidden behind a WorldAccessor interface).

This PR takes the component management code out of Entity and World and puts it into a "StoreManager". Eventually this StoreManager code will live in a standalone package. The dependency graph will be:

ecs -> store_manager -> component, filter, storage.

instead of the current: ecs <--> storage.

The eventual goal is to completely remove the WorldAccessor interface.

Note: The code in store_manager.go is pretty much exactly the removed code from world.go and storage/entity.go.

## Testing and Verifying

No functional changes are intended, so existing unit tests should validate that everything is working correctly.

### DIFF
--- a/cardinal/ecs/component.go
+++ b/cardinal/ecs/component.go
@@ -54,15 +54,15 @@ func (c *ComponentType[T]) SetID(id component.TypeID) error {
 
 // Get returns component data from the entity.
 func (c *ComponentType[T]) Get(w *World, id storage.EntityID) (comp T, err error) {
-	entity, err := w.Entity(id)
+	value, err := w.StoreManager().GetComponentForEntity(c, id)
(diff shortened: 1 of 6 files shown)


namrapatel (approved):
I like this, this pretty closely matches a previous ECS implementation I was familiar with, except they had `Store` but no `StoreManager`, these functions were simply exposed from within the `Store` object

PR #277, opened 2023-09-19
refactor(cardinal): Finish removing the WorldAccessor interface

jerargus:
Closes: #WORLD-321

- Finish removing the WorldAccessor interface
- Move move the Create/Remove code for entities into the store_manager.go source file.
- Make an alternative way to check if a component is in a slice of components. This is a preparation step for the eventual removal of the Layout struct.

## Testing and Verifying

Existing unit tests verify that there are no functional changes.

### DIFF
--- a/cardinal/ecs/storage/layout.go
+++ b/cardinal/ecs/storage/layout.go
@@ -31,6 +31,8 @@ func (l *Layout) Components() []component.IComponentType {
 }
 
 // HasComponent returns true if the ArchLayout has the given component type.
+//
+// Deprecated: Use component.Contains instead
 func (l *Layout) HasComponent(componentType component.IComponentType) bool {
 	for _, ct := range l.ComponentTypes {
 		if ct == componentType {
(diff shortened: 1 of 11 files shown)


cardinal/ecs/storage/layout.go
```diff
@@ -31,6 +31,8 @@ func (l *Layout) Components() []component.IComponentType {
 }
 
 // HasComponent returns true if the ArchLayout has the given component type.
+//
+// Deprecated: Use component.Contains instead
 func (l *Layout) HasComponent(componentType component.IComponentType) bool {
```
namrapatel, line 36: Are we gonna remove this later or is the plan to leave it in anyway?

namrapatel (approved):

cardinal/ecs/storage/layout.go
jerargus: I plan on taking out the Layout struct entirely, so this will end up being taken out.

PR #278, opened 2023-09-19
refactor(cardinal): update game loop to tick exactly at the specified interval

technicallyty:

before, we our ticks would run at `loopInterval` + the time taken to execute the systems. 

this PR introduces a change that calculates the time it takes to execute the systems, and then sleeps if there is remaining time in the specified loop interval. 

also changes the `panic` if errors occur to a `log.Fatal` so that the entire system shuts down, instead of just the go game loop routine.

## Brief Changelog

- update tick timing logic
- change panic to log.Fatal if error returned in system


namrapatel (approved):
amazing

technicallyty:
Ticker takes care of this automatically actually https://pkg.go.dev/time#NewTicker (h/t @jerargus )

PR #279, opened 2023-09-19
feat(cardinal): prepare swagger routes for starter-template

pyrofolium:
Closes: #WORLD-317

Added new route /query/game/cql for the endpoint route.
Used some macro library to embedd the swagger.yml into the library so it doesn't have to be referenced by path. 
Replaced NewHandler with the swagger definition and retitled the old definition to OldHandler

When this is approved I will tag this with a new version and use it in the starter-template. 

## Testing and Verifying

ran go test ./... in cardinal.

### DIFF
--- a/cardinal/server/server.go
+++ b/cardinal/server/server.go
@@ -326,6 +326,7 @@ func createAllEndpoints(world *ecs.World) (*EndpointsResult, error) {
 	queryEndpoints = append(queryEndpoints, "/query/http/endpoints")
 	queryEndpoints = append(queryEndpoints, "/query/persona/signer")
 	queryEndpoints = append(queryEndpoints, "/query/receipt/list")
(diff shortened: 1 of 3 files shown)


cardinal/server/server.go
```diff
@@ -58,18 +59,27 @@ const (
...
+func NewHandler(w *ecs.World, opts ...Option) (*Handler, error) {
+	h, err := newSwaggerHandlerEmbed(w, opts...)
+	if err != nil {
+		return nil, err
+	}
+	return h, nil
+}
+
+//go:embed swagger.yml
+var swaggerData []byte
```
pyrofolium, line 73: The above is some reflection based thing that automatically pulls swagger.yml into swaggerData.

jerargus (approved):

PR #282, opened 2023-09-20
refactor(cardinal): move simple ids into dedicated packages

jerargus:

Move a handful of simple structs out of the storage package and into dedicated packages. Specifically:
storage.ArchetypeID -> archetype.ID 
storage.ComponentIndex -> component.Index 
storage.EntityID -> entity.ID 
storage.Location -> entity.Location

This ends up touching a lot of files because entity IDs, Archetype IDs, locations, and component indices are used all over the place. 

This is beneficial because it means a lot of storage logic can be moved out of the storage package without maintaining a dependency on storage. 

In addition, some packages that previously imported storage (e.g. cql, evm) no longer have a direct dependency on storage.

## Testing and Verifying

Existing unit tests verify there are no functional changes.


smsunarto:
dumb questionc what is location and index and how are they different than id?

docs/pages/Cardinal/API-Reference/Systems.mdx
```diff
@@ -68,7 +68,7 @@ func AttackSystem(world *cardinal.World, queue *cardinal.TransactionQueue, logge
 		atk := tx.Value()
 
 		// update the component based on the attack
-		err := HealthComponent.Update(world, storage.EntityID(atk.TargetPlayer), func(health Health) Health {
+		err := HealthComponent.Update(world, entity.ID(atk.TargetPlayer), func(health Health) Health {
```
technicallyty, line 71: noting here: wonder if we may need a more complex workflow - i.e. docs should be rebuilt/republished to the web *after* a tagged release, so its always always in sync with changes, since now - the only way these docs make sense is if someone pulls cardinal from main

technicallyty (approved):
LGTM

jerargus:
> dumb questionc what is location and index and how are they different than id?

An Entity ID is a number that uniquely identifies an entity across the World object. Creating new entities will generate new ids, and entity ids are not reused.

A location is a combination of an ArchetypeID and a ComponentIndex. The ArchetypeID represents a unique collection of components and each ArchetypeID is going to have some list of entities that actually belong to that ArchetypeID.

The ComponentIndex is an index into the list of entities associated with an ArchetypeID. It's used to actually store data about the entity in the storage layer.

Put another way: Entity ID uniquely identifies an entity, but tells you nothing about where the data of that entity is stored.

Location is a reference to some component data in the storage layer, but it doesn't tell you what entity it belongs to.

The entity.Entity struct combines both the Entity ID and the Location information.

jerargus:
> dumb questionc what is location and index and how are they different than id?
@smsunarto 


> > dumb questionc what is location and index and how are they different than id?
> 
> An Entity ID is a number that uniquely identifies an entity across the World object. Creating new entities will generate new ids, and entity ids are not reused.
> 
> A location is a combination of an ArchetypeID and a ComponentIndex. The ArchetypeID represents a unique collection of components and each ArchetypeID is going to have some list of entities that actually belong to that ArchetypeID.
> 
> The ComponentIndex is an index into the list of entities associated with an ArchetypeID. It's used to actually store data about the entity in the storage layer.
> 
> Put another way: Entity ID uniquely identifies an entity, but tells you nothing about where the data of that entity is stored.
> 
> Location is a reference to some component data in the storage layer, but it doesn't tell you what entity it belongs to.
> 
> The entity.Entity struct combines both the Entity ID and the Location information.
> 
> 
> 
>

namrapatel (approved):

PR #285, opened 2023-09-20
feat(cardinal): proper shutdown handling

pyrofolium:
Closes: #world-332

Proper shutdown handling. Make sure all requests have finished processing. Then make sure a tick is fully complete before shutting down. 

## Brief Changelog

- Added a shutdown package in server. Server owns world, so for access to both world and server the shutdown manager needs to be located here. Shutdown manager instantiates with a reference to the Server and the World object.
- Added new methods to World and Handler for shutting down server and stopping game loop.  
- Added a new channel to world to help it break out of looping when something else sends it a signal. 

## Testing:
Added two new tests to `server_test.go`


technicallyty (changes requested):
this doesn't seem to be working, i edited your test to sleep; ticks are still logging - also added a counter to the system to verify, and it counted up the amount of times i slept for:

```
func TestShutDownViaMethod(t *testing.T) {
	w := inmem.NewECSWorldForTest(t)
	sendTx := ecs.NewTransactionType[SendEnergyTx, SendEnergyTxResult]("sendTx")
	assert.NilError(t, w.RegisterTransactions(sendTx))
	count := 0
	w.AddSystem(func(world *ecs.World, queue *ecs.TransactionQueue, _ *ecs.Logger) error {
		count++
		return nil
	})
	assert.NilError(t, w.LoadGameState())
	txh := makeTestTransactionHandler(t, w, "./swagger.yml", DisableSignatureVerification())
	ctx := context.Background()
	w.StartGameLoop(ctx, 1*time.Second)
	shutdownObject := NewShutdownManager(w, txh.Handler)
	err := shutdownObject.Shutdown()
	time.Sleep(15 * time.Second)
	assert.NilError(t, err)
	fmt.Println(count)
}
```

pyrofolium:
@technicallyty hmmm let me verify what you're seeing here.

technicallyty:
ah i think its cause when u break in a select, it'll just break out of the select

pyrofolium:
@technicallyty Yeah this test should have failed. Will change it.

pyrofolium:
@technicallyty fixed. 

Additionally I put in some extra atomic state to track whether or not the server and the game loop is running. The tests should fail now if either loop didn't shut down.

cardinal/ecs/world.go
```diff
@@ -565,20 +567,30 @@ func (w *World) StartGameLoop(ctx context.Context, loopInterval time.Duration) {
 		w.Logger.Warn().Msg("No systems registered.")
 	}
 	go func() {
+		w.isGameLoopRunning = true
+		wasEndGameLoopSignalReceived := false
```
technicallyty, line 573: you can actually do away with the `wasEndGameLoopSignalReceived` bool entirely by using labels and breaking on the specific label:

```go
loop:
	for {
		select {
		default:
			break loop
		}
	}
```

cardinal/server/shutdown.go
technicallyty, line 1: can we use zerolog here instead of the fmt.Prints?

smsunarto:
Can we call this instance manager or runtime manager instead? shutdown manager sounds oddly specific

technicallyty (commented):
couple nits, overall looks good! although i'd like to wait for @jerargus to have a look; i have used sleeps in my test before and he had some suggestions around it. im not sure what else u could do here but he might have an idea 🤷🏼‍♂️

technicallyty:
> Can we call this instance manager or runtime manager instead? shutdown manager sounds oddly specific

that's all it does though - manages shutdown signals 😁 it is indeed quite specific

pyrofolium:
@technicallyty The sleep is just there so those loops aren't hammering the atomic value with reads, it actually doesn't matter that much because this is just a test. 

@smsunarto sure, will do tomorrow. 

There's one more change I just thought of. Shutdown needs to block until the loop is completely finished. The reason is if you shutdown and exit the program to quickly the game loop and server loop may not have finished. You need to shutdown and the shutdown needs to block the current thread until all loops have stopped. Will update this tomorrow.

cardinal/server/server_test.go
```diff
@@ -99,6 +99,84 @@ func makeTestTransactionHandler(t *testing.T, world *ecs.World, opts ...Option)
 	}
 }
 
+func TestShutDownViaMethod(t *testing.T) {
+	go func() {
+		select {
+		case <-time.After(10 * time.Second):
```
jerargus, line 105: The select portion of this is redundant.

```
go func() {
    <-time.After(10*time.Second)
    panic("took to long to shutdown"
} 
```
has the same behavior.

cardinal/server/server.go
```diff
@@ -33,7 +35,8 @@ type Handler struct {
 	port                   string
 
 	// plugins
-	adapter shard.WriteAdapter
+	adapter         shard.WriteAdapter
+	isServerRunning atomic.Bool
```
jerargus, line 39: Is this only used to make sure the server is running before tests start?

Is it required? The `makeTestTransactionHandler` helper appears to make a health check endpoint to ensure the server is up and serving traffic before letting any tests continue.

cardinal/server/server_test.go
```diff
@@ -99,6 +99,84 @@ func makeTestTransactionHandler(t *testing.T, world *ecs.World, opts ...Option)
...
+	}() // If this test is frozen then it failed to shut down, create failure with panic.
+	w := inmem.NewECSWorldForTest(t)
+	sendTx := ecs.NewTransactionType[SendEnergyTx, SendEnergyTxResult]("sendTx")
+	assert.NilError(t, w.RegisterTransactions(sendTx))
+	w.AddSystem(func(world *ecs.World, queue *ecs.TransactionQueue, _ *ecs.Logger) error {
+		return nil
+	})
+	assert.NilError(t, w.LoadGameState())
+	txh := makeTestTransactionHandler(t, w, DisableSignatureVerification())
+	for !txh.IsServerRunning() {
```
jerargus, line 117: See my comment above. I believe the makeTestTransactionHandler doesn't return until the server is up and running.

cardinal/ecs/world.go
```diff
@@ -562,14 +569,34 @@ func (w *World) StartGameLoop(ctx context.Context, loopInterval time.Duration) {
...
 		for range time.Tick(loopInterval) {
+			if wasEndGameLoopSignalReceived {
+				break
+			}
 			if err := w.Tick(ctx); err != nil {
 				w.Logger.Panic().Err(err).Msg("Error running Tick in Game Loop.")
 			}
+			select {
+			case <-w.endGameLoopCh:
```
jerargus, line 582: Both time.Tick and w.endGameLoopCh are channels, and in this pattern you don't check for the end of the game loop until after a tick has been run. I'd suggest:
```
ticker := time.Tick(loopInterval)
for {
    select{
        case <-ticker:
            // Run the tick.
        case <-w.endGameLoopCh:
            // cleanup/break out. I like @technicallyty suggestion of using a label to break out of this nested context
    }
}
```

cardinal/server/server_test.go
```diff
@@ -99,6 +99,84 @@ func makeTestTransactionHandler(t *testing.T, world *ecs.World, opts ...Option)
...
+		time.Sleep(1 * time.Millisecond)
+	}
+	shutdownObject := NewShutdownManager(w, txh.Handler)
+	err := shutdownObject.Shutdown()
+	assert.NilError(t, err)
+	//wait for both game loop and server to shut down.
+	for w.IsGameLoopRunning() {
+		time.Sleep(1 * time.Millisecond)
+	}
+	for txh.IsServerRunning() {
```
jerargus, line 134: If you remove the `isServerRunning`, this will need to be replaced with a helper that hits the healthcheck endpoint until it no longer returns a 200.

cardinal/server/shutdown.go
```diff
@@ -0,0 +1,50 @@
...
+import (
+	"fmt"
+	"os"
+	"os/signal"
+	"syscall"
+
+	"pkg.world.dev/world-engine/cardinal/ecs"
+)
+
+type ShutDownManager struct {
```
jerargus, line 12: nit: You have a mix of "ShutDown" and "Shutdown" in this code. It should be consistent. "Shutdown" looks more natural to me, but either is fine.

cardinal/server/server.go
pyrofolium: It's very common for to check if a daemon is "running" even outside of testing. Plenty of things you don't want to do while a daemon is running and plenty of things you do want to do Only when the main loop is cycling. 

Right now it's only used for that application but this is a useful method.

cardinal/server/server_test.go
pyrofolium: Hm I don't believe this is true. What I saw when I asserted IsServerRunning is that it actually failed. You have to wait for it. The atomics I placed are as near as possible to the running loop so it might be your function returns a little before the running loop actually occurs.

cardinal/server/server_test.go
pyrofolium: I think isServerRunning works fine. If the functionality is the same then I think it's ok to keep it. We can talk about it at standup.

cardinal/server/shutdown.go
```diff
@@ -0,0 +1,50 @@
...
+	handler *Handler
+	world   *ecs.World
+}
+
+func NewShutdownManager(world *ecs.World, handler *Handler) ShutDownManager {
```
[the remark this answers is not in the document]
pyrofolium: I don't think this will work for my case here. The reason is because this manager is operating in a very specific way. It has to shutdown the server first. Then wait for that to finish, and then shut down the game loop. 

First all ongoing requests have to be processed. That could fill up the queue with Txs. Then all Txs have to be processed by the game loop. Then We shutdown. So this manager is operating on a very specific two things: Server and World.

cardinal/server/server_test.go
pyrofolium: yeah makes sense.

cardinal/ecs/world.go
pyrofolium: yeah makes sense.

cardinal/ecs/world.go
pyrofolium: Actually I think it can still occur even if it's placed after a tick. Yeah will discuss a bit during standup.

cardinal/server/server_test.go
jerargus: I did some testing and I don't think this works as expected. Nothing cleans up this goroutine, so it continues to track time even after this test has exited. If this test is being run in isolation, it's fine, but if this test is being run with other tests (other tests in the package, or with the `--count` flag), it will trigger after the 10 seconds have elapsed.

I think this helper would do the trick:
```
func setTestTimeout(t *testing.T, timeout time.Duration) {
	if _, ok := t.Deadline(); ok {
		// A deadline has already been set. Don't add an additional deadline.
		return
	}
	success := make(chan bool)
	t.Cleanup(func() {
		success <- true
	})
	go func() {
		select {
		case <-success:
			// test was successful. Do nothing
		case <-time.After(timeout):
			assert.Check(t, false, "test timed out")
		}
	}()
}
```

pyrofolium: Actually the above code didn't terminate after timeout.

pyrofolium: You have to panic if it's in another green thread.

jerargus: Shoot. You could do a pattern like this: https://stackoverflow.com/questions/24929790/how-to-set-the-go-timeout-flag-on-go-test -- but you'll have to adjust your code in the 4 places that you're waiting for the server to start or stop.

jerargus: Or yeah, panic instead of assert.Check.

cardinal/server/server.go
pyrofolium: As agreed in stand up I will be removing this.

cardinal/server/server_test.go
pyrofolium: I did some tests here. You're right I think I'm mistaken on this one.

cardinal/server/server_test.go
pyrofolium: As discussed in the standup I removed isServerRunning and i'm doing the healthcheck thingie here.

cardinal/ecs/world.go
pyrofolium: As discussed in standup I will be following the select pattern above but also I'll need to check if the queue is empty before exiting this loop. If the queue is not empty tick it.

cardinal/server/shutdown.go
pyrofolium: This has been changed to GameManager. @smsunarto mentioned using another less specific name here. I think game manager is the most clear while keeping the functionality general. In the future we can have this object manage all the things that need to coordinate world and server.

cardinal/server/server_test.go
```diff
@@ -100,6 +101,98 @@ func makeTestTransactionHandler(t *testing.T, world *ecs.World, opts ...Option)
...
+	assert.NilError(t, err)
+	assert.Assert(t, !w.IsGameLoopRunning())
+	for {
+		time.Sleep(500 * time.Millisecond)
+		_, err := http.Get("http://localhost:4040/health")
+		if err != nil {
+			break //server is supposed to be shutdown.
+		}
+	}
+
```
technicallyty, line 153: can't you just do:

```go
	_, err = http.Get("http://localhost:4040/health")
	assert.Check(t, err != nil)
```

at this point?

cardinal/server/server_test.go
```diff
@@ -100,6 +101,98 @@ func makeTestTransactionHandler(t *testing.T, world *ecs.World, opts ...Option)
...
+	}
+	for {
+		time.Sleep(500 * time.Millisecond)
+		_, err := http.Get("http://localhost:4040/health")
+		if err != nil {
+			break //server is supposed to be shutdown.
+		}
+
+	}
+
```
technicallyty, line 193: same here

cardinal/server/server_test.go
pyrofolium: The for loop here will call the url multiple times. 

Sometimes there's no error. Sometimes there is an error. In both cases it is correct behavior. I am waiting for the server to shutdown so there is no incorrect behavior here. 

I am more using the timeout thing above to see if there's a problem.

cardinal/server/server_test.go
pyrofolium: same reply as above

cardinal/server/server_test.go
pyrofolium: This is the thing that replaces `txh.IsServerRunning()` btw. Similar pattern.

technicallyty: have you been able to generate the case where there isn't an error? i've ran it multiple times and always get an error - which makes sense to me, it seems to me this part of the code can't be reached unless the server has been shut down

pyrofolium: I think I see what you're getting at now. You're saying because we expect shutdown to block until everything stops running the server should always error. Yeah I see what you're saying.

cardinal/server/shutdown.go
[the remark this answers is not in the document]
pyrofolium: @jerargus forgot to address this in standup. But I'm just going to leave it for the reasoning above.

cardinal/server/server_test.go
pyrofolium: @technicallyty I changed lmk what you think.

technicallyty: it should also work for the signal one too, as you spinlock on the atomic boolean:

```go
for w.IsGameLoopRunning() {
		time.Sleep(500 * time.Millisecond)
	}
```

after this, we should be able to assert that the handler is dead, as it is shut down before the game loop

pyrofolium: oh hm good point. If I checked it in a different order this wouldn't be the case. But yeah. ok I'll change it.

pyrofolium: changed @technicallyty

technicallyty (approved):
LGTM

PR #286, opened 2023-09-21
chore(cardinal): remove all code associated with old routes. Removes tests.

pyrofolium:
Closes: #world-315

branch name is wrong!

Removes all code associated with old routes. 

Also modifies some tests that I didn't touch in the last testing round to hit the new routes. 

## Brief Changelog

See above

## Testing and Verifying

Ran go tests looks like it works.


technicallyty (approved):
lgtm

PR #290, opened 2023-09-22
feat(cardinal): Add ability to be notified when a tick has finished

jerargus:
Closes: #WORLD-306

- Verify that a Query can be called inside of a cardinal based system (without needing the ecs package).
- Add a mechanism to manually start ticks, and be notified when ticks are completed.
- remove some code that was causing a for loop to unnecessarily spin.

## Testing and Verifying

Added a test to the cardinal package to demonstrate the usage of Query inside a system, as well as the new mechanism for starting a tick and blocking until the tick is done.

### DIFF
--- a/cardinal/cardinal_test.go
+++ b/cardinal/cardinal_test.go
@@ -1,0 +1,41 @@
+package cardinal_test
+
+import (
+	"testing"
+	"time"
+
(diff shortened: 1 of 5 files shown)


pyrofolium (approved):
Lgtm

PR #291, opened 2023-09-25
feat(cardinal): got rid of reflection in cql package.

pyrofolium:
Closes: #WORLD-311

CQL routes use reflection to work. Get this to work without reflection. 

## Brief Changelog

- Some formatting on imports
- Use type assertion to grab a new method on the concrete component type. 

## Testing and Verifying

Ran the existing unit tests.

### DIFF
--- a/cardinal/ecs/component.go
+++ b/cardinal/ecs/component.go
@@ -1,6 +1,7 @@
 package ecs
 
 import (
+	"encoding/json"
 	"fmt"
 	"reflect"
 	"strconv"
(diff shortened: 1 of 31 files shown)


cardinal/ecs/component.go
```diff
@@ -12,6 +13,10 @@ import (
 	"pkg.world.dev/world-engine/cardinal/ecs/filter"
 )
 
+type IGettableRawJsonFromEntityId interface {
+	GetRawJson(w *World, id entity.ID) (json.RawMessage, error)
+}
```
jerargus, line 18: Include this:

```
var _ IGettableRawJsonFromEntityId = &ComponentType[int]{}
```

To ensure at compile time that the ComponentType implements that interface.

jerargus (approved):

cardinal/ecs/component.go
pyrofolium: Will do. Question: I don't completely the risk here if I don't do this? because it did pass the tests without this.

technicallyty: the risk from not doing it isn't as large in this case - however its a nice visual indicator as to what type implements the interface.

PR #293, opened 2023-09-25
fix(cardinal): submit all transactions to chain

technicallyty:

persona-related transactions were not being submitted to the chain. this PR fixes that. some various cleanups as well. also updates functions that previously required the `Handler` as an argument, to have it as a receiver instead.

## Brief Changelog

- renames gamemanager.go -> game_manager.go
- deletes `handler.go`
- moves some persona related stuff to persona.go
- removes old handler code from persona.go
- move query handler related code to its own file, `query.go`
- change instances where handler was being passed in as an argument to use it as a receiver instead.
- move tx handler related code to its own file, `tx.go`
- change receiver from `t` to `handler`

## Testing and Verifying

ran cardinal/server unit tests


cardinal/server/query.go
```diff
@@ -0,0 +1,163 @@
...
+			}
+			resultElement := cql.QueryResponse{
+				id,
+				make([]json.RawMessage, 0),
+			}
+
+			// The way our framework is set up it's not designed to retrieve components dynamically at runtime.
+			// As a result we have to use reflection which is generally bad and expensive.
+			for _, c := range components {
+				val := reflect.ValueOf(c)
```
jerargus, line 128: This will change very soon to not use reflection.

pyrofolium (approved):
LGTM

jerargus (approved):

pyrofolium:
Maybe add some tests to verify the success of the submission from adapter

pyrofolium:
> > Maybe add some tests to verify the success of the submission from adapter from the http endpoint level. But I don't know how feasible that is.
> 
> added a test - thought we had one before, mb! 🙈

Dunno, maybe I nuked that test when I was adding the new routes. But I believe this bug was in the code even before I added my stuff? Not sure.

PR #294, opened 2023-09-26
feat(cardinal): add new health endpoint

pyrofolium:
Closes: #world-350

Add a health endpoint to cardinal. This endpoint can be augmented in the future with more things. Should also mention this health endpoint is required for the CLI project.  

## Brief Changelog

Added entry into swagger.
Added new health.go file

## Testing and Verifying

Added new test to server_test.go

### DIFF
--- a/cardinal/server/health.go
+++ b/cardinal/server/health.go
@@ -3,17 +3,18 @@ package server
 import (
 	"github.com/go-openapi/runtime"
 	"github.com/go-openapi/runtime/middleware/untyped"
-	"pkg.world.dev/world-engine/cardinal/ecs"
 )
 
-func (handler *Handler) registerHealthHandlerSwagger(world *ecs.World, api *untyped.API) error {
+type HealthResponse struct {
+	IsServerRunning   bool `json:"is_server_running"`
+	IsGameLoopRunning bool `json:"is_game_loop_running"`
+}
+
+func (handler *Handler) registerHealthHandlerSwagger(api *untyped.API) error {
 	healthHandler := runtime.OperationHandlerFunc(func(params interface{}) (interface{}, error) {
-		res := struct {
-			IsServerRunning   bool `json:"is_server_running"`
-			IsGameLoopRunning bool `json:"is_game_loop_running"`
-		}{
+		res := HealthResponse{
 			true, //see http://ismycomputeron.com/
-			world.IsGameLoopRunning()}
+			handler.w.IsGameLoopRunning()}
 		return res, nil
 	})
 	api.RegisterOperation("GET", "/health", healthHandler)
--- a/cardinal/server/health.go
+++ b/cardinal/server/health.go
@@ -1,0 +1,21 @@
+package server
+
+import (
+	"github.com/go-openapi/runtime"
+	"github.com/go-openapi/runtime/middleware/untyped"
+	"pkg.world.dev/world-engine/cardinal/ecs"
+)
+
+func (handler *Handler) registerHealthHandlerSwagger(world *ecs.World, api *untyped.API) error {
+	healthHandler := runtime.OperationHandlerFunc(func(params interface{}) (interface{}, error) {
+		res := struct {
+			IsServerRunning   bool `json:"is_server_running"`
+			IsGameLoopRunning bool `json:"is_game_loop_running"`
+		}{
+			true, //see http://ismycomputeron.com/
+			world.IsGameLoopRunning()}
+		return res, nil
+	})
+	api.RegisterOperation("GET", "/health", healthHandler)
+	return nil
+}
--- a/cardinal/server/server_test.go
+++ b/cardinal/server/server_test.go
@@ -118,7 +118,7 @@ func setTestTimeout(t *testing.T, timeout time.Duration) {
 }
 
 func TestHealthEndpoint(t *testing.T) {
-	//setTestTimeout(t, 10*time.Second)
+	setTestTimeout(t, 10*time.Second)
 	w := inmem.NewECSWorldForTest(t)
 	sendTx := ecs.NewTransactionType[SendEnergyTx, SendEnergyTxResult]("sendTx")
 	assert.NilError(t, w.RegisterTransactions(sendTx))
@@ -130,15 +130,11 @@ func TestHealthEndpoint(t *testing.T) {
 	resp, err := http.Get("http://localhost:4040/health")
 	assert.NilError(t, err)
 	assert.Equal(t, resp.StatusCode, 200)
-	body := make(map[string]bool)
-	err = json.NewDecoder(resp.Body).Decode(&body)
+	var healthResponse HealthResponse
+	err = json.NewDecoder(resp.Body).Decode(&healthResponse)
 	assert.NilError(t, err)
-	v, ok := body["is_server_running"]
-	assert.Assert(t, ok)
-	assert.Assert(t, v)
-	v, ok = body["is_game_loop_running"]
-	assert.Assert(t, ok)
-	assert.Assert(t, !v) //game loop is not running!
+	assert.Assert(t, healthResponse.IsServerRunning)
+	assert.Assert(t, !healthResponse.IsGameLoopRunning)
 	ctx := context.Background()
 	w.StartGameLoop(ctx, time.Tick(1*time.Second), nil)
 	isGameLoopRunning := false
@@ -147,12 +143,9 @@ func TestHealthEndpoint(t *testing.T) {
 		resp, err = http.Get("http://localhost:4040/health")
 		assert.NilError(t, err)
 		assert.Equal(t, resp.StatusCode, 200)
-		err = json.NewDecoder(resp.Body).Decode(&body)
-		v, ok = body["is_server_running"]
-		assert.Assert(t, ok)
-		assert.Assert(t, v)
-		isGameLoopRunning, ok = body["is_game_loop_running"]
-		assert.Assert(t, ok)
+		err = json.NewDecoder(resp.Body).Decode(&healthResponse)
+		assert.Assert(t, healthResponse.IsServerRunning)
+		isGameLoopRunning = healthResponse.IsGameLoopRunning
 	}
 	gameObject := NewGameManager(w, txh.Handler)
 	err = gameObject.Shutdown()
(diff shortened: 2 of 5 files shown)


cardinal/server/health.go
```diff
@@ -0,0 +1,21 @@
+package server
+
+import (
+	"github.com/go-openapi/runtime"
+	"github.com/go-openapi/runtime/middleware/untyped"
+	"pkg.world.dev/world-engine/cardinal/ecs"
+)
+
+func (handler *Handler) registerHealthHandlerSwagger(world *ecs.World, api *untyped.API) error {
```
technicallyty, line 9: the `handler` receiver already has a ref to `World`, so can remove it from the args here

cardinal/server/server_test.go
```diff
@@ -122,8 +117,51 @@ func setTestTimeout(t *testing.T, timeout time.Duration) {
 	}()
 }
 
+func TestHealthEndpoint(t *testing.T) {
+	//setTestTimeout(t, 10*time.Second)
```
technicallyty, line 121: should this still be commented?

cardinal/server/server_test.go
```diff
@@ -122,8 +117,51 @@ func setTestTimeout(t *testing.T, timeout time.Duration) {
...
+	assert.Equal(t, resp.StatusCode, 200)
+	body := make(map[string]bool)
+	err = json.NewDecoder(resp.Body).Decode(&body)
+	assert.NilError(t, err)
+	v, ok := body["is_server_running"]
+	assert.Assert(t, ok)
+	assert.Assert(t, v)
+	v, ok = body["is_game_loop_running"]
+	assert.Assert(t, ok)
+	assert.Assert(t, !v) //game loop is not running!
```
technicallyty, line 141: this test would probably be nicer if we use a concrete struct for the response type, rather than inlining it in the handler funciton.

that way we can just directly `json.Unmarshal` on the type and don't need the extra asserts that the values exist in a map

cardinal/server/health.go
```diff
@@ -0,0 +1,21 @@
...
+	healthHandler := runtime.OperationHandlerFunc(func(params interface{}) (interface{}, error) {
+		res := struct {
+			IsServerRunning   bool `json:"is_server_running"`
+			IsGameLoopRunning bool `json:"is_game_loop_running"`
+		}{
```
technicallyty, line 14: lets define this instead of inlining

cardinal/server/server_test.go
```diff
@@ -122,8 +117,51 @@ func setTestTimeout(t *testing.T, timeout time.Duration) {
...
+	ctx := context.Background()
+	w.StartGameLoop(ctx, time.Tick(1*time.Second), nil)
+	isGameLoopRunning := false
+	for !isGameLoopRunning {
```
technicallyty, line 145: the test passes for me without this loop, do we actually need it?

cardinal/server/server_test.go
```diff
@@ -122,8 +117,51 @@ func setTestTimeout(t *testing.T, timeout time.Duration) {
...
+		err = json.NewDecoder(resp.Body).Decode(&body)
+		v, ok = body["is_server_running"]
+		assert.Assert(t, ok)
+		assert.Assert(t, v)
+		isGameLoopRunning, ok = body["is_game_loop_running"]
+		assert.Assert(t, ok)
+	}
+	gameObject := NewGameManager(w, txh.Handler)
+	err = gameObject.Shutdown()
+	assert.NilError(t, err)
```
technicallyty, line 159: do we need this stuff here if we're not testing the responses after shutdown?

cardinal/server/server_test.go
pyrofolium: Yes. StartGameLoop does not block it launches a goroutine.

cardinal/server/server_test.go
pyrofolium: Doesn't hurt to have this. It's proper shutdown anyway. But yes there is a reason why I put this here. 

I put this here because I think there's a slight bug in our test code. There's a lingering side effect that occurs in our code. 

It may not be reproducible on your computer but what I see happening is that if I remove this shutdown code another test will begin failing at the storage layer. You might be able to repo and see this error: `{"level":"panic","error":"dial tcp 127.0.0.1:50596: connect: connection refused","time":"2023-09-25T21:03:57-07:00","message":"Error running Tick in Game Loop."}`. miniredis is basically doing some weird shit here as that port changes. 

If I run the tests individually it works, if I run them together we have a problem....

What I think is going on is we aren't properly shutting down something related to miniredis. The real redis likely handles all of this better so this is a subtle error related to tests and outside of the scope of this ticket. 

I have this on my mind. Likely will make another issue for it soon, but it's not related to what I'm doing here. Calling the shutdown code temporarily bypasses the error on my end.

cardinal/server/server_test.go
pyrofolium: Yeah sure, I don't mind changing all of this.

cardinal/server/server_test.go
technicallyty: ah i see. we might benefit from using that [TestSuite](https://pkg.go.dev/github.com/stretchr/testify/suite) thingy. that way we can define pre-post test run functions to handle these things, and don't have to remember to do it each test. but we can just open another issue for this as you said its OOS here

pyrofolium: The fix is likely exposing the redis Shutdown method.

cardinal/server/health.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/go-openapi/runtime/middleware/untyped"
+)
+
+type HealthResponse struct {
+	IsServerRunning   bool `json:"is_server_running"`
+	IsGameLoopRunning bool `json:"is_game_loop_running"`
+}
+
+func (handler *Handler) registerHealthHandlerSwagger(api *untyped.API) error {
+	healthHandler := runtime.OperationHandlerFunc(func(params interface{}) (interface{}, error) {
```
jerargus, line 14: why not just inline this to:
api.RegisterOperation("GET", "/health", runtime.OperationHandlerFunc(...)

cardinal/server/server_test.go
```diff
@@ -146,10 +177,12 @@ func TestShutDownViaMethod(t *testing.T) {
 	assert.Assert(t, !w.IsGameLoopRunning())
 	_, err = http.Get("http://localhost:4040/health")
 	assert.Check(t, err != nil)
+	err = txh.Close()
+	assert.NilError(t, err)
 }
 
 func TestShutDownViaSignal(t *testing.T) {
-	setTestTimeout(t, 10*time.Second) // If this test is frozen then it failed to shut down, create a failure with panic.
+	//setTestTimeout(t, 10*time.Second) // If this test is frozen then it failed to shut down, create a failure with panic.
```
technicallyty, line 185: do we need to uncomment this? if not lets just remove it

technicallyty (approved):
LGTM, one question

cardinal/server/server_test.go
jerargus: I suspect the problem is 2 instances of miniredis are trying to bind to the same port. I agree that exposing a redis shutdown is the right thing (eventually, not necessarily in this PR)

cardinal/server/health.go
```diff
@@ -0,0 +1,22 @@
...
+		res := HealthResponse{
+			true, //see http://ismycomputeron.com/
+			handler.w.IsGameLoopRunning()}
```
jerargus, line 17: An alternative pattern here would be to return a 503 if handler.w.IsGameLoopRunning() is false. In this case the client wouldn't have to parse the response body to know the status of the health check.

I'm fine with either pattern.

cardinal/server/server_test.go
```diff
@@ -122,8 +117,44 @@ func setTestTimeout(t *testing.T, timeout time.Duration) {
 	}()
 }
 
+func TestHealthEndpoint(t *testing.T) {
+	setTestTimeout(t, 10*time.Second)
+	w := inmem.NewECSWorldForTest(t)
+	sendTx := ecs.NewTransactionType[SendEnergyTx, SendEnergyTxResult]("sendTx")
```
jerargus, line 123: this sendTx and the System blow aren't adding anything to the test. They should be removed.

cardinal/server/server_test.go
jerargus, line 185: You want to leave this timeout in the test, right?

pyrofolium: Yes we do, that's suppose to be there, it's just removed for testing. 

If you look at the test not all tests are just asserts. Some tests are just spinning and a passing test is if it breaks out of those loops. A failing test would spin forever. The setTestTimeout places an assert on those spinlocks. Basically if a test spins for > 10s I classify it as an error and trigger a panic.

cardinal/server/health.go
pyrofolium: it's following the pattern as all the previous endpoints. If we add more logic to this endpoint this will be the better format imo.

cardinal/server/server_test.go
pyrofolium: yes

cardinal/server/health.go
pyrofolium: I think it's quite likely that this endpoint will have more information with it related to the status of world-engine. Actually health/ might not be a good name as it's too specific. Will change it when I add more stuff.

cardinal/server/server_test.go
```diff
@@ -122,8 +117,44 @@ func setTestTimeout(t *testing.T, timeout time.Duration) {
...
+	err = gameObject.Shutdown()
+	assert.NilError(t, err)
+
+}
+
 func TestShutDownViaMethod(t *testing.T) {
-	//setTestTimeout(t, 10*time.Second) // If this test is frozen then it failed to shut down, create failure with panic.
+	setTestTimeout(t, 10*time.Second) // If this test is frozen then it failed to shut down, create failure with panic.
 	w := inmem.NewECSWorldForTest(t)
 	sendTx := ecs.NewTransactionType[SendEnergyTx, SendEnergyTxResult]("sendTx")
```
pyrofolium, line 159: Hmm accidentally merged this, will place this in next PR, Transactions on this test should be removed.

PR #296, opened 2023-09-27
feat(cardinal): cardinal.world now owns GameManager

pyrofolium:
Closes: #WORLD-358

Move game manager into cardinal.world so that cardinal.world now has the ability to shutdown. 

## Brief Changelog

- renamed impl to implWorld, because now cardinal.world will own a server, gameManager and the ecs.World. A name like impl doesn't give any information about what that variable is and is even more ambiguose with cardinal.World owning other entities. 
- added a method IsGameRunning for aspects of our library and the user to detect if the game is already running. Game will not start if a game is already running. Just a note here, this method is NOT only used it tests. This method IS used before starting the game. 
- had to duplicate a utility function setTestTimeout. This is technical debt that will need to be handled in another ticket which I commented on here: https://linear.app/arguslabs/issue/WORLD-345/clean-up-cardinalserver-code

## Testing and Verifying

Tested shutdown in cardinal tests.

### DIFF
--- a/cardinal/world.go
+++ b/cardinal/world.go
@@ -3,6 +3,7 @@ package cardinal
 import (
 	"context"
 	"errors"
+	"sync/atomic"
 	"time"
 
 	"github.com/rs/zerolog/log"
@@ -16,7 +17,10 @@ import (
 )
 
 type World struct {
-	impl            *ecs.World
+	implWorld       *ecs.World
+	server          *server.Handler
+	gameManager     *server.GameManager
+	isGameRunning   atomic.Bool
 	tickChannel     <-chan time.Time
 	tickDoneChannel chan<- uint64
 	serverOptions   []server.Option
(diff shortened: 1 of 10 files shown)


namrapatel (approved):

cardinal/world.go
```diff
@@ -16,7 +17,10 @@ import (
 )
 
 type World struct {
-	impl            *ecs.World
+	implWorld       *ecs.World
```
technicallyty, line 20: i think i prefer just having `impl`, as now we have a bit of a stuttering problem here e.g.

```go
// stutter
server := http.HTTPServer{}

// no stutter
server := http.Server{}

```

for example, if an object that contains world as field names that field `world`, we then get `world.implWorld`, which is kinda redundant imo. not a big deal in the grand scheme, but maybe this is a chance to get a consistent codebase policy around stuttering. thoughts? cc @jerargus @namrapatel @pyrofolium

pyrofolium: impl is a bigger problem imo, because what is it? Is it implementing server? Is is implementing world? Is it implementing gameManager? 

Also impl itself is a bad name even without the additional because it's an acronym that can get mistaken for some other meaning. My philosophy has been clarity over aesthetics.

If your want to get rid of the "stuttering" we can change the name, my main issue with impl is that it's not clear at all.

pyrofolium:
### Merge Activity

* **Sep 28, 11:05 AM**: @pyrofolium started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Argus-Labs/world-engine/296).

PR #298, opened 2023-10-03
Do not merge. Moved Every package out of ecs. Tests working. 

pyrofolium:
You should pull this package and just browse the directory to see the changes. The diff can only be confusing. 

1. Created a package called "public".
2. Create an interface called IWorld for ecs.World in "public"
3. Make sure NO OTHER PACKAGE references ecs.World and that everything references public.IWorld.
4. Adjust things to deal with the consequences and fallout of 3. above. 
5. Moved every package out of ecs. 

A lot of things are a bit rough. 

Don't change the title. Make sure the CI prevents merging.


namrapatel (commented):
I didn't read every code change in this PR, but I understood the general change and think that this would be the wrong direction to go in. I think rather than using this kind of "Piping" mechanism we just genuinely need to remove code that should **not** be in `ecs` from there. 

Here are a few examples:

1.  Take `abi`, `transaction`, and `read` as examples, all of those things seem like they are somewhat `server` related but are kept in the `ecs` package, and if they aren't necessarily server related, perhaps there is some **other** package alongside `ecs` and `server` such as `engine` that they should live in. 
2. `persona` should definitely not be inside of `ecs`, it has nothing to do with ECS other than it is written using `ecs`, @technicallyty made a good point in saying that `persona` should be a "plugin" and should be a top level package in `world-engine` alongside `cardinal`, I agree here.
4. The `cardinal.world` package seems like a bad idea to me too. I think that functionality would also benefit from being moved to a `engine` package that lives beside `ecs`.

pyrofolium:
1. The way it's organized right now: abi, transaction, and read live in their own package. 

That means they can be moved anywhere. You can put them wherever you want. Put them in server if it seems like it belongs in server. 

2. persona can be moved out of ecs. It currently lives in it's own package. In fact, Everything can be moved out of ecs and that folder can be deleted the functionality will remain the same. 

3. Cardinal.world is mostly unrelated to this change. I didn't really touch it. I just messed with stuff in ECS.

The short summary of what I did here is I put everything that was in ECS in a package and now you can move all those packages and put them anywhere you want.

namrapatel:
> The main complaint people have right now is that there's this extra folder I created called "public". All circular dependencies that perviously could have existed in packages have been shifted into public. Because public is a single package now, the golang compiler no longer complains about circular dependencies.
> 
> Now with this pattern you can package and move anything and everything where ever you want. Heck I can delete every parent folder in the hierarchy and have everything live as a package inside a singular giant folder called "blah". That's how modular everything is now. But like I said the main complaint is it requires that "public" folder and all the interfaces that live in there.

@pyrofolium Yeah I totally get what you mean, my few bullet points were somewhat unrelated to your specific PR, they kinda just echo my thoughts about how this refactor *actually* should be done. I see what you did with the `public` directory, I see that it will most likely work, but I think the cleaner way to do it is to actually remove stuff from ECS cuz I think **literally half* this stuff doesn't belong in there

technicallyty:
gonna draft this to make our PR list look a lil less noisy

PR #300, opened 2023-10-04
Refactor. Take as many methods as possible that take in "World" and make those functions take in components of world. 

pyrofolium:
Then after doing the above move as many files as possible into independent packages. Move things around to make everything work. I did not create any new interfaces here.


cardinal/ecs/filter/and.go
technicallyty, line 1: while we're moving things around, i think `filter` should actually be a sub-dir of `query`. it feels odd that it's a standalone directory in the root of ecs, when it can't exist without the help of query.

cardinal/ecs/icomponent/icomponent.go
technicallyty, line 1: would prefer if we can find a way to keep all component related things under one roof. not much a fan of having 3 different component packages

cardinal/ecs/read/read.go
technicallyty, line 1: looks like the code went missing 😲

cardinal/ecs/world_namespace/world_namespace.go
technicallyty, line 1: can this just go in the root?

cardinal/ecs/itransaction/itransaction.go
technicallyty, line 1: also would like to figure out a way not to have two diff transaction packages `itransaction` and `transaction`

cardinal/ecs/component_types/component_types.go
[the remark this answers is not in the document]
pyrofolium: Leads to circular dependency.

cardinal/ecs/ecs_abi/abi.go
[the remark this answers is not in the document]
pyrofolium: There's a naming conflict.

cardinal/ecs/read/read.go
pyrofolium: I tried to move it and it didn't work so I moved it back to ecs. This is a relic of that move, please ignore. Will remove it.

cardinal/ecs/icomponent/icomponent.go
pyrofolium: Same the problem is circular dependencies.

cardinal/ecs/ecs_abi/abi.go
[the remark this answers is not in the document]
technicallyty: what is the conflict?

cardinal/ecs/filter/and.go
pyrofolium: Sure I can make this work. One thing I realized about golang though is that subfolders don't signify that filter can only exist with query as it can be imported from other places regardless of where you put it. So doing this has no bearing on actual structure... it's purely aesthetic as far as I know.

I think there may be a way to make a package private to another package? Not sure.

technicallyty: yeah i think the aesthetic benefit will be nice, i like the idea of having `pkg.caridinal.dev/ecs/query/filter` rather than `pkg.cardinal.dev/ecs/filter`

technicallyty: one less folder to look at is quite nice too

cardinal/ecs/world_namespace/world_namespace.go
pyrofolium: Sure, when something is packaged, you can put it anywhere. I believe the location of the package is purely aesthetic? Not entirely sure I just kept in the original ecs folder.

cardinal/ecs/itransaction/itransaction.go
pyrofolium: You probably can but TxHash is the main culprit here I believe that causes a circular dependency with receipt. So TxHash would have to live in it's own package. Aesthetically I feel it's better to put everything in an indirection layer here to prevent future circular dependencies rather than a specific package just for TxHash. Nothing outside of a transaction should know what a transaction is, they should only know about Itransaction.

cardinal/ecs/icomponent/icomponent.go
pyrofolium: oh hmm maybe I can merge this with component_types

cardinal/ecs/component_types/component_types.go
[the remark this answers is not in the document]
pyrofolium: This has been dealt with. Merged this with icomponents. There are two packages related to components. IComponent and component.

cardinal/ecs/icomponent/icomponent.go
pyrofolium: At best we can have two packages same with transaction and component.

cardinal/ecs/ecs_abi/abi.go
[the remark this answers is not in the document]
technicallyty: i would strongly opt for stronger package names, especially if our only reason is to avoid aliasing, which is not an uncommon thing to do in Go 

if we really want to avoid having to import 2 different imports across the stack, we can just create type aliases against the ones we need from the ethereum ABI package, into our ABI package

cardinal/ecs/itransaction/itransaction.go
technicallyty: hm.. maybe we need receipts to be part of txs? since its not really something that stands on its own - there are no receipts without txs

cc @jerargus

cardinal/ecs/world_namespace/world_namespace.go
technicallyty: ah whoops, i should've said `ecs` not `root`, i think it would simplest to have `type Namespace string` in `world.go`

pyrofolium: Can't do your second suggestion... this causes circular dependency. That's why it was removed in the first place. I agree with your policy here, it's stupid to have one alias in it's own package but golang. I can move it back to ecs.

technicallyty:
the intuition im getting after playing around with the OG codebase myself - and this PR, is that we should probably find the root of _why_ this circular deps are happening (we have things that shouldnt be where they are), instead of just shifting things around to different packages. most of the changes here _are_ an improvement over the current state of the repo, and i have a hunch that with a few more reorganizings, we can get incredibly fine tuned package encapsulation without simply separating out things so much.

i think another big thing we should address as part of the refactoring is how we currently have our tests set up. i really dislike the fact that in some packages that should be entirely self-encapsulated, we have full integration test suites that use world, storage, the whole shebang, when we really should only be unit testing against the functionality that package has.

cardinal/ecs/ecs_abi/abi.go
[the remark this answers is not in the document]
pyrofolium: so abi is stronger than ecs_abi? abi is more unclear... an abi of what? anything can be an abi. ecs_abi is the stronger name imo. 

But you know what. I don't care that much lol. Alright I'll alias it.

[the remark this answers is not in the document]
technicallyty: well, since `abi` is under `ecs`, i feel we don't need to tack on another `ecs` prefix for the package, since the import path would then be `..../ecs/abi`

cardinal/ecs/world_namespace/world_namespace.go
technicallyty: I think we should dig deeper into exactly why the circular dep is happening - it looks like query needs to know about namespace, but i don't think thats actually correct.

pyrofolium:
> the intuition im getting after playing around with the OG codebase myself - and this PR, is that we should probably find the root of _why_ this circular deps are happening (we have things that shouldnt be where they are), instead of just things around too much.
> 
> most of the changes here _are_ an improvement over the current state of the repo, and i have a hunch that with a few more reorganizings, we can get incredibly fine tuned package encapsulation without simply separating out things so much.
> 
> i think another big thing we should address as part of the refactoring is how we currently have our tests set up. i really dislike the fact that in some packages that should be entirely self-encapsulated, we have full integration test suites that use world, storage, the whole shebang, when we really should only be unit testing against the functionality that package has.

The main problem is really straight forward. World needs to know about other packages and much of those other packages need to know about each other and world as well. This leads to the main circular dependency we often see. But you have to realize this pattern isn't intrinsically wrong. It's easier for us to think in these terms. That's why you guys modeled it with a circular dependency in the first place. The circular dependency is the more natural and intuitive way to think about the design. 

What you see enforced in golang is an opinion. Someone arbitrarily decided for various reasons that circular dependencies shouldn't exist in go between folders. This flies contrary to human intuition on dependencies and things that exist in real life. There are valid reasons why you would want to restrict circular dependencies, but just because a system has circular things doesn't mean it's "wrong". Don't take golang conventions as dogma, it's a good language but not everything should be followed religiously. 

In short, I don't think it's correct to find a "root cause" for circular dependencies as if there's something intrinsically wrong with the code. "The world needs to know about people and people need to know about world" there is nothing intrinsically wrong with that hierarchy but in golang they enforce the policy: "Either only the world knows about people or people know about the world, pick one." What we're doing is making a choice about how to fit our code within golangs restrictions. It's ultimately an illusion to think that golangs restrictions are the "most" correct way. Ultimately we can have a design discussion around this stuff but it's just going to lead to opinionated proposals because no "correct" way actually exists. What there is, is a mismatch between human intuition on taxonomy and nomenclature and golangs idea of Zero circular dependencies in packages. This will ultimately lead to things that "feel" awkwardly named or organized. But you have to realize this is just ocd talking, it's a cosmetic issue that's mostly unsolvable with golang so developers just endlessly iterate on this problem with horizontal progress. More important is we should get this code to be easy to understand.

cardinal/ecs/itransaction/itransaction.go
jerargus: Sorry, I've been staring at this comment box since you pinged me (an hour ago). I don't know the right answer here. Maybe the way transactions are organized are a good case study for how we should organize our code. My gut here says TxHash should get pushed down to it's own package which depends on nothing. 

I think we're bumping up against two competing patterns. Within a user system a user can set a tx result like this:

```
myCoolTx.SetResult(world, myResult)
```
(we currently do this)
or like this:
```
world.SetTxResult(myCoolTx, myResult)
```
(so far we've chosen not to do this because we lose type checking).

I want to talk more about this tomorrow.

technicallyty:
> In short, I don't think it's correct to find a "root cause" for circular dependencies as if there's something intrinsically wrong with the code. 

i don't agree with this - a circular dependency can be indicative of a larger design problem - using the `world_namespace` problem as an example: sure, we can make a new package, but i think that kind of ignores why there was a circular dep in the first place - query is using namespace for no apparent reason - we can just refactor the query code to not use it

pyrofolium:
> query is using namespace for no apparent reason - we can just refactor the query code to not use it.

This is just refactoring the code to fit golangs restrictions. What if it was an explicit requirement by the user that query MUST use the world namespace? Then how do you deal with this problem? You can for sure get real world requirements that don't fit golangs idea of dependencies. 

There are small things that you can do to make it "fit" better. Small examples here and there but there will be places where it doesn't fit for sure. For example Read. Like transactions I want to organize read into it's own package. But I can't because read needs world and world needs read. So read belongs with world while every other package related to an http route has it's own package? There it is, golang restriction and human intuition mismatch.

cardinal/ecs/itransaction/itransaction.go
pyrofolium: I have no opinion on this. I can just do it.

cardinal/ecs/ecs_abi/abi.go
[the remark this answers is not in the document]
pyrofolium: Sometimes when you're code following you follow the wrong path. And you end up in the wrong file and it's not immediately obvious. 

Within the code you'll see stuff like abi.CallThis() or abi.CallThat()

Now if it was ecs_abi.CallThis() or ecs_abi.CallThat() it's much more clear. I don't know if this has happened to you but it has happened to me with World. ecs.World and cardinal.World.

cardinal/ecs/itransaction/itransaction.go
pyrofolium: I don't think there is a right answer here. I'm finding it hard to express to you guys that any which way we do it is basically arbitrary. We have certain intuitions that pull us in certain directions but this intuition isn't rational, it can often run counter to the way golang wants you to organize things. I'll try to find a way to explain this tomorrow.

technicallyty:
gonna draft this to make our PR list look a lil less noisy

PR #301, opened 2023-10-04
DNM: refactor attempt

namrapatel:
- [x] Create a new package beside `ecs` called engine, move `transaction`, `read`, `storage`, `store`, `abi`, `receipt`, and `log` into there.
    - Note that this removes other files that don't have anything to do with `ecs` as well: `tick_test.go`, `chain_recover_test.go`, `options.go`, and `abi_test.go`
- [ ] Delete `cardinal.World` and refactor `ecs.World` to remove non-ecs related parts and bring them into `engine/engine.go`
- [ ] Remove `persona` from cardinal itself, have it use `ecs` as it normally does, but treat it like a plugin outside of Cardinal 
- [ ] Figure out if/why it makes sense to leave `transaction.go`, `read.go`, `query.go`, etc. in top-level `cardinal`


namrapatel:
I'm personally pretty disappointed with the direction we're taking with the refactor, I think it's missing the forest for the trees in that we are ignoring the fact that 50% of the `ecs` package is completely unrelated to `ecs`, and using a "bottom-up" approach to figuring out what kind of redesign needs to be made. 

Going to close this anyway because I don't think my words currently carry the weight to change the direction we're going, in any meaningful way.

PR #302, opened 2023-10-06
feat(cardinal): Make Component Get and Set methods into functions. 

pyrofolium:
Some people approved the PR, that was just a sample it would never go up. This is a real PR. 

Changed all calls got ComponentType.Get and ComponentType.Set into ecs.GetComponent[](ecs.World) ad ecs.SetComponent....

The only two left are on update, once update is removed then Get and Set will be deleted as well.

Unfortunately even just changing the Get and Set methods was huge so there's more changes to component that will come. May submit PRs for two methods at a time or something along those lines.


jerargus:
Awesome. This is a good start. Two things I'd like to see next.

1) ecs.SetComponent[Whatever] is currently how a user will use SetComponent. I think that could be improved by moving it to a package named "component" so it looks like this: "component.Set[Owner]". This package can't be the existing cardinal/ecs/component because of that IComponent type. As you've previously noted, the actual location is arbitrary, but getting it to a separate package will 1) enforce this pattern of "SetComponent knows about ecs.World but ecs.World does NOT know about SetComponent" and 2) make the actual calls in Systems a little nicer.

2) You added the comment "//Below disapears and should be handled by RegisterComponents.". Let's see the consequences of that. Can the `owner := ecs.NewComponentType[Owner]("owner", ecs.WithDefault(wantOwner))` be completely removed and just handle the setup entirely in RegisterComponents? I image the call will look something like:

```
world.RegisterComponents(Owner{}, MyStructOnce{}, MyStructTwo{})
```

I think we're forced to instantiate these structs for RegisterComponents, but I could be wrong.

pyrofolium:
1. No there are still circular dependencies with the current edits and other methods. This doesn't ultimately doesn't untangle the problem. Most circular dependencies are not direct cycles from package to package. There are intermediary packages. 
2. Get rid of the variadic method call. The Registers are done procedurally. Only one component at a time and it becomes generic passing the struct the same way it's passed with NewComponent.

smsunarto:
> ecs.SetComponent[Whatever] is currently how a user will use SetComponent. I think that could be improved by moving it to a package named "component" so it looks like this: "component.Set[Owner]". This package can't be the existing cardinal/ecs/component because of that IComponent type. As you've previously noted, the actual location is arbitrary, but getting it to a separate package will 1) enforce this pattern of "SetComponent knows about ecs.World but ecs.World does NOT know about SetComponent" and 2) make the actual calls in Systems a little nicer.

I don't think we want to this? Per discussion yesterday, the idea was to lift up SetComponent to a high-level package yes?

cardinal/ecs/component/component.go
```diff
@@ -14,11 +14,15 @@ type (
...
-		// Name returns the name of the component.
-		Name() string
 
-		Decode([]byte) (any, error)
 		Encode(any) ([]byte, error)
+		Decode([]byte) (any, error)
+		INameable
+	}
+
+	INameable interface {
```
smsunarto, line 23: Instead of `Name()` I would just consolidate this with `ID()` that `IComponentType` currently has. 

I would also remove `SetID` since the user would already be setting it via their user-defined method. Also, not sure if possible, but I would prefer to consolidate these two interfaces because they seem to serve the same purpose.

cardinal/ecs/component.go
```diff
@@ -213,3 +214,108 @@ func WithDefault[T any](defaultVal T) ComponentOption[T] {
...
+	}
+	comp, ok = value.(*T)
+	if !ok {
+		return comp, fmt.Errorf("type assertion for component failed: %v to %v", value, c)
+	}
+	return comp, nil
+}
+
+// Set sets component data to the entity.
+func SetComponent[T component.INameable](w *World, id entity.ID, component any) error {
```
smsunarto, line 238: is `component any` intentional? I think this could be `component T`?

pyrofolium:
> > ecs.SetComponent[Whatever] is currently how a user will use SetComponent. I think that could be improved by moving it to a package named "component" so it looks like this: "component.Set[Owner]". This package can't be the existing cardinal/ecs/component because of that IComponent type. As you've previously noted, the actual location is arbitrary, but getting it to a separate package will 1) enforce this pattern of "SetComponent knows about ecs.World but ecs.World does NOT know about SetComponent" and 2) make the actual calls in Systems a little nicer.
> 
> I don't think we want to this? Per discussion yesterday, the idea was to lift up SetComponent to a high-level package yes?

the location is arbitrary. Once you package something up it can go anywhere with zero impact on functionality. It's largely aesthetic. Jeremy is saying that he just wants to put it in a package? But the problem with that is we'll still hit circular dependencies, for example world -> persona -> SetComponent -> world.

cardinal/ecs/component.go
pyrofolium: unintentional will change.

cardinal/ecs/component/component.go
pyrofolium: IComponent previously had a name. I just modularized it into a additional interface so that it can be public for a user. Name is not a new parameter, name is used extensively in the framework. 

I don't think it's good to expose ID to the user because it's a field that's sort of set internally. This has to be a unique  increasing number. And it's set dynamically in register components to ensure correctness.

jerargus:
> > ecs.SetComponent[Whatever] is currently how a user will use SetComponent. I think that could be improved by moving it to a package named "component" so it looks like this: "component.Set[Owner]". This package can't be the existing cardinal/ecs/component because of that IComponent type. As you've previously noted, the actual location is arbitrary, but getting it to a separate package will 1) enforce this pattern of "SetComponent knows about ecs.World but ecs.World does NOT know about SetComponent" and 2) make the actual calls in Systems a little nicer.
> 
> I don't think we want to this? Per discussion yesterday, the idea was to lift up SetComponent to a high-level package yes?

Right, the high-level package is where SetComponent should live. Giving it the package name "component" seems natural. This package would NOT be the existing `world-engine/cardinal/ecs/component`. Maybe it would go in new `world-engine/cardinal/component`.

technicallyty:
> > > ecs.SetComponent[Whatever] is currently how a user will use SetComponent. I think that could be improved by moving it to a package named "component" so it looks like this: "component.Set[Owner]". This package can't be the existing cardinal/ecs/component because of that IComponent type. As you've previously noted, the actual location is arbitrary, but getting it to a separate package will 1) enforce this pattern of "SetComponent knows about ecs.World but ecs.World does NOT know about SetComponent" and 2) make the actual calls in Systems a little nicer.
> > 
> > 
> > I don't think we want to this? Per discussion yesterday, the idea was to lift up SetComponent to a high-level package yes?
> 
> the location is arbitrary. Once you package something up it can go anywhere with zero impact on functionality. It's largely aesthetic. Jeremy is saying that he just wants to put it in a package? But the problem with that is we'll still hit circular dependencies, for example world -> persona -> SetComponent -> world.
> 
> I'm saying keep it in ecs for now, I can move it in a different pr as it's a different problem to tackle.
> 
> Are you saying you want all the SetComponents and methods from transaction and read to go into a single package?

if persona is causing problems, what if we just moved it out to its own go.mod, and had it be an optional tx/sys/read that cardinal devs could add during their registration process?

cardinal/ecs/component/component.go
jerargus: +1 with brian. In addition, this Name() is a part of CQL, and a human readable string is more desirable here.

pyrofolium:
> > > > ecs.SetComponent[Whatever] is currently how a user will use SetComponent. I think that could be improved by moving it to a package named "component" so it looks like this: "component.Set[Owner]". This package can't be the existing cardinal/ecs/component because of that IComponent type. As you've previously noted, the actual location is arbitrary, but getting it to a separate package will 1) enforce this pattern of "SetComponent knows about ecs.World but ecs.World does NOT know about SetComponent" and 2) make the actual calls in Systems a little nicer.
> > > 
> > > 
> > > I don't think we want to this? Per discussion yesterday, the idea was to lift up SetComponent to a high-level package yes?
> > 
> > 
> > the location is arbitrary. Once you package something up it can go anywhere with zero impact on functionality. It's largely aesthetic. Jeremy is saying that he just wants to put it in a package? But the problem with that is we'll still hit circular dependencies, for example world -> persona -> SetComponent -> world.
> > I'm saying keep it in ecs for now, I can move it in a different pr as it's a different problem to tackle.
> > Are you saying you want all the SetComponents and methods from transaction and read to go into a single package?
> 
> if persona is causing problems, what if we just moved it out to its own go.mod, and had it be an optional tx/sys/read that cardinal devs could add during their registration process?

It's not just persona. query too. Each. There could be more. I would tackle it as a separate problem. Just change the api for now. 

My question now is, do we want to commit to this pattern?

pyrofolium:
> > > ecs.SetComponent[Whatever] is currently how a user will use SetComponent. I think that could be improved by moving it to a package named "component" so it looks like this: "component.Set[Owner]". This package can't be the existing cardinal/ecs/component because of that IComponent type. As you've previously noted, the actual location is arbitrary, but getting it to a separate package will 1) enforce this pattern of "SetComponent knows about ecs.World but ecs.World does NOT know about SetComponent" and 2) make the actual calls in Systems a little nicer.
> > 
> > 
> > I don't think we want to this? Per discussion yesterday, the idea was to lift up SetComponent to a high-level package yes?
> 
> Right, the high-level package is where SetComponent should live. Giving it the package name "component" seems natural. This package would NOT be the existing `world-engine/cardinal/ecs/component`. Maybe it would go in new `world-engine/cardinal/component`.

He's talking about Every method that previously took an `*ecs.World` to go into a singular package maybe called `manager`. That means transaction and read too. `Component` wouldn't be a high level package imo.

cardinal/ecs/component/component.go
smsunarto: ok, aligned

cardinal/ecs/component.go
```diff
@@ -213,3 +214,110 @@ func WithDefault[T any](defaultVal T) ComponentOption[T] {
...
+//// If you would like to stop the iteration, return false to the callback. To continue iterating, return true.
+//func EachComponent[T component.INameable](w *World, callback QueryCallBackFn) {
+//	var t T
+//	name := t.Name()
+//	c, ok := w.nameToComponent[name]
+//	if !ok {
+//		panic(fmt.Errorf("%s is not registered, please register it before updating", t.Name()))
+//	}
+//	c.GetQuery().Each(w, callback)
+//}
```
technicallyty, line 287: should we just remove the component query stuff and require users to ONLY use the ecs.Query? this is esentially the same as `ecs.NewQuery(filter.Exact(FooComp).Each(....)`

pyrofolium: I'm cool with this. Yeah let's do that. I'll get rid of these functions unless anyone says otherwise.

technicallyty: agreements? @namrapatel @jerargus @smsunarto

pyrofolium: @technicallyty i'll still do the package stuff in a separate PR. Keeps potential problems segregated and the PRs are easier to read.

jerargus: Yeah, a single blessed way to do queries sounds nice.

jerargus:
> if persona is causing problems, what if we just moved it out to its own go.mod, and had it be an optional tx/sys/read that cardinal devs could add during their registration process?

(Why can't i respond to this comment directly in a thread? Why do I have to quote reply =/)

@technicallyty @pyrofolium That's the ultimate goal of the plugin system. I think taking out persona (and signer, and probably a couple other things) is not going to be straight forward. I'm coming around to the idea of leaving these component functions in `ecs` until we can properly server the connection to the automatically registered components.

In some other thread on this page, I think removing the Query stuff from components is worth doing now (I think it involves straight forward re-writing of test code).

cardinal/ecs/component/component.go
smsunarto, line 23: nit: i dont think INameable communicates that this is something related to components.

how about `IAbstractComponent`? Open to alternatives.

jerargus (approved):

namrapatel (approved):
I like this, makes sense to me :)

cardinal/ecs/component/component.go
pyrofolium: Sure that works.

pyrofolium: Done. We can always change this later as well.

technicallyty: I don't think we should go with IAbstractComponent. too vague imo. I would also like to get away from prefixing all our interfaces with `I`. 

in some places it makes sense, i.e. differenciating the `ComponentType` generic struct and the interface equivalent, but i don't think using it as a first resort is a pattern we should adopt

can we think of some others here? `type NamedComponent interface`?

pyrofolium: idc too much for this one, but I see scotts perspective. The user is creating a component. His ONLY interaction with anything called component is through this interface. IComponent and ComponentType becomes private and the user doesn't use or interact with it anymore.  

Therefore, from the user perspective it makes more sense to call it a Component or something along those lines, from the dev perspective it makes sense to call it Nameable because that interface can literally be used with anything that has a name method NOT just components. 

In fact, because this is user facing it might be better to have this Interface take up the IComponent mantle and have all the other types be called something like IPrivateComponent, or PrivateComponentType as the user should no longer be interacting with these primitives. 

Additionally, prefixing interfaces with I is both a common pattern in other languages and a useful/practical one as Structs and Interfaces are used in the same places interchangeably. What is your reasoning for why you don't want to prefix it with an I?

technicallyty: Agree with everything there 

> Additionally, prefixing interfaces with I is both a common pattern in other languages and a useful/practical one as Structs and Interfaces are used in the same places interchangeably. What is your reasoning for why you don't want to prefix it with an I?

It's not really common in Go. I haven't seen any interfaces provided by Go's std libs that prefix with `I`. https://sweetohm.net/article/go-interfaces.en.html

would prefer to keep this as Go-aligned as possible. If there are compelling reasons to delineate, I'm OK with it, but prefer not to default to `I` prefixes.

pyrofolium: Eh. I disagree with this convention. But of course, you're right, we should follow the convention if it exists because that's what users expect.

cardinal/ecs/component.go
```diff
@@ -225,17 +225,19 @@ func GetComponent[T component.INameable](w *World, id entity.ID) (comp *T, err e
...
+		return nil, fmt.Errorf("type assertion for component failed: %v to %v", value, c)
 	}
+	comp = &t
+
 	return comp, nil
 }
 
 // Set sets component data to the entity.
-func SetComponent[T component.INameable](w *World, id entity.ID, component any) error {
+func SetComponent[T component.INameable](w *World, id entity.ID, component *T) error {
```
jerargus: The final `component *T` should just be `component T`. It should just match the exact type that the user initially passes in.

jerargus (approved):

cardinal/ecs/component.go
pyrofolium: Are you sure? This will lead to a copy. 

If I have: 

`type SomethingWithmanyFields {
    field1 int
    ...
    field99 int
}`

That leads to a full copy. More efficient to pass the pointer.

technicallyty: it'll probably be more common to have a component struct with at most a handful of fields (1-5) - components with _many_ fields would go against ECS design patterns

jerargus: Also, there's nothing stopping a user from using a pointer as the T (which would result in copies by reference). IMO, it's confusing to ask for the user to supply a type and then de-reference it.

cardinal/ecs/component.go
```diff
@@ -213,3 +214,114 @@ func WithDefault[T any](defaultVal T) ComponentOption[T] {
...
+	}
+	val, err := GetComponent[T](w, id)
+	if err != nil {
+		return err
+	}
+	updatedVal := fn(val)
+	return SetComponent[T](w, id, updatedVal)
+}
+
+//// EachComponent iterates over the entities that have the component.
```
technicallyty, line 281: is this commented block being addressed in another PR? i think we're OK with removing component queries and requiring the user to use the query package

technicallyty (approved):
LGTM - noting here we will change the interface name in another PR

cardinal/ecs/component.go
pyrofolium: > it'll probably be more common to have a component struct with at most a handful of fields (1-5) - components with _many_ fields would go against ECS design patterns

Yeah of course, but I'm just pointing out worst case scenario here to make it obvious the issue with the copy.

pyrofolium: > Also, there's nothing stopping a user from using a pointer as the T (which would result in copies by reference). IMO, it's confusing to ask for the user to supply a type and then de-reference it.

But if I don't specify the pointer type in the function parameter And the user doesn't use a pointer as the T, then a pass by COPY becomes mandatory.

jerargus: > We want to make it mandatory that the the only thing passed here is a pointer.

Go supports passing data around by value and passing data around by pointers. There are tradeoffs; passing by pointer isn't always the best option. We want to leave that choice to the user.

pyrofolium: Go supports defining what function is pass by value or pass by pointer. We define the function here and since we know exactly what this function does We know that a pass by value here results in an unnecessary copy. 

Yes the user should have the choice if they defined the function and they know that a pass by copy is necessary. The key difference here is we are the ones defining the function. The user should not be given a choice here since a pass by copy results in a functionally pointless copy. 

Why would a user need a pass by copy here?

cardinal/ecs/component.go
pyrofolium: Yeah this comment block will be addressed in another PR. As i address these issues in each PR I will remove the related comment.

cardinal/ecs/component.go
jerargus: Maybe they want to be confident the passed in value won't be modified. Maybe the struct size ends up being smaller than a pointer. Maybe they've set up methods on their structs to not use pointers and they want to be consistent.

Again, with a signature of:
```
func SetComponent[T component.INameable](w *World, id entity.ID, component T) error {
```
The user is allowed to do:
```
SetComponent[*Foo](world, 1, &Foo{})
```

to get the benefits you're talking about.

pyrofolium: >Maybe the struct size ends up being smaller than a pointer.

The chances of this are too minuscule to be practical imo. More likely the user will be accidentally copying big structs everywhere. 

>Maybe they want to be confident the passed in value won't be modified.

I disagree. We don't modify that value during a set, they can look at the code. I don't think a user should be given an option to not pass a pointer because their nervous about something out code doesn't do. 

>to get the benefits you're talking about.

Users are unlikely to to user`SetComponent[*Foo](...)
`. They will most likely do: SetComponent[Foo](...) as that seems more natural. Then because of this they will always be copying the struct. 

@technicallyty what do you think?

technicallyty: I don't think the user will be passing around big structs - the most common case will be small structs

pyrofolium: @technicallyty the big struct thing is is just an extreme case but this does effect less extreme cases. 

Here's the thing. If Set is called a bunch of times even a component with say 10 fields becomes 10x slower. This magnifies the more set is called. So if Set is called 30 times, or 20 times it becomes 200 copies or 300 copies vs. 30 or 20. There is an O(N) slow down here vs. O(1).

technicallyty: sounds like pointer is faster. might be nice to have benchmarks to affirm later on

pyrofolium: @jerargus I'm gonna go with the majority votes for now. We can always change this back later.

PR #303, opened 2023-10-06
feat(cardinal): allow for looping over transactions

jerargus:
Closes: #WORLD-365

Add an ability to loop over a number of transactions, and using the returned result/error to set the transaction receipts result/error.

## Testing and Verifying

Unit test added


technicallyty (approved):
( ͡ʘ ͜ʖ ͡ʘ)

PR #305, opened 2023-10-10
chore(cardinal): refactor remove all component set/get/update methods. Fully replaces with functions. 

pyrofolium:
Closes: #390 (branch name is wrong, ignore)

removes get and set and update methods, replaces them with functions. 

Hmmm not understanding why this PR doesn't have a diff. 

## Testing and Verifying

Ran tests


namrapatel (approved):
LGTM

PR #306, opened 2023-10-10
chore(cardinal): change removeFrom addto.

pyrofolium:
Closes: #391

Change RemoveFrom and AddTo methods to functions. 

## Testing and Verifying

Rewrote tests that used these methods.

### DIFF
--- a/cardinal/ecs/component.go
+++ b/cardinal/ecs/component.go
@@ -85,13 +85,23 @@ func (c *ComponentType[T]) MustFirst(w *World) entity.ID {
 	return id
 }
 
-// RemoveFrom removes this component from the given entity.
-func (c *ComponentType[T]) RemoveFrom(w *World, id entity.ID) error {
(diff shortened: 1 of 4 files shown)


jerargus (approved):

cardinal/ecs/component.go
```diff
@@ -85,13 +85,23 @@ func (c *ComponentType[T]) MustFirst(w *World) entity.ID {
 	return id
 }
 
-// RemoveFrom removes this component from the given entity.
-func (c *ComponentType[T]) RemoveFrom(w *World, id entity.ID) error {
+func RemoveFromComponent[T component.IAbstractComponent](w *World, id entity.ID) error {
```
namrapatel, line 88: Think we need to go `RemoveComponentFrom` as we discussed, same for the Add one

namrapatel (changes requested):

PR #307, opened 2023-10-11
feat(relay): add nakama module

technicallyty:

adds nakama from starter-game-template into the world engine repo so that we may maintain it as a canonical relayer package. this will allow us to make changes, and upload new docker images as part of CI for devs to pull into their game shard projects.

## Testing and Verifying
copy pasta


namrapatel (changes requested):
Not sure if I'm missing something here, but seems like it would be nice if we didn't have `relay/nakama` but just offered `relay` as a package itself?

jerargus:
> Not sure if I'm missing something here, but seems like it would be nice if we didn't have `relay/nakama` but just offered `relay` as a package itself?

I don't think this will be offered up as a package that any end-users will actually import/use directly. Instead, they will end up using some pre-built docker container that is based on this code. They should never actually see the `relay/nakama` path.

jerargus (approved):
LGTM. Will you be removing this code from the starter-game-template as well?

namrapatel (approved):

technicallyty:
> LGTM. Will you be removing this code from the starter-game-template as well?

once we get CI setup to automagically push docker images yes. will be working with @heronimus to set that up

PR #308, opened 2023-10-11
refactor(cardinal): update AuthorizePersonaAddressSystem to use tx iterator

technicallyty:

updates the authorizepersonaaddresssystem to use the tx iterator

## Brief Changelog

_(for example:)_

- _The metadata is stored in the blob store on job creation time as a persistent artifact_
- _Deployments RPC transmits only the blob storage reference_
- _Daemons retrieve the RPC data from the blob cache_

## Testing and Verifying

_(Please pick one of the following options)_

_(or)_

This change is already covered by existing tests, such as _(please describe tests)_.

_(or)_

This change added tests and can be verified as follows:

_(example:)_

- _Added unit test that validates ..._
- _Added integration tests for end-to-end deployment with ..._
- _Extended integration test for ..._
- _Manually verified the change by ..._


pyrofolium (approved):
lgtm

PR #309, opened 2023-10-11
refactor(cardinal): Rewrite Query to use Filters involving IAbstractComponentType. 

pyrofolium:
Closes: #399

took out Each, First and MustFirst, replaced all unit tests that used these methods.

Made a new high level filter object.

Made a new method on world that can produce queries based on this high level filter object.

took out Get and Set which are now unused. These methods were replaced in a previous PR but the actual deletion of the methods did not occur. 

## Testing

Replaced all testing with new style queries. Internal queries on `IComponentType`... those tests remained untouched.


technicallyty:
are we eventually getting rid of `IComponentType` and the generic wrapper `ComponentType`? i'm a bit confused on what we're doing there as it seems to relate to the need to do conversions in this PR

pyrofolium:
> are we eventually getting rid of `IComponentType` and the generic wrapper `ComponentType`? i'm a bit confused on what we're doing there as it seems to relate to the need to do conversions in this PR

What we talked about in the refactor meeting was "Yes" that the objective will be to get rid of it. 

However after looking at the code, I realized that this is not a wise decision. Much of our logic depends on IComponentType and ComponentType. If we get rid of it we have to rewrite much of our logic, and likely that rewrite will end up with data structures and types that are virtually identical to IComponentType and ComponentType but probably with a different name.

What ended up happening is that Jeremy's proposal at it's core was just an aesthetic change. The structure of the code remains virtually identical but we create an API layer where the user only interacts with `IAbstractComponent.` Then this layer can be tidied up however we see fit.

The API will meet the spirit of the initial objective. IComponent and ComponentType will not be gotten rid of but they will be made Private and forever attached to ECS. Like I stated before there are intrinsic properties in Golang that prevents us from fully organizing our code exactly the way we want to.

I believe Namra has a better handle on the final organization. It's still fuzzy but I believe there will be a Cardinal package which is the API Layer that is public. Namra and Scott have a better outlook on this.

cardinal/ecs/ecs_test.go
```diff
@@ -74,16 +79,18 @@ func TestECS(t *testing.T) {
 	assert.NilError(t, world.LoadGameState())
 
 	assert.NilError(t, world.Tick(context.Background()))
-
-	Energy.Each(world, func(id entity.ID) bool {
+	query, err := world.NewQuery(ecs.And(ecs.Contains(EnergyComponent{})))
```
jerargus, line 82: This ecs.And is redundant.

cardinal/ecs/ecs_test.go
```diff
@@ -418,26 +430,29 @@ func TestEntriesCanChangeTheirArchetype(t *testing.T) {
...
 	// 3 entities have alpha
-	alpha.Each(world, countAgain())
+	alphaQuery, err := world.NewQuery(ecs.Contains(Alpha{}))
+	assert.NilError(t, err)
+	alphaQuery.Each(world, countAgain())
 	assert.Equal(t, 3, count)
 
 	// 0 entities have gamma
-	gamma.Each(world, countAgain())
+	gammaQuery, err := world.NewQuery(ecs.Contains(Gamma{}))
```
jerargus, line 439: err is not checked.

jerargus (approved):
These comments are optional.

cardinal/ecs/ecs_test.go
pyrofolium: fixed

cardinal/ecs/ecs_test.go
pyrofolium: fixed

PR #310, opened 2023-10-11
refactor(cardinal): get rid of Create and CreateMany world methods, Replace with functions that use IAbstractComponentType

pyrofolium:
Replace with functions that take in IAbstractComponentType.

Closes: #400

To make entity creation rely on IAbstractComponentType

Introduces two functions: `Create(world, IAbstractComponentType)` and `CreateMany(world, int, IAbstractComponentType...)`

Removes two methods: `world.Create(IComponentType)` and world.CreateMany(int, IComponentType)

## Testing

All associated tests have been changed. To work.


cardinal/ecs/component.go
```diff
@@ -115,6 +115,26 @@ func (c *ComponentType[T]) MustFirst(w *World) entity.ID {
 	return id
 }
 
+func CreateMany(world *World, num int, components ...component.IAbstractComponent) ([]entity.ID, error) {
+	acc := make([]IComponentType, 0, len(components))
+	for _, comp := range components {
+		c, ok := world.GetComponentByName(comp.Name())
```
jerargus, line 121: This pattern of calling GetComponentByName, and returning an error when it's absent seems to be repeated. I suggest making GetComponentByName return an error, and assigning the "Must register component before creating an entity" to an actual error variable.

cardinal/ecs/ecs_test.go
```diff
@@ -469,25 +473,35 @@ func TestCannotSetComponentThatDoesNotBelongToEntity(t *testing.T) {
...
+
+func (A) Name() string { return "a" }
+func (B) Name() string { return "b" }
+func (C) Name() string { return "c" }
+func (D) Name() string { return "d" }
+
 func TestQueriesAndFiltersWorks(t *testing.T) {
 	world := ecs.NewTestWorld(t)
-	a, b, c, d := ecs.NewComponentType[int]("a"), ecs.NewComponentType[int]("b"), ecs.NewComponentType[int]("c"), ecs.NewComponentType[int]("d")
+	a, b, c, d := ecs.NewComponentType[A]("a"), ecs.NewComponentType[B]("b"), ecs.NewComponentType[C]("c"), ecs.NewComponentType[D]("d")
```
jerargus, line 496: This is out of scope for this PR, but now that Name() is defined on these structs, is that first parameter to NewComponentType still required?

jerargus (approved):
These comments are optional.

cardinal/ecs/ecs_test.go
pyrofolium: No. This will be removed in a subsequent PR.

PR #313, opened 2023-10-11
refactor(chain): queue based router

technicallyty:
Closes: WORLD-393, WORLD-11

refactors the router to used a queue based mechanism to allow for atomicity within EVM smart contract router calls. also introduces a new `rift` package, where our cross-shard code will live. this will also remove the dependency on hosted code generation on buf. 

lots of TODOs left in this PR. the next steps to address here:

1. change how txs work in Cardinal so that it can understand and operate on transactions that originate from the EVM. send results back through Cardinal <> Base Shard channel. (maybe streaming grpc?)
2. once above is implemented, then we can implement callback object storage. this will open the path for a precompile method to read from this, allowing EVM users to read and operate on callback data.


chain/app/app.go
```diff
@@ -147,7 +147,7 @@ func NewApp(
 	baseAppOptions ...func(*baseapp.BaseApp),
 ) *App {
 	app := &App{}
-	app.setPlugins()
+	app.setPlugins(logger)
```
jerargus, line 150: Did you consider adding a logger field to the App struct?

technicallyty: yeah - it gets set in the embedded `runtime.App`. i'm opting to just pass it directly than move things around. router needs to be setup first thing

jerargus (approved):

PR #318, opened 2023-10-13
DO NOT SUBMIT - sample code for combining system parameters into a si…

jerargus:
Here is some sample code that show how a "Context" or "Climate" object will be implemented across ecs.

"Climate" was chosen here to ensure the name doesn't collide with the commonly-used [contex.Context](https://pkg.go.dev/context#Context). 

See the [Sys/Read Refactor](https://www.notion.so/arguslabs/World-Engine-V1-2fe15adfaa1a40e9ad0936558afda6f5?p=2586dc3392984100ac08557ad4d3e9d8&pm=s) doc in notions for more details.

### DIFF
--- a/cardinal/ecs/climate/climate.go
+++ b/cardinal/ecs/climate/climate.go
@@ -1,0 +1,24 @@
+package climate
+
+type Climate interface {
+	Set(key string, value any)
+	Get(key string) any
+}
+
+func NewClimate() Climate {
+	return &climateImpl{
+		m: map[string]any{},
+	}
+}
+
+type climateImpl struct {
+	m map[string]any
+}
+
+func (c *climateImpl) Set(key string, value any) {
+	c.m[key] = value
+}
+
+func (c *climateImpl) Get(key string) any {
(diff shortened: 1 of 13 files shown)


pyrofolium:
I think whatever we submit into the system it has to be called `World` because this is the `world-engine`.  The current `ecs.World` can be called `State` or some other thing. But naming aside for now I refer to your new thing as `Context` and our current `ecs.World` as `world`

I think this pattern can work but I would make the `Context` just a struct with specific fields for each thing. 

Instead of a struct holding `map[string]Any` make everything a field. That way it's type checked. Unless you anticipate us dynamically adding untyped information to this `Context` I think fields are better.

pyrofolium:
Also whatever this `Context` thing ends up being I think it should in principle be as immutable as possible. We should construct it dynamically and add all the relevant transient state to it before we call a system and pass this in. This helps reduce the complexity of dealing with mutating state. 

`map[string]Any` sort of implies mutating state like Context is a highly dynamic object being passed around.

PR #319, opened 2023-10-13
refactor(cardinal): remove the deprecated GetEntity and return an error…

jerargus:
… for GetEntitiesForArchID

Closes: #WORLD-394, WORLD-395

Remove GetEntity method - GetEntity returns an "entity" which is a combination of an ID and a Location. Location contains information about a "Component Index" which was a concept used in the old snapshotting version of the storage layer, but isn't used with the entity command buffer. 

Update GetEntitiesForArchID to return an error. Previously, the mapping of ArchetypeID to entities was handles completely in memory so this method didn't need to return an error. With ECB, this information is sometimes fetched from redis, so it's possible for this request to fail. This change ends up changing our query/filter code to also allow for the returning of errors.

### DIFF
--- a/cardinal/ecs/ecb/ecb.go
+++ b/cardinal/ecs/ecb/ecb.go
@@ -123,13 +123,6 @@ func (m *Manager) DiscardPending() {
 	m.pendingArchIDs = m.pendingArchIDs[:0]
 }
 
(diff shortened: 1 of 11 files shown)


pyrofolium (approved):

pyrofolium:
lgtm

PR #320, opened 2023-10-13
refactor(cardinal): rename read to query

jerargus:
Closes: #WOLD-111

Rename most instances of "read" to "query" in our code base. We've decided our public endpoints that attempt to fetch data without changing anything should be called "queries", so this PR updates our internal code base to match that convention.

While most of this was just simple renaming, there are two extra changes that deserve attention:

1) Strings in the swagger spec were changes. I believe there are existing tests that make sure the strings in the swagger spec match the strings in our code.
2) previously the files "read.go" and "query.go" existed in the ecs directory. read.go has been renamed to "query.go" and "query.go" has been renamed to "search.go". The entity Query struct has been renamed to Search

## Testing and Verifying

Existing tests verify there are no functional changes.

Also, from the cardinal directory:

```
grep -ri "[^l]read"  * | grep -v bytes.NewReader | grep -v io.ReadAll | grep -v os.ReadFile  | grep -v "/ecb/" | grep -i read
```

Does a case insensitive search for "read", but omits the word "already", instances of bytes.NewReader, io.ReadAll, os.ReadFile, and everything in the ecb package (there's lots of "read-only" stuff that has nothing to do with queries. I see these results:

```
ecs/transaction/transaction_test.go:func TestReadTypeNotStructs(t *testing.T) {
ecs/world.go:           // if a page response was in the reply, that means there is more data to read.
ecs/storage/redis_storage.go:           // Unlink is like Del, but the actual reclamation of memory happens in an asynchronous thread
ecs/receipt/receipt.go:// from ticks in the past are read only.
ecs/abi_test.go:        type FooReadBroken struct {
ecs/abi_test.go:        _, err := GenerateABIType(FooReadBroken{})
ecs/abi_test.go:        type FooReadBrokenSlice struct {
ecs/abi_test.go:        _, err = GenerateABIType(FooReadBrokenSlice{})
ecs/store/iface.go:type Reader interface {
ecs/store/iface.go:     Reader
go.sum:github.com/chzyer/readline v0.0.0-20180603132655-2972be24d48e/go.mod h1:nSuG5e5PlCu98SY8svDHJxuZscDgtXS6KTTbou5AhLI=
server/server.go:// Serve serves the application, blocking the calling thread.
server/events.go:               mt, message, err := c.ReadMessage()
server/events.go:                       log.Println("read:", err)
server/tx.go:           return nil, errors.New("params not readable")
server/tx.go:           return nil, nil, errors.New("params not readable")
server/server_test.go:func mustReadBody(t *testing.T, resp *http.Response) string {
server/server_test.go:  assert.Equal(t, 200, resp.StatusCode, "request failed with body: %v", mustReadBody(t, resp))
server/server_test.go:  //create readers
server/server_test.go:  body := mustReadBody(t, resp)
server/server_test.go:  // setup this read business stuff
server/server_test.go:  messageType, message, err := dial.ReadMessage()
```
Let me know if anything else from there looks like a candidate for `s/read/query`


namrapatel (approved):

pyrofolium:
Let's find a more intuitive name. We can rename both read and query or one of each. 

query_entity vs. query_world is kind of confusing. 

If I was a user approaching this fresh it wouldn't be immediately clear to me the difference.

technicallyty (approved):
we should def consider renaming the internal ecs Query stuff, i can see that being confusing. maybe that can be read instead?

jerargus:
I updated the internal/entity Query names to "Search".

technicallyty (approved):

pyrofolium (approved):
lgtm

PR #321, opened 2023-10-14
refactor(cardinal): refactor/organize component names

pyrofolium:
Closes: #397

When you look at the diff it can look complicated because most of a big renaming of components and moving alot of ece/component.go primitives into this new package. I hi-light the critical changes below: 

renames `IAbstractComponent` to `Component`

renames `IComponentType` and `ComponentType` into `IComponentMetadata` and `ComponentMetadata` respectively.

Puts both of these primitives in a new package called `component_metadata` This is a private package used behind the scenes. Users do not interact with component_metadata.  

The reasoning behind this is because ComponentType never really held information about the component. It's all just metadata information related to the component.

All tests should still work.


cardinal/component.go
```diff
@@ -1,24 +1,23 @@
...
 import (
-	"pkg.world.dev/world-engine/cardinal/ecs"
-	"pkg.world.dev/world-engine/cardinal/ecs/component"
+	"pkg.world.dev/world-engine/cardinal/ecs/component_metadata"
 )
 
-// AnyComponent is implemented by the return value of NewComponentType and is used in RegisterComponents; any
-// component created by NewComponentType can be registered with a World object via RegisterComponents.
+// AnyComponent is implemented by the return value of NewComponentMetaData and is used in RegisterComponents; any
+// component created by NewComponentMetaData can be registered with a World object via RegisterComponents.
```
smsunarto, line 8: nit: i think we should `Metadata` instead of `MetaData`

cardinal/ecs/component_metadata/component_metadata.go
smsunarto, line 1: should this just go to package `component` instead? i would want to avoid being too granular to avoid circular dependencies

pyrofolium: There is no more component package. The component is simply the struct the user creates.

What's going on here is I sort of made IComponent and ComponentType private and I gave these two primitives a private package. I also changed the name to Metadata too break the confusion we're having with The actual component data and type data related to where components are stored on the back end.  I think the old names also caused a lot of taxonomy and nomenclature issues. We called it "Components" but these things were heavily related to storage but the names caused us to categorize it as little components to pass around.

You still see the Component interface (previously IAbstractComponent)  in the metadata package because there's no real place to put it right now. In another PR I will be moving it to Cardinal or Cardinal.component or something like that.

pyrofolium: The end result will be two packages (or whatever combination of files/packages):

`component_metadata`. Which the user should never touch. 

and `components`, which is the only thing related to components the user touches. This package/file contains only one interface: `Component`

smsunarto: 👍

cardinal/ecs/world.go
```diff
@@ -592,7 +596,7 @@ func (w *World) getITx(id transaction.TypeID) transaction.ITransaction {
 }
 
 // Namespace returns the world's namespace.
-func (w *World) Namespace() string {
+func (w *World) NamespaceAsString() string {
```
jerargus, line 599: You should add a String() method to the Namespace instead of having a special accessor. Calls would then look like this:

w.Namespace().String()

jerargus:
I don't want to hold this PR up, so I'm going to approve it, but I don't agree with the `component` -> `component_metadata` change. If we eventually want a public facing component package we can either 1) put it in `world-engine/cardinal/component` or 2) move this internal component package to `world-engine/ecs/internal/component`. 

I think going from ComponentType to Metadata is good, but I'd rather see `Metadata` instead of `ComponentMetadata`. So usage in other files would be: `component.Metadata`.

jerargus (approved):

pyrofolium:
> I don't want to hold this PR up, so I'm going to approve it, but I don't agree with the `component` -> `component_metadata` change. If we eventually want a public facing component package we can either 1) put it in `world-engine/cardinal/component` or 2) move this internal component package to `world-engine/ecs/internal/component`.
> 
> I think going from ComponentType to Metadata is good, but I'd rather see `Metadata` instead of `ComponentMetadata`. So usage in other files would be: `component.Metadata`.

Internal would be critical here. Two component packages even in different locations is confusing. Internal would help but I think metadata adds more clarity. The package is not for components it literally only holds component_metadata. We can decide on this later.

PR #325, opened 2023-10-16
Add ability to send notifications to connected clients

jerargus:
Closes: #WORLD-379

Add a feature that will send the results of transactions to users via the Nakama notifications system.

Notifications are not saved, so they will only be delivered if the relevant client is actively connected and waiting for notifications. See [this PR](https://github.com/Argus-Labs/starter-game-template/pull/14) for the starter-game-template for some sample code of how to connect and view the notifications.

Note: The old and busted mechanism for sending tx results via match data has been left in place. Both can work side by side. Eventually the match data path will be removed.

## Testing and Verifying

Manual testing can be done by syncing to the above mentioned starter-game-templte branch, and using the `mage nakamaAt` target to run this world-engine/relay version of nakama.

Once Cardinal and Nakama are running, run the jsclient in the starget-game-template code base with:

```
node main.mjs
```
Once running, visit the Nakama UI console at localhost:7351 and go to the API Explorer.

issue a `nakama/claim-persona` transaction using the session user_id that was printed to the console.

The payload should be something like:

````
{"persona_tag": "my-awesome-name"}
```

After hitting `Send Request` you should automatically see the transaction receipt printed to the screen via the js client. It should look something like:

```
Notification content {                                                                                         
  errors: [ [length]: 0 ],                                                                                     
  result: { Success: true },                                                                                                                                                                                                  
  tx_hash: '0x65adf6b10e91edcac391155b39505df02ede221ff07f0db9ec923067d25611fc'
}
```

\

### DIFF
--- a/relay/nakama/main.go
+++ b/relay/nakama/main.go
@@ -73,7 +73,12 @@ func InitModule(ctx context.Context, logger runtime.Logger, db *sql.DB, nk runti
 	}
 
 	if err := initReceiptMatch(ctx, logger, db, nk, initializer); err != nil {
-		return fmt.Errorf("unable to init matches for receipt streaming")
+		return fmt.Errorf("unable to init match for receipt streaming: %w", err)
+	}
+
+	notifier, err := newReceiptNotifier(logger, nk)
+	if err != nil {
+		return fmt.Errorf("unable to init receipt notifier: %w", err)
 	}
 
 	if err := initPrivateKey(ctx, logger, nk); err != nil {
@@ -89,11 +94,11 @@ func InitModule(ctx context.Context, logger runtime.Logger, db *sql.DB, nk runti
 		return fmt.Errorf("failed to init persona tag verifier: %w", err)
 	}
 
-	if err := initPersonaTagEndpoints(logger, initializer, ptv); err != nil {
+	if err := initPersonaTagEndpoints(logger, initializer, ptv, notifier); err != nil {
 		return fmt.Errorf("failed to init persona tag endpoints: %w", err)
 	}
 
-	if err := initCardinalEndpoints(logger, initializer); err != nil {
+	if err := initCardinalEndpoints(logger, initializer, notifier); err != nil {
 		return fmt.Errorf("failed to init cardinal endpoints: %w", err)
 	}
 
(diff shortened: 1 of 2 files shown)


relay/nakama/notifications.go
```diff
@@ -0,0 +1,119 @@
...
+	txHash string
+	userID string
+}
+
+// receiptNotifier is a struct that sends out notifications to users based on transaction receipts.
+type receiptNotifier struct {
+	// txHashToTargetInto maps a specific transaction hash to a user ID. A timestamp is also tracked so "stale" transaction
+	// can be cleaned up.
+	txHashToTargetInfo map[string]targetInfo
+	// newTxHash is a channel that takes in txHash/userID tuples. An item on this channel signalts to the receiptNotifier
```
namrapatel, line 29: nit: typo, `signalts` to `signals`

relay/nakama/main.go
```diff
@@ -332,6 +343,17 @@ func initCardinalEndpoints(logger runtime.Logger, initializer runtime.Initialize
 				if err != nil {
 					return logError(logger, "can't read body: %w", err)
 				}
+				var asTx txResponse
+
+				if err = json.Unmarshal(bodyStr, &asTx); err != nil {
+					return logError(logger, "can't decode body as tx response: %w", err)
+				}
+				userID, err := getUserID(ctx)
```
namrapatel, line 351: I suppose it probably doesn't make sense to extend the functionality of this PR to support this, but I can imagine a use cases where a client may want to specify several user ids that should be sent a "notification" to. Take for example almost any board game like Chess, if `Player1` makes a move you probably want to register `Player2` to also get a notification for the result of that move just because that would increase the liveness of the game at minimal cost.

That being said, letting a client specify which user id's to reply is a feature that seems kind of across both Cardinal and Nakama to me at this point, unclear where it should be. 

If we are going to ignore this for now, should we backlog a ticket to remember to come back to this?

namrapatel (approved):

relay/nakama/main.go
jerargus: I added a ticket: https://linear.app/arguslabs/issue/WORLD-424/decide-which-users-should-see-which-transaction-receipts

PR #326, opened 2023-10-17
feat(cardinal): get basic event broadcasting working. 

pyrofolium:
Closes: #419

Get event broadcasting working via channels. Events are still incomplete. 

## Brief Changelog

Created EventHub struct. 
EventHub has channels for websocket connection registration and unregistration
has broadcast channel for messages. 

see events.go

## Testing and Verifying

Created TestEvents function.


cardinal/server/server_test.go
```diff
@@ -1222,3 +1230,36 @@ func TestWebSocket(t *testing.T) {
...
+		dialers[i] = dial
+	}
+	for i := 0; i < numberToTest; i++ {
+		i := i
+		go func() {
+			txh.eventHub.Broadcast <- []byte(fmt.Sprintf("test%d", i))
+		}()
+	}
+
+	for i, dialer := range dialers {
```
jerargus, line 1254: nit: `for _, dialer := range dialers {...}`

cardinal/server/server.go
```diff
@@ -50,7 +50,7 @@ func NewHandler(w *ecs.World, opts ...Option) (*Handler, error) {
 //go:embed swagger.yml
 var swaggerData []byte
 
-func newSwaggerHandlerEmbed(w *ecs.World, opts ...Option) (*Handler, error) {
+func newSwaggerHandlerEmbed(w *ecs.World, builder *middleware.Builder, opts ...Option) (*Handler, error) {
```
jerargus, line 53: same as above, the "*" can be removed.

cardinal/server/server.go
```diff
@@ -78,13 +78,20 @@ func newSwaggerHandlerEmbed(w *ecs.World, opts ...Option) (*Handler, error) {
...
+	}))
+
 	if err := api.Validate(); err != nil {
 		return nil, err
 	}
 
 	app := middleware.NewContext(specDoc, api, nil)
-
-	th.mux.Handle("/", app.APIHandler(nil))
+	if builder != nil {
```
jerargus, line 90: Once you remove the pointer up on line 53, the if statement here can be removed:

cardinal/server/server.go
```diff
@@ -39,8 +39,8 @@ var (
 // NewHandler instantiates handler function for creating a swagger server that validates itself based on a swagger spec.
 // transactions and queries registered with the given world are automatically created. The server runs on a default port
 // of 4040, but can be changed via options or by setting an environment variable with key CARDINAL_PORT.
-func NewHandler(w *ecs.World, opts ...Option) (*Handler, error) {
-	h, err := newSwaggerHandlerEmbed(w, opts...)
+func NewHandler(w *ecs.World, builder *middleware.Builder, opts ...Option) (*Handler, error) {
```
jerargus, line 42: middleware.Builder is already a pointer (to a function), so the "*" here doesn't add anything.

jerargus, line 42: Since the builder parameter is optional, why not make it an Option?

cardinal/server/events.go
```diff
@@ -1,32 +1,159 @@
...
+				if err != nil {
+					log.Logger.Error().Err(err)
+				}
+			}
+		case message := <-h.Broadcast:
+			for conn := range h.websocket_connections {
+
+				err := conn.SetWriteDeadline(time.Now().Add(3 * time.Second))
+				if err != nil {
+					go func() {
```
jerargus, line 55: This for/select loop ensures that data is only modified in a single goroutine, so there's no reason to use `go` to later Unregister a connection. I'd suggesting making an "unregisterConnection" function (either in this run() function or as a method on EventHub) and call it directly, instead of spinning up a goroutine.

cardinal/server/events.go
```diff
@@ -1,32 +1,159 @@
...
+						h.Unregister <- conn
+					}()
+					log.Logger.Error().Err(err)
+					break Loop
```
jerargus, line 59: This doesn't seem right. If ANY connection fails their SetWriteDeadline call we bail out of run entirely?

cardinal/server/events.go
```diff
@@ -1,32 +1,159 @@
...
+						h.Unregister <- conn
+					}()
+					log.Logger.Error().Err(err)
+				}
+
+			}
+		case <-h.Shutdown:
+			go func() {
+				for conn, _ := range h.websocket_connections {
+					h.Unregister <- conn
```
jerargus, line 73: This connection will never be unregistered. break Loop is called directly after spinning up this goroutine, so this select statement will never be used again.

cardinal/server/events.go
```diff
@@ -1,32 +1,159 @@
...
+
+	"github.com/go-openapi/runtime/middleware"
 
 	"github.com/gorilla/websocket"
+	"github.com/rs/zerolog/log"
 )
 
+func CreateEventHub() *EventHub {
+	res := EventHub{
+		websocket_connections: map[*websocket.Conn]bool{},
```
jerargus, line 15: this should be websocketConnections

jerargus (changes requested):

cardinal/server/events.go
pyrofolium: You have to use the go routine or it triggers a deadlock. h.Unregister will block until something grabs the message. 

But it's blocking in the Same loop that tries to grab the message.

cardinal/server/events.go
pyrofolium: yeah you're right, should not bail for a deadline failure. Will fix.

cardinal/server/events.go
pyrofolium: hmm yeah good point. Will fix.

cardinal/server/events.go
pyrofolium: as discussed, Will packager unregister logic into a closure then call directly.

cardinal/server/server.go
pyrofolium: because of the way it's passed in. Options in server operate on the handler overall, but the builder parameter needs to be passsed into another method that's activated on a route handler.

cardinal/server/events.go
pyrofolium: Oh wait. This should be true. Yes this means something is wrong. Getting a logic error on SerWriteDeadline is akin to a 500 error. It should never error unless the underlying connection has an issue. 

I think what should be fixed is if the WriteMessage call errors out BECAUSE of a deadline. That one has to be checked. Right now it's bailing.

pyrofolium: Documentation says that websocket connections completely close on deadline. So that means it's dead forever. I'll set deadline to 30s.

jerargus: 👍

cardinal/server/server.go
jerargus: 👍

PR #327, opened 2023-10-17
refactor(cardinal): create wrapper functions in cardinal for component related functions.

pyrofolium:
Closes: #421

Make public component package. Note there are temporary coding changes to get rid of circular dependencies. 

## Brief Changelog

Moved components out of ecs. 

Took GetComponent, SetComponent, Create, CreateMany and UpdateComponent and made private copies of these functions within persona. 

The objective here is to temporarily use these private functions to break the circular dependency. When we reimplement persona we can get rid of these private functions. The current version of persona is going to be deprecated anyway, so I think this pattern works. I can use it to decouple other things away from persona. 

## Testing and Verifying

All original tests still work.


cardinal/ecs/persona.go
```diff
@@ -193,3 +195,112 @@ func (w *World) GetSignerForPersonaTag(personaTag string, tick uint64) (addr str
 	}
 	return addr, nil
 }
+
+// TODO private component function used to temporarily remove circular dependency until we replace components.
```
jerargus, line 199: I created a ticket to capture the plugin work: https://linear.app/arguslabs/issue/WORLD-423/ecs-plugin-feature -- please add a link to the ticket in your comment here.

jerargus (approved):

cardinal/world.go
```diff
@@ -99,13 +99,13 @@ func NewMockWorld(opts ...WorldOption) (*World, error) {
 // CreateMany creates multiple entities in the world, and returns the slice of ids for the newly created
 // entities. At least 1 component must be provided.
 func (w *World) CreateMany(num int, components ...component_metadata.Component) ([]EntityID, error) {
```
namrapatel, line 101: Can/should we just get rid of these duplicates?

namrapatel (commented):
Are we pruning the `cardinal.World` function in a separate PR? If so, feel free to resolve, this looks good.

pyrofolium:
@namrapatel @jerargus @technicallyty 

Ok so there was a lot of discussion. Here are the overall changes. 

There are now two packages for components in `ecs`

`component_metadata`
`components`

The reason this exists is because of the dependency graph to prevent circular dependencies. 

`world` -> `component_metadata`
`components` -> `world`

`components` was moved back into `ecs` which is "private" with the possibility of making it internal later. The reason is if I move it into `cardinal` then that forces the user to import `ecs`.

jerargus:
Per our discussion, make sure to add the SetComponent/RemoveComponent/etc wrappers to world-engine/cardinal.

pyrofolium:
> Per our discussion, make sure to add the SetComponent/RemoveComponent/etc wrappers to world-engine/cardinal.

Done

jerargus (approved):

cardinal/world.go
pyrofolium: As discussed the other Create and CreateMany in components is now being moved to components in the ecs folder.

PR #333, opened 2023-10-19
feat(cardinal): integrate eventhub with cardinal. 

pyrofolium:
Closes: #426

Integrate eventHub into systems

## Brief Changelog
made two new methods on world:

EmitEvent and SetEventHub and FlushEvents

ecs.World now owns EventHub

## Testing and Verifying

Created new test that demonstrates eventHub working with systems and ticks.


cardinal/ecs/options.go
```diff
@@ -42,3 +43,9 @@ func WithStoreManager(s store.IManager) Option {
 		w.storeManager = s
 	}
 }
+
+func WithEventHub(eventHub *events.EventHub) Option {
+	return func(w *World) {
+		w.eventHub = eventHub
+	}
+}
```
technicallyty, line 51: i think we'd want this to be in world by default, not optional

cardinal/events/events_test.go
```diff
@@ -52,10 +56,92 @@ func TestEvents(t *testing.T) {
...
+			counter1.Add(1)
+			return nil
+		})
+	}
+	assert.NilError(t, ecs.RegisterComponent[garbageStructAlpha](w))
+	assert.NilError(t, ecs.RegisterComponent[garbageStructBeta](w))
+	assert.NilError(t, w.LoadGameState())
+	txh := test_utils.MakeTestTransactionHandler(t, w, server.DisableSignatureVerification())
+	url := txh.MakeWebSocketURL("events")
+	dialers := make([]*websocket.Conn, numberToTest, numberToTest)
```
technicallyty, line 106:
```suggestion
	dialers := make([]*websocket.Conn, numberToTest)
```
you can drop the second arg here, first arg to make is length, second is capacity

technicallyty (commented):
looks good! just a couple smol things

cardinal/ecs/options.go
pyrofolium: It has to be passed into ecs world externally because this eventHub is something that has to be used by the Server package as well. 

Thus the cardinal layer is where instantiation of `eventHub` occurs. It is "default" in from the cardinal perspective, but from the ecs perspective it is not. We could make it so that ecs.World won't even run without an eventHub, but that complicates every test as now you have to Spawn an unused eventHub

technicallyty: ahh gotcha, this is fine then - thanks! 🙏

pyrofolium: I can make a mock in some later PR. some eventHub that emits things to stdout that doesn't need to be owned by server so we can test stuff.

technicallyty: ahh good idea, if you can make a ticket for that that would be sweet

technicallyty (approved):
LGTM

PR #334, opened 2023-10-19
chore(cardinal): expose AddToQueue in Cardinal 

namrapatel:
Closes: WORLD-431

## Brief Changelog

- [x] Expose `ecs/transaction.go`'s `AddToQueue` in `cardinal.World`

## Testing and Verifying

We are ignoring the fact that we don't have great tests for `cardinal.World` for the time being.

### DIFF
--- a/cardinal/transaction.go
+++ b/cardinal/transaction.go
@@ -44,8 +44,9 @@ func NewTransactionTypeWithEVMSupport[Msg, Result any](name string) *Transaction
 	}
 }
 
+// AddToQueue is not meant to be used in production whatsoever, it is exposed here for usage in tests.
 func (t *TransactionType[Msg, Result]) AddToQueue(world *World, data Msg, sigs ...*sign.SignedPayload) {
-	ecs.AddToQueue[Msg, Result](world.implWorld, t.impl, data, sigs...)
+	t.impl.AddToQueue(world.implWorld, data, sigs...)
 }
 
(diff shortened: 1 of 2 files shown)


pyrofolium (approved):
lgtm

pyrofolium:
I think people will be opinionated on this one might stir up opinions. So wait on it before merge.

cardinal/transaction.go
```diff
@@ -44,6 +44,10 @@ func NewTransactionTypeWithEVMSupport[Msg, Result any](name string) *Transaction
 	}
 }
 
+func (t *TransactionType[Msg, Result]) AddToQueue(world *World, data Msg, sigs ...*sign.SignedPayload) {
+	ecs.AddToQueue[Msg, Result](world.implWorld, t.impl, data, sigs...)
```
jerargus, line 48: Can't you just make this t.impl.AddToQueue and leave out the change to cardinal/ecs/world.go?

jerargus (approved):
I don't think should be added to the top level cardinal package as a user shouldn't be using this method directly. Also, generating signatures is probably going to be confusing in normal usage. But since we don't have an alternate plan to allow for testing transactions, I think it's fine.

Please add a comment that this method is only for testing.

cardinal/transaction.go
namrapatel: Yeah, great idea, done!

jerargus (approved):

PR #337, opened 2023-10-23
chore(cardinal): expose ForEach on cardinal.TransactionType

namrapatel:
Closes: WORLD-441

- We need to expose ForEach on cardinal.TransactionType so users can use the Tx Iterator functionality via Cardinal and without ecs.

## Testing and Verifying

We are ignoring tests related to `cardinal.World` for the time being. [redacted] is the test.

### DIFF
--- a/cardinal/transaction.go
+++ b/cardinal/transaction.go
@@ -63,7 +63,11 @@ func (t *TransactionType[Msg, Result]) GetReceipt(wCtx WorldContext, hash TxHash
 }
 
 func (t *TransactionType[Msg, Result]) ForEach(wCtx WorldContext, fn func(TxData[Msg]) (Result, error)) {
-	t.impl.ForEach(wCtx.getECSWorldContext(), fn)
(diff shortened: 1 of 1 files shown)


jerargus (approved):

pyrofolium (changes requested):
```go
func (t *TransactionType[Msg, Result]) ForEach(wCtx WorldContext, fn func(TxData[Msg]) (Result, error)) {
	adapterFn := func(ecsTxData ecs.TxData[Msg]) (Result, error) {
		adaptedTx := TxData[Msg]{impl: ecsTxData}
		return fn(adaptedTx)
	}
	t.impl.ForEach(wCtx.getECSWorldContext(), adapterFn)
}
```

When the user defines `fn func(t TxData[Msg])` he can't do anything with t which accesses impl. Impl is private. So methods in this wrapper need to be moved over from ecs.

pyrofolium (approved):

PR #338, opened 2023-10-24
chore(cardinal): make these methods use wCtx.

pyrofolium:
Doesn't close any tickets. We missed a couple of methods for worldContext.


namrapatel (approved):
I was going to do this next because I needed it, thanks for taking care of this

PR #339, opened 2023-10-24
test(cardinal): Add test samples to cardinal package and extend test_utils package

jerargus:

Add test files to the cardinal package to demonstrate how to test various features of cardinal (and to test the cardinal package itself).

There are two aspects of this PR that I'd like some specific feedback on:
1) The new private method in test_utils called "uniqueSignature" is lifted directly from `world-engine/cardinal/ecs/internal/testutil`. Should this internal testutil package get lifted out into a common area?

2) In test_utils.AddTransactionToWorldByAnyTransaction I have 2 methods for converting an externally facing *cardinal.World into an internal *ecs.World. Do you all have a preference? I'd prefer to keep it so the cardinal godoc package does not list any TestOnly methods, but the resulting pattern for accessing the data is a little weird.

## Brief Changelog

- Updated ecs code to allow for the shutting down of an in-memory redis server (otherwise, multiple tests will end up trying to use the same port for redis and fail).
- Updated evm code so grpc server can be shut down (again, so tests don't fight over the same port).
- Added some code to the test_utils package so a user can get a cardinal.WorldContext object outside of systems and queries (this allows for setting up and asserting the game state is correct).

## Testing and Verifying

Unit tests were added.

### DIFF
--- a/cardinal/cardinal_test.go
+++ b/cardinal/cardinal_test.go
@@ -5,10 +5,25 @@ import (
 	"time"
 
 	"gotest.tools/v3/assert"
-
 	"pkg.world.dev/world-engine/cardinal"
 	"pkg.world.dev/world-engine/cardinal/test_utils"
 )
+
+func makeWorldAndTicker(t *testing.T) (world *cardinal.World, doTick func()) {
+	startTickCh, doneTickCh := make(chan time.Time), make(chan uint64)
+	world, err := cardinal.NewMockWorld(
+		cardinal.WithTickChannel(startTickCh),
+		cardinal.WithTickDoneChannel(doneTickCh))
+	t.Cleanup(func() {
+		world.ShutDown()
+	})
+	assert.NilError(t, err)
+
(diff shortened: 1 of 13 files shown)


namrapatel (approved):
(1) I'm personally pretty liberal about the things that I think are fine to expose to the user because I think pro-users should be allowed to look into things like test_utils and some internal functions and abuse them if it serves a purpose for them that we have not considered. So I'd be in favor of lifting any and all functions that you think fit from the `internal/test_utils` folder out to the new top-level `test_utils`

(2) I prefer the "TestOnly" way of getting the ECS world, doing the interface casting thing and other stuff like that feels somewhat unnecessary to me and I feel like I have used npm/pip packages before that have functions exposed that are only for tests and I think this does little harm if any at all

jerargus:
@namrapatel Ok, I added some "Testing" methods in a testing.go file.

I'll add a ticket to unify the two uniqueSigner methods.

PR #343, opened 2023-10-24
refactor(cardinal): expose ability to add a create persona tx to queue

namrapatel:
Closes: WORLD-449

Dark Forest tests are currently calling:

`ecs.CreatePersonaTx.AddToQueue(world, transaction)`

There is no way to do this because we do not have access to ecs.World which is the world expected in that argument. Thus, we'll wrap this functionality and expose it in Cardinal.

## Testing and Verifying


pyrofolium (approved):
lgtm

PR #355, opened 2023-10-25
test(cardinal): Re-add cardinal package tests and add helper function to test_utils t…

jerargus:
…hat will start/stop a cardinal.World object correctly between tests

- Re-add the test files to the cardinal package.
- Add a helper in test_utils that (should) start up a cardinal.World within a test context and properly shut it down between tests. This requires a change to the world.Shutdown method that makes shutting down a non-started world NOT an error.

## Testing and Verifying

Please checkout this branch and run the `cardinal` tests locally to verify they don't hang. You can also use `--count` to make extra sure resources are cleaned up between similar tests.

`go test --count=10`

### DIFF
--- a/cardinal/evm/server.go
+++ b/cardinal/evm/server.go
@@ -39,6 +39,7 @@ type Server interface {
 	routerv1.MsgServer
 	// Serve serves the application in a new go routine.
 	Serve() error
+	Shutdown()
 }
 
 // txByID maps transaction type ID's to transaction types.
@@ -58,6 +59,8 @@ type msgServerImpl struct {
 	// opts
 	creds credentials.TransportCredentials
 	port  string
+
+	shutdown func()
 }
 
 // NewServer returns a new EVM connection server. This server is responsible for handling requests originating from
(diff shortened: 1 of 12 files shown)


namrapatel (approved):

cardinal/evm/server.go
```diff
@@ -170,9 +173,16 @@ func (s *msgServerImpl) Serve() error {
...
+	s.shutdown = server.GracefulStop
 	return nil
 }
 
+func (s *msgServerImpl) Shutdown() {
+	if s.shutdown != nil {
+		s.shutdown()
+	}
+}
+
```
technicallyty, line 185: perhaps it would be useful to do an `else { panic("evm.Server was shutdown without calling Serve() }` or something of that nature? this feels a tad leaky

cardinal/system_test.go
technicallyty, line 1: can we also add a test with multiple systems? want to avoid that go for/range iterator bug again if it happens to pop up.

cardinal/evm/server.go
jerargus: Per our conversation, I'm treating this "Shutdown" as "make sure the server is not running", so it shouldn't produce an error. Similarly, calling Shutdown multiple times shouldn't produce an error. Go's http.Server.Shutdown follows a similarly pattern.

cardinal/system_test.go
jerargus: Good idea. I'll add it.

technicallyty (approved):

PR #365, opened 2023-10-27
chore(chain): token denom from eth -> ether, update polaris versions, remove replace directive for precompiles/contracts

technicallyty:

removes the needless fork of precompile/contracts directories. these can be imported and we can simply write our own and pass to the precompiles in the evm keeper.

renamed the denom from eth -> ether. 

updates polaris version to latest from the celestia team.


namrapatel (approved):
LGTM

PR #369, opened 2023-10-30
feat(relay): finish nakama migration from starter to world engine

pyrofolium:
Does not close #462

requires additional PR to starter-game-template to complete

Upgrade world-engine nakama so it has all the features from the nakama in starter and all the tests as well. 

## Testing and Verifying

There's a new test to test out the event system.


internal/e2e/tester/cardinal/sys/join.go
```diff
@@ -26,6 +28,7 @@ func Join(ctx cardinal.WorldContext) error {
 		}
 		PlayerEntityID[jtx.Sig().PersonaTag] = entityID
 		logger.Info().Msgf("player %s successfully joined", jtx.Sig().PersonaTag)
+		ctx.EmitEvent(fmt.Sprintf("%d player: %d created, %d/%d", entityID, entityID, entityID, 3))
```
technicallyty, line 31: - not sure we need the "3" at the end
- do we need to list the entity id 3 times?

internal/test_nakama/util.go
```diff
@@ -30,6 +32,65 @@ func newClient(t *testing.T) *nakamaClient {
 	return h
 }
 
+type NotificationItem struct {
+	ID         string    `json:"id"`
+	Subject    string    `json:"subject"`
+	Content    string    `json:"content"`
+	Code       int       `json:"code"`
+	CreateTime time.Time `json:"create_time"`
```
technicallyty, line 40:
```suggestion
	CreateTime time.Time `json:"createTime"`
```

internal/test_nakama/util.go
```diff
@@ -30,6 +32,65 @@ func newClient(t *testing.T) *nakamaClient {
...
+	Persistent bool      `json:"persistent"`
+}
+
+type Content struct {
+	Message string `json:"message"`
+}
+
+type NotificationCollection struct {
+	Notifications   []NotificationItem `json:"notifications"`
+	CacheableCursor string             `json:"cacheable_cursor"`
```
technicallyty, line 50:
```suggestion
	CacheableCursor string             `json:"cacheableCursor"`
```

relay/nakama/main.go
```diff
@@ -122,6 +126,32 @@ func initReceiptDispatcher(log runtime.Logger) error {
 	return nil
 }
 
+func initEventHub(ctx context.Context, log runtime.Logger, nk runtime.NakamaModule) error {
+	eventHub, err := createEventHub(log)
+	if err != nil {
+		return err
+	}
+	go func() {
+		_ = eventHub.dispatch(log)
```
technicallyty, line 135:
```suggestion
		err := eventHub.dispatch(log)
```

internal/e2e/tester/cardinal/sys/join.go
pyrofolium: hmm this is for a test. Previously I could actually get the amount transactions. Now with this new method there's no access to the length. I was thinking that in the future I could create a method that could access the length and replace it here but I think I'm just going to modify the test.

technicallyty: saw the fix, but still seems unnecessary to send the entity ID 3 times here, we can just simplify to 1?

pyrofolium: Well, this test actually caught one of the huge bugs I was trying to work on. Basically events were lossy, if I sent events too fast I would actually be missing events. If I didn't send 3 this bug would not have been caught.

technicallyty: missing event data or missing entire events? my point is that you have 
```go
ctx.EmitEvent(fmt.Sprintf("%d player: %d created, %d/%d", entityID, entityID, entityID, 3))
```
when you could just do 
```go
ctx.EmitEvent(fmt.Sprintf("player entity id %d created", entityID))
```

pyrofolium: I meant Missing entire events. 

Ok I thought you meant emitting the id, three times not the string itself. 

Yeah, idc, The test, however requires the first value to be the entityid. So I'll just make it work with that.

technicallyty (approved):

PR #375, opened 2023-11-01
feat(cardinal): various fixes and logging updates. 

pyrofolium:
Closes Multiple tickets: #133 and 475. Partially complete 465, 465 will not be closed until a push to starter-game-template. 

- Shutdown function was made non-blocking by using a buffered channel. This can be achieved with a goroutine instead getting rid of the arbitrary buffer number. 
- logging was added to when cardinal serves 
- some logging calls were edited to be more compatible with pretty printing. 
-a test was changed to make sure that the test only finishes when startgame unblocks.


cardinal/cardinal_test.go
```diff
@@ -59,9 +59,11 @@ func TestShutdownViaSignal(t *testing.T) {
 		_, err := cardinal.CreateMany(worldCtx, wantNumOfEntities, Foo{})
 		assert.NilError(t, err)
 	})
+	otherFinish := make(chan bool)
```
jerargus, line 62: optional: You could use a [WaitGroup](https://pkg.go.dev/sync#WaitGroup) here.

jerargus (approved):

cardinal/server/server.go
```diff
@@ -197,8 +198,12 @@ func (handler *Handler) Initialize() {
 // Serve serves the application, blocking the calling thread.
 // Call this in a new go routine to prevent blocking.
 func (handler *Handler) Serve() error {
-	err := handler.server.ListenAndServe()
-	return err
+	hostname, err := os.Hostname()
+	if err != nil {
+		return err
+	}
+	log.Info().Msg(fmt.Sprintf("serving cardinal at %s:%s", hostname, handler.Port))
```
technicallyty, line 205: you can use `Msgf` here

cardinal/world.go
```diff
@@ -57,19 +56,20 @@ type (
...
-		log.Log().Msg("Redis password is not set, make sure to set up redis with password in prod")
+		log.Info().Msg("Redis password is not set, make sure to set up redis with password in prod")
 	}
 
 	redisStore := storage.NewRedisStorage(storage.Options{
 		Addr:     addr,
 		Password: password, // make sure to set this in prod
 		DB:       0,        // use default DB
 	}, "world")
+	log.Info().Msg(fmt.Sprintf("redis address: %s", addr))
```
technicallyty, line 72: and here

cardinal/cardinal_test.go
technicallyty: agree there - we should remove both channels and use waitgroup for their replacements

PR #378, opened 2023-11-01
feat(cardinal): allow user to generate and keep track of custom sigs for testing.

pyrofolium:
is part of : #434 Needed to test the tic tac toe game. 

Allow users to keep track of signatures for tests. 

added custom delay for doTick. The reason this is needed is because of debugging. When stepping through code related to a tick it can incur a delay of > 5 seconds. This leads to a subtle broken connection error with miniredis. Probably should have some major panic here to let the user know what actually happened. but that's for another issue. 

I'm open to other ways to mess with this. lmk.


cardinal/testutils/test_utils.go
```diff
@@ -125,7 +125,16 @@ var (
 	privateKey *ecdsa.PrivateKey
 )
 
-func uniqueSignature() *sign.SignedPayload {
+func UniqueSignature(optionalPersonaTagName ...string) *sign.SignedPayload {
```
jerargus, line 128: Using variadic parameters for optional parameters is not standard. I'd suggest making a second function:

```
UniqueSignatureWithPersonaTag(personaTag string) *sign.SignedPayload {...}
```

cardinal/testutils/test_utils.go
```diff
@@ -136,14 +145,18 @@ func uniqueSignature() *sign.SignedPayload {
...
 	}
 	return sig
 }
 
-func AddTransactionToWorldByAnyTransaction(world *cardinal.World, cardinalTx cardinal.AnyTransaction, value any) {
+func AddTransactionToWorldByAnyTransaction(
+	world *cardinal.World,
+	cardinalTx cardinal.AnyTransaction,
+	value any,
+	signedPayload ...*sign.SignedPayload) {
```
jerargus, line 159: Same as above. In this case, the single AddTransactionToWorldbyAnyTransaction should just have a required signature parameter; it's easy to make that signature with the above UniqueSignature method.

cardinal/testutils/test_utils.go
```diff
@@ -188,8 +212,13 @@ func MakeWorldAndTicker(t *testing.T, opts ...cardinal.WorldOption) (world *card
 
 	startGameOnce := sync.Once{}
 	// Create a function that will do a single game tick, making sure to start the game world the first time it is called.
-	doTick = func() {
+	doTick = func(timeoutTime ...int) {
```
jerargus, line 215: For this one, either make it a required time.Duration or a boolean that disables the timeout.

Alternatively, remove the timeout portion of this doTick method completely. Debugging tests without having to worry about this method forcing a test failure sounds useful, and if we find that tests in CI or locally are hanging instead of failing we can start using the `-timeout` flag for tests.

jerargus (approved):

PR #379, opened 2023-11-01
chore(cardinal): add log for iterated txs that encounter an error

technicallyty:

adds an error log for when systems using tx iterators return an error in the iterator.

## Brief Changelog

- add log

## Testing and Verifying

ran existing test:

```
{"level":"error","system":"ecs_test.TestForEachTransaction.func1","error":"some error","time":"2023-11-01T15:57:57-07:00","message":"tx 0x43fdd7497422eacb6f8af0cf70cabeeb1589190b777e4ca03b3a4db34dbde381 from some-persona-tag encountered an error with tx={GenerateError:true}"}
```


namrapatel (approved):
I like it, but do you think this would be annoying if someone wanted to add custom errors and just had this one getting in there as well?

technicallyty:
> I like it, but do you think this would be annoying if someone wanted to add custom errors and just had this one getting in there as well?

you can still do that, just write whatever error you want and return it

PR #381, opened 2023-11-02
test(cardinal): test to maintain interface component.

pyrofolium:
Closes: #481

see code change.


cardinal/ecs/component/metadata/component_metadata_test.go
```diff
@@ -27,6 +28,14 @@ type ComponentDataB struct {
 
 func (ComponentDataB) Name() string { return "b" }
 
+func getNameOfComponent(c metadata.Component) string {
+	return c.Name()
+}
+
+func TestComponentInterfaceSignature(t *testing.T) {
+	assert.Equal(t, getNameOfComponent(&ComponentDataA{}), "a")
+}
```
technicallyty, line 37: does this give us some sort of edge over just directly calling `componentdata.Name()`?

technicallyty (approved):

namrapatel (approved):

cardinal/ecs/component/metadata/component_metadata_test.go
pyrofolium: No the purpose of this test is to throw a compiler error IF the interface for `metadata.Component` changes. 

So like say one of us decides to change the method to be called String(). This test will throw a compiler error. Actually I should put a comment here to be clear.

technicallyty: ahh ok got it

pyrofolium: actually tbh, this test is a little redundant because a lot of integration tests cover it. But whatever doesn't hurt.

PR #396, opened 2023-11-06
refactor(cardinal)!: rename Transaction/TransactionType to Message/MessageType

technicallyty:
## NOTE

im going to leave the TransactionQueue as it is currently named. Would change it to MessageQueue if it only contained messages, but i believe Transaction takes precedence here

renames cardinal's Transaction types to Message.

## Testing and Verifying

run cardinal unit tests.


cardinal/ecs/world.go
```diff
@@ -647,16 +647,15 @@ func (w *World) decodeTransaction(bz []byte) (*shardv1.Transaction, error) {
...
-// getITx iterates over the registered transactions and returns the ITransaction associated with the TypeID.
-func (w *World) getITx(id transaction.TypeID) transaction.ITransaction {
-	var itx transaction.ITransaction
-	for _, tx := range w.registeredTransactions {
-		if id == tx.ID() {
-			itx = tx
-			break
+// getITx iterates over the registered all message.Message and returns the message.Message associated with the
+// message.TypeID.
+func (w *World) getITx(id message.TypeID) message.Message {
```
jerargus, line 652: Should this be getMsg?

jerargus (approved):

cardinal/ecs/world.go
technicallyty: yes! good catch [rename getITx to getMessage](https://github.com/Argus-Labs/world-engine/pull/396/commits/5839579c765310785034e507a2dbb0f3729121c3)

PR #397, opened 2023-11-06
feat(relay): add back allowlist module

technicallyty:

adds back the allowlist module that was previously only present in DF backend's fork of nakama.

## Testing and Verifying
manually


namrapatel (approved):

namrapatel:
gotta change the docker compose to use `build: ./nakama`

PR #402, opened 2023-11-07
chore: move makefiles to their own folder

technicallyty:
move makefiles to their own folder.

rename chain.mk to build.mk 

## Testing and Verifying

manually


namrapatel (approved):
just exceptional

PR #403, opened 2023-11-07
feat(cardinal): allow cors on dev mode

pyrofolium:
Closes: #500

Allow CORS as an option for servers. 

## Testing and Verifying

Added an additional check to an existing test


cardinal/option.go
```diff
@@ -82,3 +82,9 @@ func WithPrettyLog() WorldOption {
 		ecsOption: ecs.WithPrettyLog(),
 	}
 }
+
+func WithCORS() WorldOption {
+	return WorldOption{
+		serverOption: nil,
+	}
+}
```
technicallyty, line 90: something missing here 🕵🏻‍♂️

pyrofolium: eh good catch.

pyrofolium: fixed added coverage for this function too.

technicallyty (approved):
LGTM

PR #405, opened 2023-11-08
fix(sign): Sort JSON blobs before signing them

jerargus:
Closes: #world-522

Swagger endpoints process transaction bodies as a map[string]any{}. This object is JSON serialized which ends up sorting the JSON fields. If the original transaction signature was generated with a JSON body that did NOT have sorted fields, different hashes will be generated which will cause the signature verification to fail. 

This PR fixes signature generation so that unsorted JSON blobs are sorted before the hash is generated.

## Testing and Verifying

- _Added unit test that validates an unsorted JSON blob can be signed and verified.

### DIFF
--- a/sign/sign.go
+++ b/sign/sign.go
@@ -126,20 +126,22 @@ func normalizeJSON(data any) ([]byte, error) {
 		return json.Marshal(data)
 	}
 
-	if !json.Valid(asBuf) {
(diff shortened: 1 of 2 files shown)


namrapatel (approved):
amazing

technicallyty (approved):

PR #406, opened 2023-11-08
chore: update sign deps for nakama & cardinal

technicallyty:
updates cardinal & nakama with new sign version that includes bugfix


pyrofolium (approved):
lgtm

PR #407, opened 2023-11-08
feat(cardinal): create new init system.

pyrofolium:
Closes: #528

Allows the developer to register a system for initialization. Sort of like a constructor. 

This replaces world.Init which would just run a function in the context of worldContet. When you used the old world.Init before starting the game loop it could trigger an error if entities were created. 

By default registered init system will do nothing.

## Brief Changelog

Modified some tests to check for stuff. 
Renamed the old init to `RunInContextOfWorldContext`
Made new `Init` function that registers a "constructor" system only at tick 0.

### DIFF
--- a/cardinal/world.go
+++ b/cardinal/world.go
@@ -6,12 +6,13 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
-	"pkg.world.dev/world-engine/cardinal/ecs/message"
 	"reflect"
 	"runtime"
 	"sync/atomic"
 	"syscall"
 	"time"
+
+	"pkg.world.dev/world-engine/cardinal/ecs/message"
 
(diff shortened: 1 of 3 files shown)


cardinal/world.go
```diff
@@ -285,7 +286,14 @@ func (w *World) Tick(ctx context.Context) error {
 	return w.implWorld.Tick(ctx)
 }
 
-func (w *World) Init(fn func(WorldContext)) {
+func (w *World) RunInContextOfWorldContext(fn func(WorldContext)) {
```
jerargus, line 289: I'd say remove this entirely and use `cardinal.TestingWorldToWorldContext` in the tests that need it.

jerargus (approved):

PR #408, opened 2023-11-08
feat(cardinal): WaitForNextTick blocks until a tick has been completed

jerargus:
Closes: #WORLD-430

Update the WaitForNextTick method to use channels to block until a tick has completed. Remove the timeout path.
 
The TestServer_SendMessage test has also been refactors. The SendMessage method waits until a tick has completed before returning, which means it wouldn't be possible to call SendMessage 5 times to queue up 5 messages in the same tick. The test has been refactored so the SendMessage calls happen sequentially across multiple ticks.

This is OK in production, because a separate goroutine will be handling each SendMessage call. 

## Testing and Verifying

- Added unit test that validates the WaitForNextTick will block until a tick has completed, and will NOT block if the world has been shut down.


technicallyty (approved):
great!

PR #411, opened 2023-11-09
chore(cardinal): log a warning if RegisterSystems encounters a duplicate System name

technicallyty:

isnt 100% fool-proof as systems that cant get their names through reflection won't be checked.

also renames `ecs`'s AddSystem(s) to RegisterSystem to align naming.

## Testing and Verifying

unit test in system_test.go


cardinal/system_test.go
```diff
@@ -83,3 +87,9 @@ func TestCanRegisterMultipleSystem(t *testing.T) {
 	assert.Check(t, firstSystemCalled)
 	assert.Check(t, secondSystemCalled)
 }
+
+func TestCannotRegisterSameSystemTwice(t *testing.T) {
```
jerargus, line 91: Add a test that inserts 2 systems with distinct names and assert that no error is generated (I predict that new test will fail).

jerargus (changes requested):

cardinal/system_test.go
technicallyty: resolved ser

jerargus (approved):

namrapatel (approved):

PR #418, opened 2023-11-13
feat(cardinal): force entity creation to happen before LoadGame and StartGame.

pyrofolium:
Closes: WORLD-530

## Overview

Entity creation should not happen before LoadGameState or you could have an obscure low level error. 
The key word is "could" as many tests throughout cardinal actually create entities before LoadGameState ignoring this overall rule but still passing. 

This change essentially makes it so at the ECS level you cannot create an entity before LoadGameState and at the Cardinal level you cannot create an Entity before StartGame and the engine will throw always throw a loud and clear error message if you attempt to do this. 

## Brief Changelog
created state on `ecs.World`: `isEntityCreated`
created two methods to access that state `IsEntityCreated` and `SetEntityCreated`
`isEntityCreated` is set to false on initialization and set to true any time entity creation occurs.
LoadGameState will immediately throw an error if `IsEntityCreated` is true when you call it. 
created a low level Error in ECS called `EntitiesCreatedBeforeLoadingGameStateError`
created a high level Error in cardinal called `EntitiesCreatedBeforeLoadingGameStateError`
Low level error will convert to high level error depending on context.
Relevant files:
`cardinal/world.go`
`cardinal/ecs/world.go`

All other changes are just changing the tests to reflect this ordering requirement.


jerargus:
I think you can catch this error without the new `isEntitiesCreated` field. the `stateIsLoaded` fields exist in ecs.World, and that can be checked near the top of CreateMany. This means the user would see the entity creation failure right when they called Create or CreateMany.

Are there use cases where a user will call Create/CreateMany but never call LoadGameState?

pyrofolium:
> I think you can catch this error without the new `isEntitiesCreated` field. the `stateIsLoaded` fields exist in ecs.World, and that can be checked near the top of CreateMany. This means the user would see the entity creation failure right when they called Create or CreateMany.
> 
> Are there use cases where a user will call Create/CreateMany but never call LoadGameState?
@jerargus 
I noticed this. The reason this is done is because there are test cases where Entities are created but game state is never loaded. We do want to do unit testing without loading game state.

We want failure to happen on LoadGameState for this reason. 

In short: 

It is incorrect to create Entities Before `LoadGameState` is called. 
It is correct to create entities without ever calling `LoadGameState`

jerargus (approved):

PR #420, opened 2023-11-13
feat(cardinal): add eris (errors with stacktraces) to eventhub

pyrofolium:
Closes: WORLD-536

## Overview

This is a sub issue of https://linear.app/arguslabs/issue/WORLD-451/find-a-way-to-get-all-errors-to-be-paired-with-stack-trace

Relevant conversation here: https://argus-labs.slack.com/archives/C03G859K5TQ/p1699905811770509

This is a first PR of many that will eventually convert world-engine to use errors from this package:

https://github.com/rotisserie/eris

For the first part I just chose a really tiny use case to get a feel for the library. EventHub does not really create many custom errors. 

One drawback is that errors instantiated from other libraries or packages will not have stack traces. As soon as these errors touch our ecosystem we will have to wrap the errors with `erin.Wrap(err, "")` We can add additional error messages in place of the empty string. 

New errors can simply be instantiated with `erin.New` or `erin.Errorf`

## Tests

See the test for a demonstration about how this works. 

`eris.ToJSON` will create a nested map with the stack trace in map["root"]["stack"] as a list of strings where each string is one level on the stack. 
`eris.ToString` will create a string with readable stack trace.

### DIFF
--- a/cardinal/events/events.go
+++ b/cardinal/events/events.go
@@ -1,6 +1,7 @@
 package events
 
 import (
+	"encoding/json"
 	"net/http"
 	"sync"
 	"sync/atomic"
@@ -166,7 +167,7 @@ func (eh *webSocketEventHub) Run() {
 			delete(eh.websocketConnections, conn)
 			err := eris.Wrap(conn.Close(), "")
 			if err != nil {
-				log.Logger.Error().Err(err)
+				log.Logger.Error().Err(err).Msg(eris.ToString(err, true))
 			}
 		}
 	}
@@ -192,7 +193,7 @@ Loop:
 							go func() {
 								eh.UnregisterConnection(conn)
 							}()
-							log.Logger.Error().Err(err)
+							log.Logger.Error().Err(err).Msg(eris.ToString(err, true))
 							break
 						}
 						err = eris.Wrap(conn.WriteMessage(websocket.TextMessage, []byte(event.Message)), "")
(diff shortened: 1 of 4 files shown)


technicallyty (commented):
overall looks good, but unsure of the places with `eris.Wrap` and then we don't use the error again

cardinal/events/events.go
```diff
@@ -289,17 +297,18 @@ func WebSocketEchoHandler(ws *websocket.Conn) error {
...
 func Echo(w http.ResponseWriter, r *http.Request) {
 	c, err := upgrader.Upgrade(w, r, nil)
+	err = eris.Wrap(err, "")
 	if err != nil {
 		log.Print("upgrade:", err)
 		w.WriteHeader(http.StatusInternalServerError)
 		return
 	}
-	err = WebSocketEchoHandler(c)
+	err = eris.Wrap(WebSocketEchoHandler(c), "")
```
technicallyty, line 306: i.e. here

pyrofolium:
> overall looks good, but unsure of the places with `eris.Wrap` and then we don't use the error again

@technicallyty I moved things around to be more go-like and I made sure that any error that was logged was logged with a stack trace. 

Additionally for errors that are transformed into http responses I made sure that the stack trace is embedded in the body of the response as json.

technicallyty (approved):
LGTM

PR #422, opened 2023-11-13
feat(sign): add eris to sign package.

pyrofolium:
Closes: WORLD-543

## Overview

Adds eris to sign package. 

Note the test. Eris basically works like like a linked list of error wrapping. You pop the head with eris.Unwrap() Any time eris wraps something it creates the stack trace at that exact point. 

I'm going to pause a bit with this ticket here. Make sure I get some reviews in, such that everyone is ok with the current changes.

### DIFF
--- a/sign/sign.go
+++ b/sign/sign.go
@@ -46,8 +46,8 @@ func UnmarshalTransaction(bz []byte) (*Transaction, error) {
 	dec := json.NewDecoder(bytes.NewBuffer(bz))
 	dec.DisallowUnknownFields()
 
-	if err := eris.Wrap(dec.Decode(s), ""); err != nil {
-		return nil, fmt.Errorf("error decoding Transaction: %w", err)
+	if err := dec.Decode(s); err != nil {
+		return nil, eris.Wrap(err, "error decoding Transaction")
 	}
 
 	if err := s.checkRequiredFields(); err != nil {
@@ -101,9 +101,9 @@ func MappedTransaction(tx map[string]interface{}) (*Transaction, error) {
 	}
 	delete(tx, "hash")
 	delete(tx, "body")
-	err = eris.Wrap(mapstructure.Decode(tx, s), "")
+	err = mapstructure.Decode(tx, s)
 	if err != nil {
-		return nil, err
+		return nil, eris.Wrap(err, "error decoding map structure")
 	}
 	s.Body = serializedBody
 	if err := s.checkRequiredFields(); err != nil {
@@ -169,9 +169,8 @@ func sign(pk *ecdsa.PrivateKey, personaTag, namespace string, nonce uint64, data
 	}
 	sp.populateHash()
 	buf, err := crypto.Sign(sp.Hash.Bytes(), pk)
-	err = eris.Wrap(err, "")
 	if err != nil {
-		return nil, err
+		return nil, eris.Wrap(err, "error signing hash")
 	}
 	sp.Signature = common.Bytes2Hex(buf)
 	return sp, nil
--- a/sign/sign.go
+++ b/sign/sign.go
@@ -135,8 +135,8 @@ func normalizeJSON(data any) ([]byte, error) {
 	// sorted keys, the resulting hashes will be different and the signature will fail.
 	// For this reason, we must Unmarshal/Marshal any pre-built JSON bodies to ensure the resulting hashes during
 	// signing match the hash during verification
-	if err := eris.Wrap(json.Unmarshal(asBuf, &asMap), ""); err != nil {
-		return nil, fmt.Errorf("data %q is not valid json", string(asBuf))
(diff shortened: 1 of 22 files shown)


sign/sign.go
```diff
@@ -45,7 +46,7 @@ func UnmarshalTransaction(bz []byte) (*Transaction, error) {
 	dec := json.NewDecoder(bytes.NewBuffer(bz))
 	dec.DisallowUnknownFields()
 
-	if err := dec.Decode(s); err != nil {
+	if err := eris.Wrap(dec.Decode(s), ""); err != nil {
 		return nil, fmt.Errorf("error decoding Transaction: %w", err)
```
technicallyty, line 50: it would be more natural to do:

```go
	if err := dec.Decode(s); err != nil {
		return nil, eris.Wrap(err, "error decoding Transaction")
	}
```

sign/sign.go
```diff
@@ -60,16 +61,16 @@ func UnmarshalTransaction(bz []byte) (*Transaction, error) {
 // faster than using reflection.
 func (s *Transaction) checkRequiredFields() error {
 	if s.PersonaTag == "" {
-		return ErrNoPersonaTagField
+		return eris.Wrap(ErrNoPersonaTagField, "")
```
technicallyty, line 64: why not just change the error definitions to use `eris.New`?

sign/sign.go
```diff
@@ -100,7 +101,7 @@ func MappedTransaction(tx map[string]interface{}) (*Transaction, error) {
 	}
 	delete(tx, "hash")
 	delete(tx, "body")
-	err = mapstructure.Decode(tx, s)
+	err = eris.Wrap(mapstructure.Decode(tx, s), "")
```
technicallyty, line 104: same as above

sign/sign.go
```diff
@@ -123,7 +124,9 @@ func normalizeJSON(data any) ([]byte, error) {
 	}
 	if asBuf == nil {
 		// The given data was neither a string nor a []byte. Just json.Marshal it.
-		return json.Marshal(data)
+		res, err := json.Marshal(data)
+		err = eris.Wrap(err, "")
+		return res, err
```
technicallyty, line 129:
```suggestion
		return res, eris.Wrap(err, "")
```

sign/sign.go
```diff
@@ -133,13 +136,14 @@ func normalizeJSON(data any) ([]byte, error) {
 	// sorted keys, the resulting hashes will be different and the signature will fail.
 	// For this reason, we must Unmarshal/Marshal any pre-built JSON bodies to ensure the resulting hashes during
 	// signing match the hash during verification
-	if err := json.Unmarshal(asBuf, &asMap); err != nil {
+	if err := eris.Wrap(json.Unmarshal(asBuf, &asMap), ""); err != nil {
 		return nil, fmt.Errorf("data %q is not valid json", string(asBuf))
 	}
 
 	normalizedBz, err := json.Marshal(asMap)
+	err = eris.Wrap(err, "")
```
technicallyty, line 144: i think this is unnecessary here

sign/sign.go
```diff
@@ -167,6 +171,7 @@ func sign(pk *ecdsa.PrivateKey, personaTag, namespace string, nonce uint64, data
 	}
 	sp.populateHash()
 	buf, err := crypto.Sign(sp.Hash.Bytes(), pk)
+	err = eris.Wrap(err, "")
 	if err != nil {
 		return nil, err
```
technicallyty, line 176: same here, just move wrapped down into the return

technicallyty (changes requested):

sign/sign.go
technicallyty: ah ok got it

sign/sign.go
pyrofolium: @technicallyty I changed it. I also took a closer look at the readme. 

They do this because some functions return both a result and an error and eris can't wrap it on one line so they wrap it in the return value. 

For functions that only return an error we should wrap it on the same line which is what's going on in my code here.

The eris.Wrap function is explicitly designed to be used this way as it returns a nil if the input is nil. 

We can address this issue in another PR.

sign/sign.go
```diff
@@ -133,13 +135,13 @@ func normalizeJSON(data any) ([]byte, error) {
 	// sorted keys, the resulting hashes will be different and the signature will fail.
 	// For this reason, we must Unmarshal/Marshal any pre-built JSON bodies to ensure the resulting hashes during
 	// signing match the hash during verification
-	if err := json.Unmarshal(asBuf, &asMap); err != nil {
+	if err := eris.Wrap(json.Unmarshal(asBuf, &asMap), ""); err != nil {
```
technicallyty, line 138: lets move this down

technicallyty (approved):
LGTM, just need to move 1 line

PR #423, opened 2023-11-14
fix(cardinal): startup and shutdown race conditions

jerargus:
Closes: WORLD-546

## Overview

Fix some race conditions and shutdown sequencing.

## Brief Changelog

- Shorten the sleep duration while waiting for the World to shut down
- Change the ecs.World tick into an atomic unit64
- Add the concept of a game sequence to cardinal to help with the starting and stopping of the game

## Testing and Verifying

Run `go test --count=10 --race` from the cardinal package and the ecs package to verify there are no race conditions. 

The server and events packages still have race conditions.

### DIFF
--- a/cardinal/world.go
+++ b/cardinal/world.go
@@ -32,13 +32,23 @@ type World struct {
 	server          *server.Handler
 	evmServer       evm.Server
 	gameManager     *server.GameManager
-	isGameRunning   atomic.Bool
 	tickChannel     <-chan time.Time
 	tickDoneChannel chan<- uint64
 	serverOptions   []server.Option
 	cleanup         func()
-	endStartGame    chan bool
+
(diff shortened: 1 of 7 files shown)


cardinal/world.go
```diff
@@ -34,14 +34,24 @@ type World struct {
...
+	endStartGame      chan bool
 }
 
+const (
+	sequencePreStart int32 = iota
+	sequenceStarting
+	sequenceRunning
+	sequenceShuttingDown
+	sequenceShutDown
+)
```
pyrofolium, line 53: This is an enum, there should be a type alias for the atomic and the underlying int32 literals.

pyrofolium (changes requested):

cardinal/world.go
jerargus: I added a gamestage package to wrap the atomic struct with some concrete types.

technicallyty (approved):

PR #424, opened 2023-11-14
feat(cardinal): add eris to evm.

pyrofolium:
Closes: WORLD-538

## Overview

This change adds the eris package to cardinal.evm. 

As mentioned in other related PRs, not all errors are wrapped here because those errors are instantiated in our own libraries. For those errors coming from our own libraries the wrapper will be made inside the library itself which will likely be another PR. 

this is the related library to learn more: https://github.com/rotisserie/eris

### DIFF
--- a/cardinal/evm/server.go
+++ b/cardinal/evm/server.go
@@ -291,7 +291,6 @@ func (s *msgServerImpl) getSignerComponentForAuthorizedAddr(
 	err = errors.Join(err, q.Each(wCtx, func(id entity.ID) bool {
 		var signerComp *ecs.SignerComponent
 		signerComp, err = component.GetComponent[ecs.SignerComponent](wCtx, id)
-		err = eris.Wrapf(err, "error getting signer component with id %d", id)
 		if err != nil {
 			return false
 		}
--- a/cardinal/evm/server.go
+++ b/cardinal/evm/server.go
@@ -8,6 +8,7 @@ import (
 	"net"
 	"os"
 
+	"github.com/rotisserie/eris"
 	zerolog "github.com/rs/zerolog/log"
 	"pkg.world.dev/world-engine/cardinal/ecs/message"
 
@@ -94,7 +95,7 @@ func NewServer(w *ecs.World, opts ...Option) (Server, error) {
 	}
 
 	if !hasEVMTxsOrQueries {
-		return nil, ErrNoEVMTypes
+		return nil, eris.Wrap(ErrNoEVMTypes, "no evm txs or queries")
 	}
 
 	s := &msgServerImpl{txMap: it, queryMap: ir, world: w, port: defaultPort}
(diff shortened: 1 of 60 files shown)


cardinal/evm/server.go
```diff
@@ -285,11 +286,12 @@ func (s *msgServerImpl) getSignerComponentForAuthorizedAddr(
 	wCtx := ecs.NewReadOnlyWorldContext(s.world)
 	q, err := wCtx.NewSearch(ecs.Exact(ecs.SignerComponent{}))
 	if err != nil {
-		return nil, err
+		return nil, eris.Wrap(err, "error creating search")
 	}
-	err = q.Each(wCtx, func(id entity.ID) bool {
+	err = errors.Join(err, q.Each(wCtx, func(id entity.ID) bool {
```
technicallyty, line 291: why do we need to join the previous error and the error returned from `Each`? the initial error is already handled on line 289.

cardinal/evm/server.go
```diff
@@ -285,11 +286,12 @@ func (s *msgServerImpl) getSignerComponentForAuthorizedAddr(
...
 		var signerComp *ecs.SignerComponent
 		signerComp, err = component.GetComponent[ecs.SignerComponent](wCtx, id)
+		err = eris.Wrapf(err, "error getting signer component with id %d", id)
```
technicallyty, line 294: we don't need to wrap the error here, we aren't using it anywhere

pyrofolium: You're right, the error doesn't need to be wrapped because it's already wrapped inside of `GetComponent`

However this error is used. It's subtle. err comes from the outside scope. It's set here and then joined with the top level error generated from q.Each

cardinal/evm/server.go
pyrofolium: You have to look ahead. The error by wCtx is overwritten by the error from the closure inside of q.Each. It's happening out of order here.

let me illustrate what was happening before:


```go
q, err := wCtx.NewSearch(ecs.Exact(ecs.SignerComponent{})) //1. Err is created here. 
	if err != nil {
		return nil, eris.Wrap(err, "error creating search") 
	}
	err = q.Each(wCtx, func(id entity.ID) bool { //3. error is overwritten here for a 3rd time. 
		var signerComp *ecs.SignerComponent
		signerComp, err = component.GetComponent[ecs.SignerComponent](wCtx, id) //2. error is overwritten here. 
		if err != nil {
			return false
```

pyrofolium: My code preserves the error created by GetComponent which was previously getting swallowed.

technicallyty: ah i see. i think the code becomes a little unintuitive this way. i think separating the error from `.Each` and `.GetComponent` and handling them separately is a bit more straightforward. its a bit more code, but worthwhile imo for readability

pyrofolium: Agreed. This does break go convention though, but idc care for convention. I'll change it tmrw.

technicallyty: i think its actually more go-convention-ish, since this way we will handle each error **explicitly**, which the gopher god prefers. with the join, we handle both each and an inner error as one error

pyrofolium: It's possible for both to occur though. Usually one error happens before another error so the second error doesn't get a chance to occur. Here we have a possibility of two errors that occured. When there are two errors, one will inevitably be swallowed if we don't join them. 

I think the problem is in the Each function. It doesn't allow errors to be propagated out of the closure. It should return a bool, err.

It's fine fow now I'll just do it on the agreed way.

technicallyty (approved):

PR #425, opened 2023-11-14
feat(cardinal): add eris to cardinal server

pyrofolium:
Closes: WORLD-537

## Overview

This adds eris to cardinal server. 

Info:
This is a sub issue of https://linear.app/arguslabs/issue/WORLD-451/find-a-way-to-get-all-errors-to-be-paired-with-stack-trace

Relevant conversation here: https://argus-labs.slack.com/archives/C03G859K5TQ/p1699905811770509

### DIFF
--- a/cardinal/server/debug.go
+++ b/cardinal/server/debug.go
@@ -2,7 +2,6 @@ package server
 
 import (
 	"encoding/json"
-	"errors"
 
 	"github.com/go-openapi/runtime/middleware/untyped"
 	"pkg.world.dev/world-engine/cardinal/ecs"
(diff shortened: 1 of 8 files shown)


cardinal/server/debug.go
```diff
@@ -25,8 +26,8 @@ func (handler *Handler) registerDebugHandlerSwagger(api *untyped.API) {
 				result := make(DebugStateResponse, 0)
 				search := ecs.NewSearch(filter.All())
 				wCtx := ecs.NewReadOnlyWorldContext(handler.w)
-
-				err := search.Each(wCtx, func(id entity.ID) bool {
+				var err error
+				err = errors.Join(err, search.Each(wCtx, func(id entity.ID) bool {
```
technicallyty, line 30: redundant join

namrapatel (approved):

cardinal/server/debug.go
pyrofolium: Yeah you're right there was a bug here, but I don't think it was what you were thinking. 

Same thing happening here:
https://github.com/Argus-Labs/world-engine/pull/424#discussion_r1395007644

The bug here is that I'm using `:=` instead of `=` and that is overwriting the err value from outer-scope.

PR #442, opened 2023-11-16
fix(relay): make claim-key use logCode instead of returning raw errors

technicallyty:
## Overview

it seems that returning raw errors, instead of using nakama's `runtime.NewError` will cause the server to reply with a 500, this in turn makes the javascript code in WebGL implode. not good!

### DIFF
--- a/relay/nakama/allowlist.go
+++ b/relay/nakama/allowlist.go
@@ -137,7 +137,7 @@ func claimKeyRPC(ctx context.Context, logger runtime.Logger, _ *sql.DB, nk runti
 ) {
 	userID, err := getUserID(ctx)
 	if err != nil {
-		return logCode(logger, Internal, "unable to get userID")
+		return logCode(logger, Internal, "unable to get userID: %v", err)
 	}
 
 	// if this user is already verified,
@@ -149,7 +149,7 @@ func claimKeyRPC(ctx context.Context, logger runtime.Logger, _ *sql.DB, nk runti
 	var ck ClaimKeyMsg
 	err = json.Unmarshal([]byte(payload), &ck)
 	if err != nil {
-		return logCode(logger, Internal, "unable to unmarshal payload")
+		return logCode(logger, InvalidArgument, "unable to unmarshal payload: %v", err)
 	}
 	if ck.Key == "" {
 		return logCode(logger, InvalidArgument, "no key provided in request")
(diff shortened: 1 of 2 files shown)


namrapatel (approved):

relay/nakama/allowlist.go
```diff
@@ -132,41 +132,41 @@ type ClaimKeyRes struct {
...
 }
 
-func claimKeyRPC(ctx context.Context, _ runtime.Logger, _ *sql.DB, nk runtime.NakamaModule, payload string) (
+func claimKeyRPC(ctx context.Context, logger runtime.Logger, _ *sql.DB, nk runtime.NakamaModule, payload string) (
 	string, error,
 ) {
 	userID, err := getUserID(ctx)
 	if err != nil {
-		return "", err
+		return logCode(logger, Internal, "unable to get userID")
```
jerargus, line 140: We're losing the contents of the error here. It can be surfaced to the client with:
`logCode(logger, Internal, "unable to get userID: %v", err)`

If we don't want to surface the specific error reason, we can just log it here (or extend logCode to take in an error that will just be logged then discarded).

relay/nakama/allowlist.go
```diff
@@ -132,41 +132,41 @@ type ClaimKeyRes struct {
...
 	if err == nil {
-		return "", eris.Wrap(errors.Join(ErrAlreadyVerified, err), "")
+		return logCode(logger, AlreadyExists, "user already verified with beta key")
 	}
 
 	var ck ClaimKeyMsg
 	err = json.Unmarshal([]byte(payload), &ck)
 	if err != nil {
-		return "", eris.Wrap(err, "")
+		return logCode(logger, Internal, "unable to unmarshal payload")
```
jerargus, line 152: The client is the source of the payload, correct? This looks like an InvalidArgument error. And the contents of error should be returned to the user.

relay/nakama/allowlist.go
```diff
@@ -132,41 +132,41 @@ type ClaimKeyRes struct {
...
 	ck.Key = strings.ToUpper(ck.Key)
 	err = claimKey(ctx, nk, ck.Key, userID)
 	if err != nil {
-		return "", err
+		return logCode(logger, Internal, fmt.Sprintf("unable to claim key: %s", err.Error()))
 	}
 	err = writeVerified(ctx, nk, userID)
 	if err != nil {
-		return "", err
+		return logCode(logger, Internal, "server could not save user verification entry. please try again"+err.Error())
```
jerargus: nit: there's a mix of using "+" and "%s" to include the error text.

relay/nakama/allowlist.go
technicallyty: went with surfacing it

relay/nakama/allowlist.go
technicallyty: yep, changed to InvalidArgument

relay/nakama/main.go
```diff
@@ -427,13 +427,13 @@ func initCardinalEndpoints(logger runtime.Logger, initializer runtime.Initialize
 func logCode(logger runtime.Logger, code int, format string, v ...interface{}) (string, error) {
 	err := eris.Errorf(format, v...)
 	logger.Error(eris.ToString(err, true))
-	return "", runtime.NewError(eris.ToString(err, true), code)
+	return "", runtime.NewError(err.Error(), code)
```
technicallyty, line 430: changing this here because i dont think we wanna give clients the stack trace. just a simple error message.

pyrofolium: hmm this will be extremely convenient for development. We should have dev mode and prod mode for this.

pyrofolium (changes requested):

technicallyty:
the error is already logged though with the stack trace @pyrofolium

pyrofolium:
> the error is already logged though with the stack trace @pyrofolium

I still think it's very convenient as opposed to digging through a bunch of log lines. I think we can forward it to another issue though.

pyrofolium (approved):

jerargus (approved):

relay/nakama/main.go
technicallyty: https://linear.app/arguslabs/issue/WORLD-573/make-a-nakama-dev-mode-that-returns-the-stack-traced-error

PR #445, opened 2023-11-17
feat(nakama): investigate nakama and ensure we use runtime.newerror for all

pyrofolium:
Closes: WORLD-572

decomposes logError and logCode into smaller granular functions for more convenience. 
Makes it so original error is still parsed with error message. 
replace Grpc codes with errors that won't trigger a 500 http error. 
Adds debug mode into environment. 
By default nakama will not return request with stack trace. If environment is set to debug mode it will. 
all errors are replaced with nakama spawned errors with error codes.


relay/nakama/main.go
```diff
@@ -39,6 +39,12 @@ const (
 )
 
 type receiptChan chan *Receipt
+type DevMode int
+
+const (
+	Debug DevMode = iota
```
technicallyty, line 45: what are your thoughts on if we just simplify this to `DebugEnabled bool`?

pyrofolium: idc, I did it this way because it's a bit more bijective with an environment variable. bool only has two possibilities but an environment variable can be any string.

technicallyty: i think as a dev id prefer something like:

`export ENABLE_DEBUG=true`

if the env is enitrely unset, we dont use debug

pyrofolium: sure

pyrofolium:
whoops this was on auto merge.

PR #446, opened 2023-11-17
Brian/world 572 investigate nakama and ensure we use runtimenewerror for all

pyrofolium:
Previous PR for world-572 went up early. @technicallyty


pyrofolium:
nm false alarm. It got in.

PR #448, opened 2023-11-17
refactor(cardinal): Rename message ForEach to Each to match search iterator

jerargus:
Closes: WORLD-542

## Overview

Rename the message iterator to Each to match the search iterator

## Testing and Verifying

No functional changes


technicallyty (approved):
nice

PR #449, opened 2023-11-18
feat(cardinal): accept empty namespace and signature when signature verification is d…

jerargus:
…isabled

Closes: WORLD-553

## Overview

- When signature verification is disabled, accept empty namespace and signature fields. 
- Add additional testing so that we get a 401 instead of a 500 when namespace or signature fields are missing.

## Testing and Verifying

Unit tests added/updated.

### DIFF
--- a/cardinal/server/server_test.go
+++ b/cardinal/server/server_test.go
@@ -610,7 +610,7 @@ func TestCanCreateAndVerifyPersonaSigner(t *testing.T) {
 	assert.NilError(t, err)
 }
 
-func TestSigVerificationChecksNamespace(t *testing.T) {
+func TestSigVerificationChecksNamespaceAndSignature(t *testing.T) {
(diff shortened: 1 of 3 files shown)


cardinal/server/server_test.go
```diff
@@ -610,7 +610,7 @@ func TestCanCreateAndVerifyPersonaSigner(t *testing.T) {
 	assert.NilError(t, err)
 }
 
-func TestSigVerificationChecksNamespace(t *testing.T) {
+func TestSigVerificationChecksNamespaceAndSignature(t *testing.T) {
```
technicallyty, line 613: this test is getting pretty long, might be nice to convert this to a table-driven test

technicallyty (approved):

cardinal/server/server_test.go
jerargus: done

PR #450, opened 2023-11-18
feat(cardinal): replace assert.NilError with testutils.AssertNilErrorWithTrace

pyrofolium:
Closes: WORLD-578

This one is huge. But read below to understand what was done. You only really need to review a couple things. The other changes are mechanical and straightforward. 

Originally there were two testutil libraries. 

`ecs/internal/testutil` and `cardinal/testutils`

Both of these libraries cannot be imported everywhere because of circular dependencies. They are only imported and used in specific places. This means certain tests can't use certain utilities. 

I renamed both of these libraries to `ecs/internal/ecstestutils` and `cardinal/cardinaltestutils`

I created a new package called `cardinal/testutils` and in that package I placed `cardinal.AssertNilErrorWithTrace` which will display the stacktrace of the eris error should the assertion fail. This package should NOT depend on any other `ecs` or `cardinal` package at all. Also this package can be eventually moved to the top level which will be some ticket for another time. 

Finally I replaced Every `assert.NilError` in cardinal with `cardinal.AssertNilErrorWithTrace`


cardinal/cardinaltestutils/cardinaltestutils.go
technicallyty, line 1: can we just rename the existing `testutils` directory to `assert` - since thats all it really has now. 

this package & filename can then be named `testutils`.

technicallyty (commented):
did you investigate if any of the other assertion libraries don't swallow the stack trace? (i.e. something from https://github.com/stretchr/testify)? it would be nice if we could use just _one_ library for our tests. with this direction, we have to use the testutils, and then the the assert package for everything else, which seems not so ideal.

cardinal/testutils/testutils.go
```diff
@@ -1,257 +1,12 @@
...
-		})
-
-		startTickCh <- time.Now()
-		<-doneTickCh
-	}
-
-	return world, doTick
+func AssertNilErrorWithTrace(t assert.TestingT, err error, args ...interface{}) {
+	args = append([]interface{}{eris.ToString(err, true)}, args...)
+	assert.NilError(t, err, args...)
```
technicallyty, line 11:
```suggestion
	assert.NilError(t, err, eris.ToString(err, true), args)
```

cardinal/cardinaltestutils/cardinaltestutils.go
technicallyty: id prefer if we could also just alias all of `gotest.tools/v3/assert`'s functions in this package as well. its going to be annoying to have to juggle these two separate things.

jerargus:
I don't agree with this PR. The eris.ToString(err, true) is a useful tool, but we 1) shouldn't create a customer wrapper around that code and 2) shouldn't insert it into every existing assert. 

If someone is investigating a test failure, they can add eris.ToString to the assert message on a case by case basis (including merging it into main).

I think unifying the two testutil packages is worth investigating. It's not clear to me if we want 1 single testutil package, or two packages (one focused on helping cardinal clients test their cardinal based games, and one for the infra team to test cardinal internals).

I think we should discuss this after standup, or in person tomorrow.

cardinal/testutils/testutils.go
technicallyty: nvm

pyrofolium:
> did you investigate if any of the other assertion libraries don't swallow the stack trace? (i.e. something from https://github.com/stretchr/testify)? 

Unlikely. The stack trace is not part of any public interface. You need eris.ToString to fetch it. Additionally why would the assert library make itself compatible with eris? I can take a look but I think this is unlikely.

pyrofolium:
>If someone is investigating a test failure, they can add eris.ToString to the assert message on a case by case basis (including merging it into main).

This is the problem. We've already had several errors where we had no idea where it was coming from. I believe Tyler hit one a while back where we literally just didn't know what was causing it. I wasn't too clear what was swallowing the stack trace of the assert, but we weren't getting either the stack trace of the error or the assert. I'm thinking this is an issue with the go test suite. 

The case by case basis idea is something I also disagree with. Almost every programming language delivers stack trace with an exception, it's never a case by case basis, instead it's case by case whether to swallow the stack trace. 

Additionally swallowing the stack trace at the unit test layer is contradictory to the overall effort. What was the point of making stack traces saved in all errors if they aren't used ever?

cardinal/cardinaltestutils/cardinaltestutils.go
pyrofolium: >id prefer if we could also just alias all of gotest.tools/v3/assert's functions in this package as well. its going to be annoying to have to juggle these two separate things.

I'm ok with this. But this would be another PR imo. It would also be rather big.

pyrofolium:
> I think unifying the two testutil packages is worth investigating. It's not clear to me if we want 1 single testutil package, or two packages (one focused on helping cardinal clients test their cardinal based games, and one for the infra team to test cardinal internals).

We already had two testutils packages. I just renamed them. That's all, so this PR is about that renaming. 

I am not unifying those packages, apologies for not being clear about that. If unification is to occur that will be a separate issue to consider in another PR that is Independent of what's going on here. Let's not let this get too into that topic here. 

What's going on here is I created a new utility package that has zero dependencies on any package in world cardinal. That means this testutils package can be used in evm, server, nakama, cardinal and everywhere. The question is whether the creation of a test utility package independent of our own internal packages is justified. 

The reason why this package was created is because of necessity. For me to define a utility one time and usable in All packages requires me to have a package that is independent of cardinal AND server and everything. Otherwise I would have to define the utility twice to avoid circular dependencies. 

If it wasn't for the circular dependency thing all testutils would likely be living in one package already.

pyrofolium:
Guys I can sort of predict how a meeting is going to go. Nothing personal but I literally don't want to meet as I felt the environment was hostile last time.

There are times where we fundamentally disagree and we will not come to an amicable agreement. I think this is one of those times. 

I do not think I can change your minds in a meeting so I don't want to waste our time and have that meeting and I want to throw it to a vote. However, If you think you can change my mind or the minds of other people, of course feel free to say so and we can still have that meeting.

From the conversation here this is what it comes down two possibilities:

- @technicallyty suggestion of wrapping the assert library in a testutils (we actually have two assert libraries, I would wrap both)
   Brian.

- @jerargus suggestion of leaving everything as is and closing this PR. 
   Jeremy.
   
tylers suggestion has my vote. Not sure how how strongly tyler feels about his suggestion so I'll mark that as just me. 
Jeremy feels pretty strongly about his stance so I'll mark that for him. lmk if this is accurate. 

Let's just vote and it will be decided quickly.

pyrofolium:
> Guys I can sort of predict how a meeting is going to go.
> 
> There are times where we fundamentally disagree and we will not come to an amicable agreement. I think this is one of those times.
> 
> I do not think I can change your minds in a meeting so I don't want to waste our time and have that meeting and I want to throw it to a vote. However, If you think you can change my mind or the minds of other people, of course, feel free to say so and we can still have that meeting but I also think this is unlikely. If what we stated here doesn't change any of your opinions let's just move forward.
> 
> From the conversation here this is what it comes down two possibilities:
> 
> * @technicallyty suggestion of wrapping the assert library in a testutils (we actually have two assert libraries, I would wrap both)
>   Brian.
> * @jerargus suggestion of leaving everything as is and closing this PR.
>   Jeremy.
> 
> tylers suggestion has my vote. Not sure how how strongly tyler feels about his suggestion so I'll mark that as just me. Jeremy feels pretty strongly about his stance so I'll mark that for him. lmk if this is accurate.
> 
> Let's just vote and it will be decided quickly.

Ok there was a small discussion and I believe we are going with the first option, but with renaming testutils to assert and just replacing all our usages of assert.

pyrofolium:
See discussion. Not going this route.

PR #451, opened 2023-11-18
feat(cardinal): replace all assert.ErrorIs and testify.assert.Error

pyrofolium:
Closes: WORLD-579

Replaces ErrorIs and testify.assert.Error With versions that will display stack trace. on error. This PR is stacked on world-578, please review that one before this one.


pyrofolium:
No longer relevant.

PR #455, opened 2023-11-21
feat(cardinal): create assert library that handles eris stack traces

pyrofolium:
Closes: WORLD-583

Big PR please read stuff below. I also commented parts for clarity. 

This ticket has two parts:
**Part 1**
 - Create a new assert package wrapper around gotest assert and testify assert
     - There are some collisions between these two libraries. These are noted in the comments. Some collisions are used in our tests too. For these cases I renamed the functions that were on the testify side. 
     - Every reference to gotest assert or testify assert is replaced by this new package. 
     
Most changes you will see involve the above. Changing the assert library and renaming some assert functions covered the majority of the files. 

**Part 2**
- update e2e nakama test. Additionally update the tests to always use a local version of cardinal. 
- fix a nig bug in nakama.


pyrofolium:
Local lint is failing with strange errors. drafting this in case it fails in CI too.

cardinal/server/tx.go
```diff
@@ -97,9 +97,9 @@ func (handler *Handler) registerTxHandlerSwagger(api *untyped.API) error {
 		payload, sp, err := handler.getBodyAndSigFromParams(params, true)
 		if err != nil {
 			if eris.Is(err, eris.Cause(ErrInvalidSignature)) || eris.Is(err, eris.Cause(ErrSystemTransactionRequired)) {
-				return middleware.Error(http.StatusUnauthorized, err), nil
+				return middleware.Error(http.StatusUnauthorized, eris.ToString(err, true)), nil
 			}
-			return nil, err
+			return middleware.Error(http.StatusInternalServerError, eris.ToJSON(err, true)), nil
 		}
```
pyrofolium, line 103: This updates it so that error stack traces will be forwarded to nakama.

docker-compose.yml
```diff
@@ -107,13 +107,13 @@ services:
     container_name: test_game
     environment:
       - REDIS_ADDR=redis:6379
-      - NAMESPACE=TESTGAME
```
pyrofolium, line 110: Above is a bug.

docker-compose.yml
```diff
@@ -107,13 +107,13 @@ services:
...
+      - CARDINAL_NAMESPACE=TESTGAME
       - SHARD_SEQUENCER_ADDR=chain:9601
       - BASE_SHARD_ADDR=chain:9090
       - ENABLE_ADAPTER=${ENABLE_ADAPTER:-true}
     build:
-      context: internal/e2e/tester/cardinal
-      dockerfile: Dockerfile
+      context: ./
+      dockerfile: internal/e2e/tester/game/Dockerfile
     restart: unless-stopped
```
pyrofolium, line 117: this keeps the directory structure of the project exactly the same in the docker instance. 

The reason why we do this is because the project now references local packages by using paths relative to the go.mod file. Thus if we change the directory structure in the docker the go.mod will be referencing wrong paths inside the docker. 

Additionally the game docker has been changed such that the context is the root project directory. It can only now be run in the context of the docker-compose from the root.

relay/nakama/cardinal.go
```diff
@@ -89,9 +89,23 @@ func doRequest(req *http.Request) (*http.Response, error) {
...
+			return nil, eris.Wrapf(err, "failed reading body in request, status code: %d", statusCode)
+		}
+		return nil,
+			eris.Errorf(
+				"error to url: %s, with request body: %s, got response of %d: %s",
+				req.URL,
+				string(reqBuf),
+				statusCode,
+				string(buf))
 	}
```
pyrofolium, line 109: This just adds more information to the error. for debugging.

relay/nakama/cardinal.go
```diff
@@ -101,6 +115,14 @@ func cardinalCreatePersona(ctx context.Context, nk runtime.NakamaModule, persona
...
 ) {
+	defer func() {
+		if r := recover(); r != nil {
+			txHash = ""
+			tick = 0
+			err = eris.Errorf("a panic occurred in nakama in the function, cardinalCreatePersona:, %s", r)
+		}
+	}()
+
 	signerAddress := getSignerAddress()
```
pyrofolium, line 126: Nakama will swallow a panic. There was a bug in this function that was just closing the connection. This catches the panic and turns it into an error.

relay/nakama/main.go
```diff
@@ -418,9 +418,7 @@ func initCardinalEndpoints(logger runtime.Logger, initializer runtime.Initialize
 				if err != nil {
 					return logErrorWithMessage(logger, err, "request failed for endpoint %q", currEndpoint)
 				}
-				if resp.Body != nil {
-					defer resp.Body.Close()
-				}
+				defer resp.Body.Close()
 				if resp.StatusCode != http.StatusOK {
```
pyrofolium, line 422: This was not the bug. fyi. changed it back.

cardinal/cardinal_test.go
```diff
@@ -9,17 +11,64 @@ import (
...
+		http.MethodPost, "http://localhost:4040/tx/persona/create-persona", bytes.NewBuffer(bodyBytes))
+	assert.NilError(t, err)
+	req.Header.Set("Content-Type", "application/json")
+	resp, err := client.Do(req)
+	assert.NilError(t, err)
+	assert.Equal(t, resp.StatusCode, http.StatusOK)
+	err = world.ShutDown()
+	assert.NilError(t, err)
+}
+
```
pyrofolium, line 71: Just a new test to help with fixing some of the existing tests.

internal/e2e/tester/game/go.mod
```diff
@@ -13,7 +13,9 @@ replace (
 	github.com/syndtr/goleveldb => github.com/syndtr/goleveldb v1.0.1-0.20210819022825-2ae1ddf74ef7
 )
 
-require pkg.world.dev/world-engine/cardinal v0.1.47-alpha.0.20231106232923-a4d2b1526011
+replace pkg.world.dev/world-engine/cardinal => ../../../../cardinal
+
```
pyrofolium, line 17: replace makes golang use the local package.

cardinal/config.go
```diff
@@ -12,6 +12,7 @@ const (
 	DefaultMode          = ModeDev
 	DefaultNamespace     = "world-1"
 	DefaultRedisPassword = ""
+	DefaultRedisAddress  = "redis:6379"
```
technicallyty, line 15: i think this should stay localhost

internal/e2e/tester/game/main.go
```diff
@@ -14,41 +17,64 @@ import (
...
 	err = cardinal.RegisterMessages(world, msg.JoinMsg, msg.MoveMsg)
 	if err != nil {
-		log.Fatal(err)
+		log.Fatal(err, eris.ToString(err, true))
+	}
+
+	err = cardinal.RegisterQueryWithEVMSupport[query.LocationRequest, query.LocationReply](
+		world,
+		"location",
+		func(ctx cardinal.WorldContext, req *query.LocationRequest) (*query.LocationReply, error) {
```
technicallyty, line 48: could also just have this be an exported function from `query/location.go`

relay/nakama/cardinal.go
technicallyty: where is the panic occurring? do we know whats triggering this?

relay/nakama/main.go
```diff
@@ -318,7 +318,7 @@ func handleClaimPersona(ptv *personaTagVerifier, notifier *receiptNotifier) naka
 			}
 			return logErrorWithMessageAndCode(
 				logger,
-				eris.New("could not set personaTag assignment"),
+				eris.Wrap(err, "could not set personaTag assignment"),
```
technicallyty, line 321: what error is being wrapped here?

relay/nakama/cardinal.go
pyrofolium: The bug is gone, but this is still here as safety. Actually ALL of our functions should have this as nakama will just hang up the connection. It may get logged in the output though.

cardinal/config.go
pyrofolium: This will break the e2e tests. Unless you want the docker-compose to specify this.

relay/nakama/cardinal.go
pyrofolium: @technicallyty the panic is here.

cardinal/config.go
technicallyty: yeah lets have it specified in docker-compose

internal/e2e/tester/game/main.go
pyrofolium: That would be just a function that takes the world. Sure. I don't mind this.

technicallyty: should just be the same function, no? they get access to WorldCtx

technicallyty: oh nvm i see how you did it, that works for me 🤙🏻

pyrofolium: RegisterQueryWithEVMSupport can only be called from main because it needs to be in the context of world. World is the only reason why I moved everything here.

technicallyty: yep yep, this way works, i originally meant to just export the _handler_ logic (the query fn itself) and then pass that in. 

but this way also works!

relay/nakama/main.go
pyrofolium: ur right this is wrong.

cardinal/config.go
pyrofolium: The environment variable was already set. This is a bug. What was originally specified in docker-compose was REDIS-ADDR, but the code looks for REDIS_ADDRESS.

technicallyty (approved):
LGTM

cardinal/go.mod
```diff
@@ -11,6 +11,7 @@ replace (
 
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
 	github.com/syndtr/goleveldb => github.com/syndtr/goleveldb v1.0.1-0.20210819022825-2ae1ddf74ef7
+	pkg.world.dev/world-engine/assert => ../assert
```
technicallyty, line 14: i dont think we should do this replace here

pyrofolium: why.

technicallyty: id prefer to use tagged versions (or a specific commit at worst case). we don't use replace for the sign or evm modules

pyrofolium: hm. Why is that way better? Wouldn't it be more convenient to always just use the local packages? That way unit tests and updates always are fresh and up to date? 

Right now that package doesn't actually exist. So it has to be used like this for now. If I am to change it, it will be in another PR.

technicallyty: > Wouldn't it be more convenient to always just use the local packages?

this is what go.work does

> Right now that package doesn't actually exist. So it has to be used like this for now. If I am to change it, it will be in another PR.

you can do it from this PR, doesn't have to be another - either grab it via commit hash, or push a new tag from `assert` package

pyrofolium: hm seems to work with tests.

PR #458, opened 2023-11-22
fix(evm): remove conditional router start

technicallyty:
## Overview

Router now looks up namespaces from storage, so we don't need to conditionally start it based off an env var.


pyrofolium (approved):
lgtm

PR #459, opened 2023-11-22
feat(cardinal): schema serialization and comparison. 

pyrofolium:
Closes: WORLD-595

This adds the ability to serialize the schema of a component in a json format called "json schema"

I use a library called jsonschema to do this. 

I compare equality between these schemas utilizing a package called jsondiff. 

There are two functions. One function serializes components into jsonschema bytes
The other function compares a component with a jsonschema. 

This is the first part of a multiple part jsonschema checking feature.


cardinal/ecs/component/metadata/component_metadata.go
```diff
@@ -135,3 +137,24 @@ func WithDefault[T any](defaultVal T) ComponentOption[T] {
...
+		return false, eris.Wrap(err, "")
+	}
+	patch, err := jsondiff.CompareJSON(componentSchemaBytes, jsonSchemaBytes)
+	if err != nil {
+		return false, eris.Wrap(err, "")
+	}
+	if patch.String() == "" {
+		return true, nil
+	}
+	return false, nil
```
technicallyty, line 159: optional nit: happy path as last return would be nice, so we can follow the pattern of "if something is wrong, return false, nil"

technicallyty (approved):
lgtm, one nit

PR #461, opened 2023-11-25
refactor(cardinal): remove codec

smsunarto:
Closes: WORLD-602

## Overview

The codec folder is so simple and unopinionated, we should just get people to use json marshal/unmarshal (remembering to use goccy/go-json instead of the default json stdlib) instead of having our own mini wrapper package.

This PR removes the codec folder in favor of inlining it directly.

## Brief Changelog

- Refactor all code that uses codec package to use goccy/go-json directly
- Remove codec package

## Testing and Verifying

- This change is already covered by existing tests.


smsunarto:
* main
  * **PR #461** <a href="https://app.graphite.dev/github/pr/Argus-Labs/world-engine/461?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/Argus-Labs/world-engine/461?utm_source=stack-comment).

technicallyty (changes requested):
i dont agree with what this PR is doing. Sure - the codec package was simple. However, it came with the advantage that if we wanted a different encoding scheme, all we do is change the encoding package implementation. 

with this PR, we'd have to go in and change every call to json.Unmarshal/Marshal, instead of just doing it one singular place.

this has happened already, where we had a PR changing from std go-lib encoding/json to goccy. what happens if theres yet a faster impl than goccy? we have to go and change all the imports instead of just changing `/codec`.

smsunarto:
> i dont agree with what this PR is doing. Sure - the codec package was simple. However, it came with the advantage that if we wanted a different encoding scheme, all we do is change the encoding package implementation.
> 
> with this PR, we'd have to go in and change every call to json.Unmarshal/Marshal, instead of just doing it one singular place.
> 
> this has happened already, where we had a PR changing from std go-lib encoding/json to goccy. what happens if theres yet a faster impl than goccy? we have to go and change all the imports instead of just changing `/codec`.

The proper solution here is to define a codec for each type that necessitates marshaling/unmarshaling (defined in the type def files within `/types` package folder made in the other PR). Why?
- Some types need stricter validation on decoding. i.e. does the json filled all the field for the type? some types do, some types don't so we can enforce it with a catch all codec like this. Some types need all field enforced, some have optional fields, what do?
- It makes benchmarking easier. If let's say we are considering a faster codec mechanism as you said, this can't be benched in a vacuum because "fast" is often contextual. What's fast for a certain payload might not be fast for another payload. A thin JSON wrapper codec like this is not helpful because it's difficult to isolate how it's used and bench that directly.

The upside on the "easier switch" doesn't stand either for the following reason:
- In an ideal switch where it's a drop in replacement (identical API, etc) like it was for goccy, then replacing it is trivial. I can just install the package and replace+all imports accordingly. As a matter of fact, I can't do this with codec because some places might not be using the codec package because it needs encode/decode instead of marshal/unmarshal, idk. 
- There's no guarantee that we are going to use the same codec everywhere. We currently use JSON for API calls and storing data in Redis. What if we want to switch codec only for one of them? For instance, I want to consider switching the redis components codec to MessagePack since it is faster and has a lower memory footprint than JSON. The only reason why we don't switch the API codec to MessagePack too is because it lacks ergonomics when interfacing with client (people are used to calling API with JSON).

smsunarto:
> Validation logic doesn't need to be at the codec level, for example, in the sign package, we do validation after decoding.

> i'd argue that we can have special cases for these types of things within a codec package -> i.e EncodeComponent, DecodeTransaction. Or, just do it inline. But addressing that can be done when the issue arises.

Well my point here is that the correct way to do this would be to the implement the Unmarshaler/Marshaler interface (https://pkg.go.dev/encoding/json#Unmarshaler) directly as the struct method instead of having a `codec` package. This guarantees that json.Marshal/Unmarshal from/to the struct will always use that validation and you can just use the regular json.Marshal/Unmarshal functions instead of calling one wrapped by a codec package.

Example:
Implementing custom unmarshaler in type def
```
type MyThing struct {
  Thing string
}
func (MyThing) UnmarshalJSON(b []byte) error {
  var mt MyThing
  if err := json.Unmarshal(b, &mt); err != nil {
   return err
  }
  
  mt.Thing = strings.ToUpper(mt.Thing)
  return nil
}
```

Unmarshaling --> Notice that this just calls the regular json function instead of having to use a "codec". The unmarshaler interface is made for this usecase. 
```
func DoTheDance(jsonBytes []byte) (MyThing, error) {
  var u MyThing
  err := json.Unmarshal(jsonBytes, &u)
  if err != nil {
    return nil, err
  }
  
  fmt.Println(u.Thing) //Prints UWOTM8
  return u, nil
}
```

> not sure im understanding this point. Doesn't having a codec package ensure consistency more than doing it inline?

No? Let's say i'm trying to unmarshal into a Message struct. What's stopping me from using the codec package to unmarshal into the Message struct and forgetting to validate?

> Disagre with this - just because we wrap the underlying encoding/decoding calls into a package, doesn't mean it makes it more difficult to benchmark how a small vs large component might perform in encoding/decoding.

This is a strawman of my original point -- what i want to do is to inline through a json marshal/unmarshaler on the struct level and that's the benefit i listed. With a struct with custom marshal/unmarshal defined, If i want to benchmark marshal/unmarshal for many different struct type, then i can just make the dummy structs then marshal/unmarshal both ways. 

However, there's no easy way to bench the generic codec for different kind of type/struct right now because it's not guaranteed that there is an exported marshal/unmarshal function for it (and i would have to write them myself now if i want to).

technicallyty:
Right, what you listed is pretty useful for structs where you control the fields and need granular control over how its marshaled/unmarshled. However, the code that was changed here was for our wrapper/generic types, where we don't even know what we're marshaling/unmarshaling.

for example:

```
func (c *componentMetadata[T]) Encode(v any) ([]byte, error) {
	bz, err := json.Marshal(v)
	if err != nil {
		return nil, eris.Wrap(err, "")
	}
	return bz, nil
}
```

this is useful if `v` has `MarshalJSON` defined. However, we have no idea of `v`'s structure, so we cannot do any validation logic. 

same thing here:

```
func (t *MessageType[In, Out]) Encode(a any) ([]byte, error) {
	bz, err := json.Marshal(a)
	if err != nil {
		return nil, eris.Wrap(err, "")
	}
	return bz, nil
}
```

we don't know what `a` is here. what are we validating? 

> No? Let's say i'm trying to unmarshal into a Message struct. What's stopping me from using the codec package to unmarshal into the Message struct and forgetting to validate?

What's stopping one from forgetting to add validation code in the interface functions mentioned above? but as i said before, im not sure what validation code we're looking to add, as most of the code changed here is unaware of the structure its marshaling/unmarshaling.

>  what i want to do is to inline through a json marshal/unmarshaler on the struct level and that's the benefit i listed. With a struct with custom marshal/unmarshal defined, If i want to benchmark marshal/unmarshal for many different struct type, then i can just make the dummy structs then marshal/unmarshal both ways. 


It seems to me this is possible with the codec package, but maybe i need to see the impl described vs how it can be done today to understand.

smsunarto:
> Right, what you listed is pretty useful for structs where you control the fields and need granular control over how its marshaled/unmarshled. However, the code that was changed here was for our wrapper/generic types, where we don't even know what we're marshaling/unmarshaling.

Validation would happen with decode not encode, since there's not really anything we can validate on encode (we are turning struct to json)

On decode, on the example of MessageType, you can use https://github.com/go-playground/validator to validate that the 

```
func (t *MessageType[In, Out]) Decode(bz []byte) (any, error) {
	in := new(In)
	err := json.Unmarshal(bz, in)
	if err != nil {
		return *in, eris.Wrap(err, "")
	}

    err = validator.New().Struct(in)
    if err != nil {
        log.Fatalf("Validation failed due to %v\n", err)
    }
	return *in, nil
}
```

On a side note, we can be more strict with the return value here and make it be `(*In, error)` instead of `(any, error)`

jerargus:
I generally agree with Tyler in that I think the codec package is fine as is. Although I'm getting a little lost with the details of this conversation. I think a real-time conversation would be helpful to identify the limitations Scott sees with the codec package.

A couple notes:

1) The initial value I saw with the codec package was being able to abstract away the implementation details of what is written to redis. The storage layer doesn't care exactly how the data is encoded; it just needs to be confident any structs written to redis can be recovered. We're currently going with JSON, but switching to some binary encoding or some kind of compressed encoding will be a 1 file change.

2) Encoding/decoding happens in several different "places" in our code. To focus on two: HTTP requests and saving data to Redis. It's really important the HTTP stuff is consistent with itself, and it's important that the Redis stuff if consistent with itself, but there's no reason the HTTP encoding and Redis encoding need to be the same. As we talk about this codec package, are we talking about how we encode/decode data across out code base? Or just for writing the game state to redis? I think the codec package should be limited to writing/reading redis data.

technicallyty:
> > Right, what you listed is pretty useful for structs where you control the fields and need granular control over how its marshaled/unmarshled. However, the code that was changed here was for our wrapper/generic types, where we don't even know what we're marshaling/unmarshaling.
> 
> Validation would happen with decode not encode, since there's not really anything we can validate on encode (we are turning struct to json)
> 
> On decode, on the example of MessageType, you can use https://github.com/go-playground/validator to validate the struct
> 
> ```
> func (t *MessageType[In, Out]) Decode(bz []byte) (any, error) {
> 	in := new(In)
> 	err := json.Unmarshal(bz, in)
> 	if err != nil {
> 		return *in, eris.Wrap(err, "")
> 	}
> 
>     err = validator.New().Struct(in)
>     if err != nil {
>         log.Fatalf("Validation failed due to %v\n", err)
>     }
> 	return *in, nil
> }
> ```
> 
> On a side note, we can be more strict with the return value here and make it be `(*In, error)` instead of `(any, error)`

just want to add here that the reason we return `any` is to enable `MessageType` to be compliant with `Message interface`. Interface definition doesn't have access to `In`

As far as validation goes -- in order for that validator library to be used, the end user here has to define struct tags on their message's `Input` type, which seems like we should discuss further on what the goals of validation here are. 

Cosmos SDK used to have something similar for their message types, a method called `ValidateBasic` that would do stateless checks on the input of the messages. However, they've since moved away from this and move the responsibility of checking the message input to the keeper functions -- or for our case -- system functions. 

Agree with jeremy here on that i'm unsure where we go from here and think in-person or on-call discussion about the goals of this are to help better understand the direction we wanna go in

smsunarto:
I'm going to close this PR until we have further conversation so i can merge my other stacked PRs

PR #466, opened 2023-11-27
feat(cardinal): update code to enable saving and storing of schemas.

pyrofolium:
Closes: WORLD-596

allows for saving and storing of schemas in world-engine. Created a sibling interface in engine.go to facilitate this. 

Deleted `GetRawJSONOfComponent` because I noticed it wasn't being used anwhere.

### DIFF
--- a/cardinal/ecs/storage/engine.go
+++ b/cardinal/ecs/storage/engine.go
@@ -4,3 +4,13 @@ type NonceStorage interface {
 	GetNonce(key string) (uint64, error)
 	SetNonce(key string, nonce uint64) error
 }
+
+type SchemaStorage interface {
+	GetSchema(componentName string) ([]byte, error)
+	SetSchema(componentName string, schemaData []byte) error
+}
+
+type SchemaAndNonceStorage interface {
+	NonceStorage
+	SchemaStorage
+}
--- a/cardinal/ecs/storage/redis_storage.go
+++ b/cardinal/ecs/storage/redis_storage.go
@@ -47,6 +47,28 @@ func NewRedisStorage(options Options, namespace string) RedisStorage {
 }
 
 // ---------------------------------------------------------------------------
+//
+//	Schema Storage
+//
+// ---------------------------------------------------------------------------
+var _ SchemaStorage = &RedisStorage{}
+
+func (r *RedisStorage) GetSchema(componentName string) ([]byte, error) {
+	ctx := context.Background()
+	schemaBytes, err := r.Client.HGet(ctx, r.schemaStorageKey(), componentName).Bytes()
+	err = eris.Wrap(err, "")
+	if err != nil {
+		return nil, err
+	}
+	return schemaBytes, nil
+}
+
+func (r *RedisStorage) SetSchema(componentName string, schemaData []byte) error {
+	ctx := context.Background()
+	return eris.Wrap(r.Client.HSet(ctx, r.schemaStorageKey(), componentName, schemaData).Err(), "")
+}
+
+// ---------------------------------------------------------------------------
 //							Nonce Storage
 // ---------------------------------------------------------------------------
 
--- a/cardinal/ecs/world.go
+++ b/cardinal/ecs/world.go
@@ -42,6 +42,7 @@ func (n Namespace) String() string {
 
 type World struct {
 	namespace              Namespace
+	schemaStore            storage.SchemaStorage
 	nonceStore             storage.NonceStorage
 	entityStore            store.IManager
 	systems                []System
@@ -307,7 +308,7 @@ func (w *World) ListMessages() ([]message.Message, error) {
 
 // NewWorld creates a new world.
 func NewWorld(
-	nonceStore storage.NonceStorage,
+	schemaAndNonceStore storage.SchemaAndNonceStorage,
 	entityStore store.IManager,
 	namespace Namespace,
 	opts ...Option,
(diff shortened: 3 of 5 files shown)


cardinal/ecs/storage/engine.go
```diff
@@ -4,3 +4,13 @@ type NonceStorage interface {
 	GetNonce(key string) (uint64, error)
 	SetNonce(key string, nonce uint64) error
 }
+
+type SchemaStorage interface {
+	GetSchema(componentName string) ([]byte, error)
+	SetSchema(componentName string, schemaData []byte) error
+}
+
+type SchemaAndNonceStorage interface {
```
smsunarto, line 13: we should call this either `EngineStorage` so it can be a catch all for any subsequent storage we might want to add

cardinal/ecs/storage/redis_storage.go
smsunarto, line 1: At this point, I think it might be a good idea to create a `/storage/redis` package then split this out into 3 different files: storage.go, schema.go, nonce.go

cardinal/ecs/storage/redis_storage.go
```diff
@@ -46,11 +47,32 @@ func NewRedisStorage(options Options, namespace string) RedisStorage {
 	}
 }
 
+// ---------------------------------------------------------------------------
```
smsunarto, line 50: Sorry if it was unclear before, what i mean is that this entire block of "Schema Storage" should just go into `/storage/redis/schema.go`

cardinal/ecs/storage/redis_storage.go
```diff
@@ -46,11 +47,32 @@ func NewRedisStorage(options Options, namespace string) RedisStorage {
...
+	}
+	return schemaBytes, nil
+}
+
+func (r *RedisStorage) SetSchema(componentName string, schemaData []byte) error {
+	ctx := context.Background()
+	return eris.Wrap(r.Client.HSet(ctx, r.schemaStorageKey(), componentName, schemaData).Err(), "")
+}
+
 // ---------------------------------------------------------------------------
```
smsunarto, line 71: subsequently, this would go into /storage/redis/nonce.go

cardinal/ecs/storage/redis_storage.go
smsunarto, line 1: CleanShot 2023-11-27 at 15 54 50@2x

This part of the code should be moved into `/storage/redis/storage.go`

The idea here would be to no longer have a `/storage/redis_storage.go` file it should go to `/storage/redis/storage.go`

cardinal/ecs/storage/redis/storage.go
```diff
@@ -0,0 +1,6 @@
+package redis
+
+type EngineStorage interface {
```
smsunarto, line 3: This could just be the RedisStorage

```
type RedisStorage struct {
	Namespace string
	Client    *redis.Client
	Log       zerolog.Logger
        NonceStorage
        SchemaStorage
}
```

pyrofolium: I don't think this will work. RedisStorage is polymorphic and we store references to RedisStorage as two different types: NonceStorage and SchemaStorage. 

RedisStorage should not contain references to itself which is sort of what's going on here. 

Unless this is what you want. This can work but Nonce and Schema are essentially pointers to itself.

pyrofolium: Then there's three owners of client and this makes it unnecessarily nested as we have to create two new structs just to hold the client and methods and these structs won't be reused anywhere else.

I think a better way is just have `world` hold a single:
```
type EngineStorage interface {
	NonceStorage
	SchemaStorage
}
```

And this interface just serves to hide the properties of the struct. This is more straightforward then polymorphing `RedisStorage` or creating nested structures. We can rename `EngineStorage` if needed too. What do you think?

cardinal/ecs/world.go
```diff
@@ -307,17 +308,26 @@ func (w *World) ListMessages() ([]message.Message, error) {
...
-	nonceStore storage.NonceStorage,
+	schemaAndNonceStore redis.EngineStorage,
 	entityStore store.IManager,
 	namespace Namespace,
 	opts ...Option,
 ) (*World, error) {
 	logger := &ecslog.Logger{
 		&log.Logger,
 	}
+	nonceStore, ok := schemaAndNonceStore.(redis.NonceStorage)
```
jerargus, line 319: These type conversions are unnecessary. The program would fail to compile if schemaAndNonceStore didn't implement these interfaces.

Lines 329 and 330 below can just be:

nonceStore: schemaAndNonceStore
schemaStore: schemaAndNonceStore

cardinal/ecs/storage/redis/schema.go
```diff
@@ -0,0 +1,29 @@
...
+	}
+	return schemaBytes, nil
+}
+
+func (r *Storage) SetSchema(componentName string, schemaData []byte) error {
+	ctx := context.Background()
+	return eris.Wrap(r.Client.HSet(ctx, r.schemaStorageKey(), componentName, schemaData).Err(), "")
+}
+
+var _ NonceStorage = &Storage{}
```
jerargus, line 29: This should be moved near the top of nonce.go

pyrofolium: This change wasn't done because scott requested a complete change in the architecture.

cardinal/ecs/world.go
pyrofolium: This change wasn't done, because scott requested a change in the architecture.

cardinal/ecs/storage/redis/storage.go
pyrofolium: So i'm implementing the initial change Scott decided on. Ultimately I'm ok with this change although I prefer my suggestion over this one. Overall it will work and it's easier to follow for developers as they won't be led astray in code following by the interfaces.

smsunarto (approved):

PR #467, opened 2023-11-28
feat(cardinal): check schema and save schemas on start up

pyrofolium:
Closes: WORLD-661

- Required adjustments to tests in order to create two different worlds that shared the same redis. This allows for testing simulated recovery. 

- During component registration, see: `ecs/world.go` The registration will try to see if the schema was saved to redis under the component name. If the component name does not exists it bypasses this section and continues. 

- The schema check algo: If the name exists it loads the schema and compares it to the one that's trying to get registered. If there's a difference with stored schema and current schema it errors out. Otherwise it will proceed with overwriting the schema. 

- Additionally NewComponent construction now have the potential to return errors because this function now saves the schema to the ComponentMetaData. the schema generation has the potential to return an error and we now bubble it up. 

See the test in: `cardinal/ecs/ecs_test.go` to see at a high level what this feature adds.


cardinal/world.go
```diff
@@ -254,6 +254,9 @@ func (w *World) IsGameRunning() bool {
 }
 
 func (w *World) ShutDown() error {
+	if w.cleanup != nil {
+		w.cleanup()
+	}
 	ok := w.gameSequenceStage.CompareAndSwap(gamestage.StageRunning, gamestage.StageShuttingDown)
```
pyrofolium, line 260: This is a bug. Cleanup can contain instructions to close miniredis. The logic write below this comment actually on occasion can stop the cleanup from ever being called which means that some tests left the port used.

cardinal/ecs/world.go
```diff
@@ -640,6 +666,31 @@ func (w *World) recoverGameState() (recoveredTxs *message.TxQueue, err error) {
...
 }
 
+func (w *World) CheckComponentSchemas() error {
+	var diffErr error
+	for _, componentMetadata := range w.registeredComponents {
+		storedSchema, err := w.redisStorage.Schema.GetSchema(componentMetadata.Name())
+		if err != nil {
+			if eris.Is(eris.Cause(err), redis.Nil) {
+				// No component schemas have been saved, skip this operation return nil error.
+				return nil
```
jerargus, line 676: So if we find that 1 component is not in redis, we can safely assume NO components are in redis? 

What about this sequence:

// Redis starts out empty.
world1 := newWorld()
world1.RegiterComponent(Alpha{})
world1.RegisterComponent(Beta{})
world1.CheckComponentSchemas() // Nothing is in redis; everything is fine.

// Redis has a record of Alpha and Beta
world2 := newWorld()
world2.RegisterComponent(Gamma{})
world2.RegisterComponent(ModifiedAlpha{})
world2.RegisterComponent(Beta{})
world2.CheckComponentSchema() // There's a problem with Alpha, but since Gamma is missing, this function will bail out early, indicating no problem.

pyrofolium: Yeah this is a bug. It should be continue. This function isn't used anyway. I'm deleting it.

pyrofolium: @jerargus The checking logic happens on Registration.

jerargus (approved):

PR #471, opened 2023-11-28
feat(cardinal): validation and error handling for register and auth persona

namrapatel:
Closes: WORLD-613

## Overview

Add some extra validation around persona tags and eth addresses in the `RegisterPersonaSystem` and the `AuthorizePersonaSystem`

## Brief Changelog

- [x] Lowercase eth address
- [x] Trim eth address to check for spaces
- [x] Validate eth address is hex
- [x] Validate persona tag is alphanumeric
- [x] Validate persona tag against lowercase map of persona tags: user can submit case sensitive tag while checks happen in a case insensitive manner

## Testing and Verifying

- Added two tests in `persona_integration_test.go` to test that invalid persona tags and invalid eth addresses are not accepted.

### DIFF
--- a/cardinal/ecs/persona.go
+++ b/cardinal/ecs/persona.go
@@ -2,7 +2,6 @@ package ecs
 
 import (
 	"errors"
-	"fmt"
 	"github.com/ethereum/go-ethereum/common"
 	"regexp"
 	"strconv"
@@ -29,6 +28,8 @@ var CreatePersonaMsg = NewMessageType[CreatePersona, CreatePersonaResult](
 	WithMsgEVMSupport[CreatePersona, CreatePersonaResult],
 )
 
+var regexpObj = regexp.MustCompile("^[a-zA-Z0-9_]+$")
+
 type AuthorizePersonaAddress struct {
 	Address string `json:"address"`
 }
@@ -45,25 +46,26 @@ var AuthorizePersonaAddressMsg = NewMessageType[AuthorizePersonaAddress, Authori
 // users who want to interact with the game via smart contract can link their EVM address to their persona tag, enabling
 // them to mutate their owned state from the context of the EVM.
 func AuthorizePersonaAddressSystem(wCtx WorldContext) error {
-	personaTagToAddress, err := buildPersonaTagMapping(wCtx)
-	if err != nil {
-		return err
-	}
 	AuthorizePersonaAddressMsg.Each(
-		wCtx, func(
-			txData TxData[AuthorizePersonaAddress],
-		) (AuthorizePersonaAddressResult, error) {
+		wCtx, func(txData TxData[AuthorizePersonaAddress]) (result AuthorizePersonaAddressResult, err error) {
 			msg, tx := txData.Msg, txData.Tx
-			result := AuthorizePersonaAddressResult{Success: false}
+			result.Success = false
+
+			personaTagToAddress, err := buildPersonaTagMapping(wCtx)
+			if err != nil {
+				return result, eris.Wrap(err, "")
+			}
 
 			// Check if the Persona Tag exists
-			data, ok := personaTagToAddress[tx.PersonaTag]
+			lowerPersona := strings.ToLower(tx.PersonaTag)
+			data, ok := personaTagToAddress[lowerPersona]
 			if !ok {
 				return result, eris.Errorf("persona %s does not exist", tx.PersonaTag)
 			}
 
 			// Check that the ETH Address is valid
 			msg.Address = strings.ToLower(msg.Address)
+			msg.Address = strings.ReplaceAll(msg.Address, " ", "")
 			valid := common.IsHexAddress(msg.Address)
 			if !valid {
 				return result, eris.Errorf("eth address %s is invalid", msg.Address)
@@ -119,7 +121,8 @@ func buildPersonaTagMapping(wCtx WorldContext) (map[string]personaTagComponentDa
 				errs = append(errs, err)
 				return true
 			}
-			personaTagToAddress[sc.PersonaTag] = personaTagComponentData{
+			lowerPersona := strings.ToLower(sc.PersonaTag)
+			personaTagToAddress[lowerPersona] = personaTagComponentData{
 				SignerAddress: sc.SignerAddress,
 				EntityID:      id,
 			}
@@ -142,23 +145,26 @@ func RegisterPersonaSystem(wCtx WorldContext) error {
 	if len(createTxs) == 0 {
 		return nil
 	}
-	personaTagToAddress, err := buildPersonaTagMapping(wCtx)
-	if err != nil {
-		return err
-	}
 
-	CreatePersonaMsg.Each(wCtx, func(txData TxData[CreatePersona]) (CreatePersonaResult, error) {
+	CreatePersonaMsg.Each(wCtx, func(txData TxData[CreatePersona]) (result CreatePersonaResult, err error) {
 		msg := txData.Msg
-		result := CreatePersonaResult{Success: false}
+		result.Success = false
 
-		if valid := isAlphanumericWithUnderscore(msg.PersonaTag); !valid {
-			err = fmt.Errorf("persona tag %s is not valid, must be alphanumeric with underscores also allowed", msg.PersonaTag)
+		personaTagToAddress, err := buildPersonaTagMapping(wCtx)
+		if err != nil {
+			return result, eris.Wrap(err, "")
+		}
+
+		if !isAlphanumericWithUnderscore(msg.PersonaTag) {
+			err = eris.Errorf("persona tag %s is not valid, must be alphanumeric with underscores also allowed", msg.PersonaTag)
 			return result, err
 		}
 
-		if _, ok := personaTagToAddress[msg.PersonaTag]; ok {
+		// Temporarily convert tag to lowercase to check against mapping of lowercase tags
+		lowerPersona := strings.ToLower(msg.PersonaTag)
+		if _, ok := personaTagToAddress[lowerPersona]; ok {
 			// This PersonaTag has already been registered. Don't do anything
-			err = fmt.Errorf("persona tag %s has already been registered", msg.PersonaTag)
+			err = eris.Errorf("persona tag %s has already been registered", msg.PersonaTag)
 			return result, err
 		}
 		id, err := create(wCtx, SignerComponent{})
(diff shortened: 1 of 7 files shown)


cardinal/ecs/persona.go
```diff
@@ -51,11 +55,20 @@ func AuthorizePersonaAddressSystem(wCtx WorldContext) error {
...
 				return result, eris.Errorf("persona %s does not exist", tx.PersonaTag)
 			}
 
+			// Check that the ETH Address is valid
+			msg.Address = strings.ToLower(msg.Address)
+			valid := common.IsHexAddress(msg.Address)
+			if !valid {
+				return result, eris.Errorf("eth address %s is invalid", msg.Address)
+			}
+
```
technicallyty, line 71: can we also remove all spaces?

```go
msg.Address = strings.ReplaceAll(msg.Address, " ", "")
```

cardinal/ecs/persona.go
```diff
@@ -133,40 +146,55 @@ func RegisterPersonaSystem(wCtx WorldContext) error {
...
 	}
-	for _, txData := range createTxs {
-		tx := txData.Msg
-		if _, ok := personaTagToAddress[tx.PersonaTag]; ok {
+
+	CreatePersonaMsg.Each(wCtx, func(txData TxData[CreatePersona]) (CreatePersonaResult, error) {
+		msg := txData.Msg
+		result := CreatePersonaResult{Success: false}
+
+		if valid := isAlphanumericWithUnderscore(msg.PersonaTag); !valid {
```
technicallyty, line 154:
```suggestion
		if !isAlphanumericWithUnderscore(msg.PersonaTag) {
```

cardinal/ecs/persona.go
```diff
@@ -133,40 +146,55 @@ func RegisterPersonaSystem(wCtx WorldContext) error {
...
+			err = fmt.Errorf("persona tag %s is not valid, must be alphanumeric with underscores also allowed", msg.PersonaTag)
+			return result, err
+		}
+
+		if _, ok := personaTagToAddress[msg.PersonaTag]; ok {
 			// This PersonaTag has already been registered. Don't do anything
-			continue
+			err = fmt.Errorf("persona tag %s has already been registered", msg.PersonaTag)
+			return result, err
```
technicallyty, line 162: theres a mix of eris and fmt. i think we just wanna use eris everywhere.

cardinal/ecs/persona.go
```diff
@@ -133,40 +146,55 @@ func RegisterPersonaSystem(wCtx WorldContext) error {
...
+			SignerAddress: msg.SignerAddress,
 			EntityID:      id,
 		}
-		CreatePersonaMsg.SetResult(
-			wCtx, txData.Hash, CreatePersonaResult{
-				Success: true,
-			},
-		)
-	}
+		result = CreatePersonaResult{Success: true}
```
technicallyty, line 180:
```suggestion
		result.Success = true
```

cardinal/ecs/persona.go
```diff
@@ -133,40 +146,55 @@ func RegisterPersonaSystem(wCtx WorldContext) error {
...
+	// Regular expression pattern for alphanumeric with underscore
+	pattern := "^[a-zA-Z0-9_]+$"
+
+	// Compile the regular expression
+	regexpObj := regexp.MustCompile(pattern)
+
+	// Use the MatchString method to check if the string matches the pattern
+	return regexpObj.MatchString(s)
+}
+
```
technicallyty, line 197: nit: each time this function is called, it'll compile that regexp again. might be better to make the regexpObj a global var.  this function will then only call the last line, (L195)

cardinal/ecs/persona_integration_test.go
```diff
@@ -0,0 +1,267 @@
...
+		},
+	)
+	// This CreatePersona has the same persona tag, but it shouldn't be registered because
+	// it comes second.
+	ecs.CreatePersonaMsg.AddToQueue(
+		world, ecs.CreatePersona{
+			PersonaTag:    wantTag,
+			SignerAddress: "some_other_address",
+		},
+	)
```
technicallyty, line 157: dont think we need this twice. we're only checking that a single message failed.

also, wantTag is a bit confusing. id suggest to remove it completely, since we won't be using it again. same with wantAddress.

cardinal/ecs/persona_integration_test.go
```diff
@@ -0,0 +1,267 @@
...
+	q, err := wCtx.NewSearch(ecs.Exact(ecs.SignerComponent{}))
+	assert.NilError(t, err)
+	err = q.Each(
+		wCtx, func(id entity.ID) bool {
+			count++
+			sc, err := ecs.GetComponent[ecs.SignerComponent](wCtx, id)
+			assert.NilError(t, err)
+			assert.NotEqual(t, sc.PersonaTag, wantTag)
+			assert.NotEqual(t, sc.SignerAddress, wantAddress)
+			return true
```
technicallyty, line 173: all we're checking is that no signer components exist. so this code should just do the count++ to ensure it didnt increment.

cardinal/ecs/persona_integration_test.go
```diff
@@ -0,0 +1,267 @@
...
+			PersonaTag:    wantTag,
+			SignerAddress: wantSigner,
+		},
+	)
+
+	wantAddr := "INVALID ADDRESS"
+	ecs.AuthorizePersonaAddressMsg.AddToQueue(
+		world, ecs.AuthorizePersonaAddress{
+			Address: wantAddr,
+		}, &sign.Transaction{PersonaTag: wantTag},
```
technicallyty, line 243: wantAddr is a little confusing. we don't "want" this addr to show up. i think `invalidAddr` is better, or just inline it since we wont use this again.

cardinal/ecs/persona_integration_test.go
```diff
@@ -0,0 +1,267 @@
...
+			count++
+			sc, err := ecs.GetComponent[ecs.SignerComponent](wCtx, id)
+			assert.NilError(t, err)
+			assert.Equal(t, sc.PersonaTag, wantTag)
+			assert.Equal(t, sc.SignerAddress, wantSigner)
+			assert.Len(t, sc.AuthorizedAddresses, 0) // Assert that no authorized address was added
+			return true
+		},
+	)
+	assert.NilError(t, err)
```
technicallyty, line 263: im seeing this pattern a lot, and i think it may be useful to make a new test helper: `getSigners() []ecs.SignerComponent { ... }`

that way the `count` variable is just `len` of the slice.

cardinal/ecs/persona_test.go
```diff
@@ -1,180 +1,29 @@
...
-				PersonaTag:    personaTag,
-				SignerAddress: fmt.Sprintf("address-%d", i),
-			},
-		)
+	for _, test := range tests {
+		t.Run(test.input, func(t *testing.T) {
+			result := isAlphanumericWithUnderscore(test.input)
+			if result != test.expected {
+				t.Errorf("Expected %v, but got %v", test.expected, result)
+			}
```
technicallyty, line 26: `assert.Equal(t, result, test.expected)`?

cardinal/ecs/persona.go
smsunarto: can prob just create a persona normalization and validator function in /types/persona

cardinal/ecs/persona.go
```diff
@@ -51,11 +55,20 @@ func AuthorizePersonaAddressSystem(wCtx WorldContext) error {
 		) (AuthorizePersonaAddressResult, error) {
 			msg, tx := txData.Msg, txData.Tx
 			result := AuthorizePersonaAddressResult{Success: false}
+
+			// Check if the Persona Tag exists
 			data, ok := personaTagToAddress[tx.PersonaTag]
```
smsunarto, line 60: if we want to allow both lowercase/uppercase characters for persona tag, we need to make this check case insensitive. the problem with this right now is i create a persona tag named `coolmage` and `CoolMage`, both of them will succeed creating ambiguity.

namrapatel: good idea @technicallyty, I knew we needed to do this fix but impl was looking thic

cardinal/ecs/persona.go
```diff
@@ -129,44 +145,56 @@ func RegisterPersonaSystem(wCtx WorldContext) error {
 	if len(createTxs) == 0 {
 		return nil
 	}
```
technicallyty, line 147: dont need this anymore

cardinal/ecs/persona_integration_test.go
technicallyty, line 1: did we add a test for a persona tag not being able to be claimed when using different casings?

i.e. 

register a persona tag "foobar"

try to claim "fooBar"

should fail

cardinal/ecs/persona.go
```diff
@@ -106,7 +121,8 @@ func buildPersonaTagMapping(wCtx WorldContext) (map[string]personaTagComponentDa
 				errs = append(errs, err)
 				return true
 			}
-			personaTagToAddress[sc.PersonaTag] = personaTagComponentData{
+			lowerPersona := strings.ToLower(sc.PersonaTag)
+			personaTagToAddress[lowerPersona] = personaTagComponentData{
 				SignerAddress: sc.SignerAddress,
 				EntityID:      id,
 			}
```
technicallyty, line 128: you'll need to update the `buildPersonaTagMapping` function to have its keys be lowered as well

cardinal/ecs/persona_integration_test.go
namrapatel: added

cardinal/ecs/persona.go
namrapatel: done, the mapping only exists during a tick and so when we append to this map in `RegisterPersonaSystem` appending as a lower cased persona will make it so the next tick's check will be performed as expected

cardinal/ecs/persona.go
```diff
@@ -125,48 +141,55 @@ func buildPersonaTagMapping(wCtx WorldContext) (map[string]personaTagComponentDa
...
-	for _, txData := range createTxs {
-		tx := txData.Msg
-		if _, ok := personaTagToAddress[tx.PersonaTag]; ok {
+
+	CreatePersonaMsg.Each(wCtx, func(txData TxData[CreatePersona]) (result CreatePersonaResult, err error) {
+		msg := txData.Msg
+		result.Success = false
+
+		if !isAlphanumericWithUnderscore(msg.PersonaTag) {
+			err = eris.Errorf("persona tag %s is not valid, must be alphanumeric with underscores also allowed", msg.PersonaTag)
```
technicallyty, line 154:
```suggestion
			err = eris.Errorf("persona tag %s is not valid: must only contain alphanumerics and underscores", msg.PersonaTag)
```

technicallyty (approved):
dank, one nit

PR #475, opened 2023-11-29
refactor(cardinal): move messages package into types.

pyrofolium:
Closes: WORLD-607 and WORLD-603

This just moves ecs/messages package to types/messages

additionally moves messages/queue.go into txqueue/queue.go


jerargus (approved):

cardinal/txqueue/queue.go
```diff
@@ -1,8 +1,10 @@
-package message
+package txqueue
```
smsunarto, line 1: nit: after some thought, i think "txpool" is technically more correct here because tx is processed in batched not FIFO

smsunarto (approved):

PR #476, opened 2023-11-30
feat(cardinal): create general eventhub, and counting aggregator. 

pyrofolium:
Closes: WORLD-615

This is a draft. 

Two things are done here. 

First created a version of EventHub that can be used very generally. 

Added a counter aggregator for metrics. 

It's not public yet, but in ecs.world you can do world.Count("attack") and there is a container that will keep track of the amount of times "attack" was called. 

The very general EventHub is hooked up to this aggregator back end but it was designed in a way where it can be hooked up to any backend. 

Additionally I decomposed the eventhub interface such that the interface can be rebuilt into different versions of eventHub.


smsunarto:
@pyrofolium is this stale?

pyrofolium:
> @pyrofolium is this stale?

It's some prep work I did before break to get ready for adding the elk stack or any analytics event bus we have in mind. If we are still going to do that then it's not. I can get this in once the Jan refactor is done. If that endeavor is cancelled then this would be stale. Im not sure but is data dog replacing this?

pyrofolium:
This will have to wait to go up until after both the server refactor is finished and merging the disabled events back into the server refactor. So this is blocked until then.

smsunarto:
closing this as stale.

PR #478, opened 2023-11-30
feat(cardinal): take out multiple shutdown logging

pyrofolium:
Closes: WORLD-618

Create some state to track shutdown or running state. 

Then use that state to determine when to display certain logs. 

I will still allow shutdown procedures to activate on something that's already shutdown. This is for safety. Maybe the state is incorrect because the shutdown didn't succeed the first time. Thus for this PR the "running" state is only used to display logs.


cardinal/server/server.go
```diff
@@ -232,14 +235,45 @@ func (handler *Handler) Serve() error {
 		return eris.Wrap(err, "error getting hostname")
 	}
 	log.Info().Msgf("serving cardinal at %s:%s", hostname, handler.Port)
-	return eris.Wrap(handler.server.ListenAndServe(), "error listening and serving")
+	err = eris.Wrap(handler.server.ListenAndServe(), "error listening and serving")
+	if err != nil {
+		return err
+	}
+	handler.running.Store(true)
```
jerargus, line 242: ListenAndServe only returns non-nil errors, and blocks while everything is healthy. The Store(true) here will never be called.

pyrofolium: I found some indeterministic output from the logging that I only caught after running this several times. I put in some mutexes to serialize shutdown calls. But this part has been changed.

cardinal/server/server.go
```diff
@@ -232,14 +237,45 @@ func (handler *Handler) Serve() error {
...
+	handler.shutdownMutex.Lock()
+	defer handler.shutdownMutex.Unlock()
+	displayLogs := false
+	if !handler.running.Load() {
+		// handler.running tracks whether the server is running.
+		// for safety allow shutdown to happen whenever this method is called
+		// EVEN if running reads that it is not running.
+		// However, only display the log message if expected running state is consistent.
+		// meaning that shutdown is called on a server where running is true.
+		displayLogs = true
```
jerargus, line 265: It looks like you display logs when the server is NOT initially running, which is the opposite of what the comment is saying.

Can Close and Shutdown be combined into a single function so that only one method is responsible for calling store.running.Store(false)? The initial Store(false) on line 60 is fine, but it would be preferable to remove the one on 242 as well.

Looking at the documentation for server.Shutdown (https://pkg.go.dev/net/http#Server.Shutdown) ListenAndServe should return right away with an error.

pyrofolium: Close and shutdown are different. Can't combine as close is used for tests and shutdown is used in other places. If we want to only use shutdown and not close even in tests I would recommend that for another ticket. 

I fixed this bug, for some reason it worked even without the fix. Also removed redundant changing of store state.

pyrofolium: @jerargus I can make another issue for getting rid of close, but I don't think it will make the beta, it's not critical imo.

jerargus (approved):

PR #481, opened 2023-12-01
feat(cardinal): get pretty printing with server and game_manager. 

pyrofolium:
Closes: WORLD-622

This gets pretty printing working with server and game_manager. So if you set it for game_manager it will be set for everything. 

I maintain this level of granularity in case someone finds it good for testing, or in the future we want server and game_manager to own their own logger.

### DIFF
--- a/cardinal/server/game_manager.go
+++ b/cardinal/server/game_manager.go
@@ -6,6 +6,7 @@ import (
 	"syscall"
 
 	"github.com/rotisserie/eris"
+	"github.com/rs/zerolog"
 	"github.com/rs/zerolog/log"
 	"pkg.world.dev/world-engine/cardinal/ecs"
 )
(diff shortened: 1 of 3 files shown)


cardinal/server/game_manager.go
```diff
@@ -19,11 +20,20 @@ func (g *GameManager) IsRunning() bool {
 	return g.handler.running.Load() && g.world.IsGameLoopRunning()
 }
 
-func NewGameManager(world *ecs.World, handler *Handler) GameManager {
+type GameManagerOptions = func(g *GameManager)
```
technicallyty, line 23:
```suggestion
type GameManagerOptions func(g *GameManager)
```

`=` not needed

technicallyty (approved):

PR #483, opened 2023-12-05
fix(cardinal): Actually shut down cardinal on fatal errors, and do not produce a fat…

jerargus:
…al error when the server is shut down normally

## Overview

log.Fatal.Err(err) does not actually call os.Exit. You have to finish off the line with .Msg(). 

Update world.go to actually shut down cardinal when there is a problem with the server. 

Don't shut cardinal down when the server is cleanly closed.

## Testing and Verifying

- This change is already covered by existing tests.

### DIFF
--- a/cardinal/world.go
+++ b/cardinal/world.go
@@ -3,6 +3,7 @@ package cardinal
 import (
 	"context"
 	"errors"
+	"net/http"
 	"os"
 	"os/signal"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


technicallyty (approved):

pyrofolium:
Apologies late to review. This should pass the stack trace to message. It will make debugging significantly easier. It's fine I'll add it in tmrw.

PR #486, opened 2023-12-05
fix(evm): switch from cosmos logger to zerolog

technicallyty:
## Overview

appication logger seemed to stop working.

### DIFF
--- a/evm/app/app.go
+++ b/evm/app/app.go
@@ -22,6 +22,7 @@ package app
 
 import (
 	"github.com/cometbft/cometbft/abci/types"
+	zerolog "github.com/rs/zerolog/log"
 	signinglib "pkg.berachain.dev/polaris/cosmos/lib/signing"
 	"pkg.berachain.dev/polaris/cosmos/runtime/miner"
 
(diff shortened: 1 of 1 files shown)


namrapatel (approved):

evm/app/app.go
```diff
@@ -219,23 +220,20 @@ func NewApp(
...
 		for _, tx := range txs {
-			app.Logger().Debug("sequencing tx", "tx", tx.String())
+			zerolog.Debug().Msgf("sequencing tx %s", tx.String())
 			_, err := handler(ctx, tx)
 			if err != nil {
-				app.Logger().Error("error handling game shard tx",
-					"error", err.Error(),
-					"tx", tx.String(),
-				)
+				zerolog.Error().Err(err).Msgf("error sequencing game shard tx")
```
pyrofolium, line 232: add stack trace. Eris. `eris.ToString(err)` If the underlying error isn't eris should make it so.

pyrofolium (changes requested):

evm/app/app.go
technicallyty: the error would come from here (https://github.com/Argus-Labs/world-engine/tree/main/evm/x/shard/keeper) which doesn't have eris (yet), so i think its okay as is

pyrofolium (approved):

PR #493, opened 2023-12-06
feat(cardinal): add timing logs for all systems and flush.

pyrofolium:
Closes: WORLD-646

Add timers on all systems and flush in milliseconds.


namrapatel (approved):
lgtm

PR #494, opened 2023-12-06
fix(relay): fill in missing errors and switch error -> debug for some

technicallyty:
## Overview

switches the log level to debug for ErrNotAllowListed in ClaimPersona RPC.

additionally fixes a bunch of issues where we may have been attempting to return nil errors. this should fix issues where clients were seeing the error message set to "0065nil0065" or something to that effect.

### DIFF
--- a/relay/nakama/main.go
+++ b/relay/nakama/main.go
@@ -261,7 +261,10 @@ func handleClaimPersona(ptv *personaTagVerifier, notifier *receiptNotifier) naka
 		// check if the user is verified. this requires them to input a valid beta key.
 		err = checkVerified(ctx, nk, userID)
 		if err != nil {
-			return logDebugWithMessageAndCode(logger, err, AlreadyExists, "unable to claim persona tag")
+			if eris.Is(eris.Cause(err), ErrNotAllowlisted) {
+				return logDebugWithMessageAndCode(logger, err, AlreadyExists, "unable to claim persona tag")
+			}
+			return logErrorMessageFailedPrecondition(logger, err, "unable to claim persona tag")
 		}
 
 		ptr := &personaTagStorageObj{}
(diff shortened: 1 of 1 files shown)


relay/nakama/main.go
```diff
@@ -494,13 +517,14 @@ func logErrorFailedPrecondition(
...
+func errToNakamaError(err error, code int) error {
+	if err != nil {
+		if DebugEnabled {
+			return runtime.NewError(eris.ToString(err, true), code)
+		}
+		return runtime.NewError(err.Error(), code)
 	}
-	return runtime.NewError(eris.Errorf("error: %v", err).Error(), code)
+	return nil
 }
```
technicallyty, line 528: i belive this was the cause of some odd errors the client was seeing where it would return

"error: 0035nil0035"

namrapatel (approved):

jerargus (approved):

relay/nakama/main.go
```diff
@@ -341,9 +349,10 @@ func handleClaimPersona(ptv *personaTagVerifier, notifier *receiptNotifier) naka
 func handleShowPersona(ctx context.Context, logger runtime.Logger, _ *sql.DB, nk runtime.NakamaModule, _ string,
 ) (string, error) {
 	ptr, err := loadPersonaTagStorageObj(ctx, nk)
-	if eris.Is(eris.Cause(err), ErrPersonaTagStorageObjNotFound) {
-		return logErrorMessageFailedPrecondition(logger, err, "no persona tag found")
-	} else if err != nil {
+	if err != nil {
+		if eris.Is(eris.Cause(err), ErrPersonaTagStorageObjNotFound) {
+			return logErrorMessageFailedPrecondition(logger, err, "no persona tag found")
+		}
```
pyrofolium, line 355: seems redundant. if err == nil, the second statement under it will return false.

PR #498, opened 2023-12-08
refactor(nakama): Better logging for beta key claiming and verification

jerargus:
## Overview

1) When a user tries to claim a beta key that is already used, log the user and beta key
2) When a user tries to claim a beta key, but they've already claimed a beta key, include a non-nil error so the resulting logs have more details.

## Testing and Verifying

Logging changes only


pyrofolium (approved):
lgtm

technicallyty (approved):
LGTM

PR #499, opened 2023-12-08
chore(cardinal): Brian/add higher res logs for redis io flush

pyrofolium:
Doesn't close any ticket. 

Changes the logging a bit to add higher resolution timing measurement to the flush of the tick to redis. FinalizeTick does some extra stuff like json deserialization. I want to take that out of the equation.

### DIFF
--- a/cardinal/ecs/ecb/tick.go
+++ b/cardinal/ecs/ecb/tick.go
@@ -72,7 +72,7 @@ func (m *Manager) FinalizeTick() error {
 	flushStartTime := time.Now()
 	_, err = pipe.Exec(ctx)
 	elapsedTime := time.Since(flushStartTime)
-	m.logger.Logger.Info().Int("flush_time", int(elapsedTime.Milliseconds()))
+	m.logger.Logger.Info().Int("flush_time", int(elapsedTime.Milliseconds())).Msg("redis_flush")
 	return eris.Wrap(err, "")
 }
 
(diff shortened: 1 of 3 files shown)


cardinal/ecs/ecb/tick.go
```diff
@@ -68,7 +69,10 @@ func (m *Manager) FinalizeTick() error {
 	if err = pipe.Incr(context.Background(), redisEndTickKey()).Err(); err != nil {
 		return eris.Wrap(err, "")
 	}
+	flushStartTime := time.Now()
 	_, err = pipe.Exec(ctx)
+	elapsedTime := time.Since(flushStartTime)
+	m.logger.Logger.Info().Int("flush_time", int(elapsedTime.Milliseconds())).Msg("redis_flush")
```
technicallyty, line 75: should we make this `flush_time_ms`? just so its clear for people who are looking at logs and not the code

cardinal/ecs/world.go
```diff
@@ -496,7 +496,7 @@ func (w *World) Tick(_ context.Context) error {
 		w.Logger.Warn().Msg(fmt.Sprintf(", (warning: tick exceeded %dms)", warningThreshold.Milliseconds()))
 	}
 	event := w.Logger.Info().
-		Int("flush_time", int(flushElapsedTime.Milliseconds())).
+		Int("finalize_tick_time", int(finalizeTickElapsedTime.Milliseconds())).
```
technicallyty, line 499: maybe here too

technicallyty (approved):
LGTM

technicallyty:
also, should those be debug rather than INFO?

PR #500, opened 2023-12-08
fix(nakama): Reject a claim persona request when cardinal says the persona tag is …

jerargus:
…"available"

Related to: WORLD-643

## Overview

If nakama has a user's claim persona request in a pending state, but cardinal (somehow) never got the transaction to claim the persona, cardinal will report that the given persona tag is "available".

In the CURRENT state of the code, Nakama only check to see if the call to cardinal is "unknown", meaning cardinal has not yet had a chance to process the claim persona tag transactions. If cardinal responds with "available", Nakama treats that as an error, and does not update the status on the Nakama side. Cardinal will forever report "available" and Nakama will forever leave the persona tag as "pending".

This PR changes the Nakama code to treat "available" as a persona tag rejection. Somehow, cardinal never got the persona tag request. By moving the claim to "rejected" on Nakama's end, this will allow the user to pick a new persona tag. The flow should be similar to when a user picks a persona tag that has already been claimed by someone else.

The bad news is: I don't know why Nakama would have a persona tag in the state of "pending" without cardinal also getting the create persona tag request (and either accepting or rejecting it).

At the very least, this PR will 1) give is more information when someone locked in a "pending" state,  2) unlock any account that is still stuck in a pending state 3) prevent future accounts from getting locked in this pending state.

1 way I KNOW a user can get locked into this state is by trying to claim a persona tag that is an upper/lower cased version of some other claimed persona tag. I don't think this is what's goin on in prod though because the client forces everyone's persona tags to be upper cased. 

## Testing and Verifying

Added a unit test to test the claiming of differently-cased persona tags.

### DIFF
--- a/CHANGELOG.MD
+++ b/CHANGELOG.MD
@@ -47,6 +47,8 @@ Ref: https://keepachangelog.com/en/1.0.0/
 
 ### Bug Fixes
 
+- (nakama) #WORLD-643: Reject a persona tag request if it seems like cardinal didn't actually get the corresponding transaction.
+
 ### Client Breaking
 
 ### API Breaking
--- a/internal/nakama/testsuite_test.go
+++ b/internal/nakama/testsuite_test.go
@@ -7,15 +7,13 @@ import (
 	"io"
 	"math/rand"
 	"net/http"
+	"strconv"
 	"strings"
 	"testing"
 	"time"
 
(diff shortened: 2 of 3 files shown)


technicallyty (approved):
LGTM. could we add a bugfix changelog entry?

PR #501, opened 2023-12-09
feat(relay): save game rpc

technicallyty:
## Overview

adds save and get-save to relay/nakama for user data storage


relay/nakama/save.go
```diff
@@ -0,0 +1,190 @@
...
+			logger,
+			eris.Errorf("persona tag %q is not yet verified: cannot save", persona.PersonaTag),
+		)
+	}
+
+	// check if the user is allowlisted. NOTE: checkVerified will return nil in two cases:
+	// case 1: if the allowlist is disabled (via ENABLE_ALLOWLIST env var).
+	// case 2: the user is actually allowlisted.
+	var verified bool
+	err = checkVerified(ctx, nk, userID)
```
jerargus, line 91: What's the reason for including the "verified" field in the game state? 

Above we make sure they have an 'accepted' persona tag status, which means they've made it past the allow-list gate. Is that enough to allow them to save/restore state?

relay/nakama/save.go
```diff
@@ -0,0 +1,190 @@
...
+		"get-save",
+		handleQueryGameSave,
+	)
+	if err != nil {
+		return eris.Wrap(err, "")
+	}
+	return nil
+}
+
+func handleQueryGameSave(ctx context.Context, logger runtime.Logger, _ *sql.DB, nk runtime.NakamaModule, _ string,
```
jerargus, line 155: nit, "handleGetSaveGame" to match the endpoint string

relay/nakama/save.go
```diff
@@ -0,0 +1,190 @@
...
+	Verified bool   `json:"verified"`
+}
+
+const (
+	gameSaveCollection = "game_saves"
+)
+
+func initSaveFileStorage(_ runtime.Logger, initializer runtime.Initializer) error {
+	err := initializer.RegisterRpc(
+		"save",
```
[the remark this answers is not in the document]
technicallyty: done

relay/nakama/save.go
technicallyty: i think i confused myself lol. basically the games team needs a way to save progress in the onboarding flow.

such that if they login to nakama with google, enter a beta key, the fact that they entered that should be saved, in the case they don't yet claim a persona, and need to switch to another PC.

i updated the flow to represent the above. lmk how it looks

relay/nakama/save.go
technicallyty: done

relay/nakama/save.go
jerargus: 👍

jerargus (approved):

jerargus (approved):

PR #503, opened 2023-12-12
feat(cardinal): add timestamp method to world

technicallyty:
## Overview

adds timestamp method to world. timestamps are recorded at the beginning of a tick, just before the systems are ran.

NOTE: this is just part 1, and there is still work to be done for getting this pushed down to the DA. see: https://linear.app/arguslabs/issue/WORLD-666/include-timestamp-in-base-shard-data-payload

## Testing and Verifying

`cardinal/cardinal_test.go`


cardinal/cardinal_test.go
```diff
@@ -117,6 +119,28 @@ func TestCanQueryInsideSystem(t *testing.T) {
 	assert.Equal(t, gotNumOfEntities, wantNumOfEntities)
 }
 
+func TestCanGetTimestampFromWorldContext(t *testing.T) {
+	var ts uint64
+	world := testutils.NewTestWorld(t)
+	err := cardinal.RegisterSystems(world, func(context cardinal.WorldContext) error {
+		fmt.Println("foobar")
```
jerargus, line 126: Did this debug line get left in?

jerargus (approved):

cardinal/cardinal_test.go
technicallyty: removed

namrapatel (approved):

PR #506, opened 2023-12-12
feat(evm): namespace registration and querying precompile 

technicallyty:
## Overview

adds namespace registration and querying as a precompile to the EVM base shard.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,7 +22,5 @@ jobs:
       #   uses: ScribeMD/docker-cache@0.3.4
       #   with:
       #     key: docker-${{ runner.os }}-${{ hashFiles('evm/Dockerfile') }}
-      - name: Forge Build evm/contracts
-        run: make forge-build
       - name: Build Rollup
         run: make rollup-build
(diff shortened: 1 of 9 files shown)


evm/precompile/router/router_test.go
technicallyty, line 1: removing the ginko tests from here cause they were copied from polaris and i also hate that cursed testing framework

.github/workflows/build.yml
```diff
@@ -22,7 +22,5 @@ jobs:
       #   uses: ScribeMD/docker-cache@0.3.4
       #   with:
       #     key: docker-${{ runner.os }}-${{ hashFiles('evm/Dockerfile') }}
-      - name: Forge Build evm/contracts
-        run: make forge-build
```
technicallyty, line 26: we don't actually need to do this, we have tests that cover if our contracts are out of date.

technicallyty:
gonna close this, we're going to make namespace registration an automatic thing.

namrapatel:
> gonna close this, we're going to make namespace registration an automatic thing.
family-man-man-gopher (1)

PR #507, opened 2023-12-13
feat(cardinal): Add environment variable so stats can be exported, and add timing met…

jerargus:
…rics for ticks

Closes: WORLD-671

## Overview

Add a statsd package that connects to a stats agent and some helper methods to track the timing of tick information.

## Testing and Verifying

Existing unit tests pass.


cardinal/statsd/statsd.go
```diff
@@ -0,0 +1,47 @@
...
+	err := Client().Timing("tick", duration, []string{stage}, 1)
+	if err != nil {
+		log.Logger.Warn().Msgf("failed to emit tick stat: %v", err)
+	}
+}
+
+func Init(address string, tags []string) error {
+	if address == "" {
+		return nil
+	}
```
technicallyty, line 31: this seems worthy of an error

jerargus: My thinking was that stats tracking can be disabled by setting this to an empty string.

technicallyty: i think it might be nice to pull that behavior out - check if the address is non-empty before calling `Init`. if it is empty, we could log a warning as well or something?

jerargus: kk, it's an error to try to init statsd with an empty address, and we'll emit a warning when stats are disabled.

technicallyty (approved):

cardinal/world.go
```diff
@@ -106,6 +106,22 @@ func NewWorld(opts ...WorldOption) (*World, error) {
 		return nil, err
 	}
 
+	var metricTags []string
```
technicallyty, line 109: whats going on with this variable?

technicallyty: oh nvm

PR #510, opened 2023-12-14
feat(cardinal): Create guid component and system

jerargus:
Closes: WORLD-679

## Overview

Add a default system that creates a unique game id as well as a query that returns the game id. This game ID will be changes to something new each time the cardinal DB is reset.

## Testing and Verifying

From the world-engine directory, `make game` will build and run nakama and cardinal. 

From the Nakama console you can see the `/query/game/guid` endpoint. Querying that endpoint with `{}` as the body will give you the guid.

### DIFF
--- a/cardinal/ecs/guid.go
+++ b/cardinal/ecs/guid.go
@@ -1,0 +1,81 @@
+package ecs
+
+import (
+	"errors"
+
+	"github.com/cometbft/cometbft/crypto"
+)
+
+const numOfCharactersInGUID = 24
+
+// GUIDComponent is a unique game identification string. It is generated randomly on tick 0 and
+// will never change.
+type GUIDComponent struct {
+	GUID string `json:"guid"`
+}
+
+func (GUIDComponent) Name() string {
+	return "GameUniqueIDComponent"
+}
+
+var ErrGUIDNotFound = errors.New("guid not found")
+
+func getGUID(wCtx WorldContext) (guid string, err error) {
+	search, err := wCtx.NewSearch(Exact(GUIDComponent{}))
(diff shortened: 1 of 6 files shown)


cardinal/ecs/guid.go
```diff
@@ -0,0 +1,81 @@
...
+		return "", err
+	}
+	guidComp, err := GetComponent[GUIDComponent](wCtx, id)
+	if err != nil {
+		return "", err
+	}
+	return guidComp.GUID, nil
+}
+
+func CreateGUIDSystem(wCtx WorldContext) error {
```
technicallyty, line 46: it would be nice if we could figure out a way later down the line to have these singleton style components not have to be a full on system to get their data injected into the ECS storage. Maybe we can make an initializer type of thing.

technicallyty (approved):

cardinal/ecs/guid.go
```diff
@@ -0,0 +1,81 @@
...
+const numOfCharactersInGUID = 24
+
+// GUIDComponent is a unique game identification string. It is generated randomly on tick 0 and
+// will never change.
+type GUIDComponent struct {
+	GUID string `json:"guid"`
+}
+
+func (GUIDComponent) Name() string {
+	return "GameUniqueIDComponent"
```
namrapatel, line 18: I would be in favour of changing this to "GUIDComponent", but I suppose either way works

namrapatel (approved):

jerargus:
I'm holding off on this PR. The e2e test is failing in a pretty mysterious way, and I'd like to get the logging stuff deployed to prod today.

PR #511, opened 2023-12-14
refactor(cardinal): make adapter instantiation automatic and required for production mode

technicallyty:
## Overview

makes adapter a requirement in production mode, user must supply two new env vars in the config.
when in dev mode, no adapter is used, unless explicitly supplied in the options.

this also refactors the WorldConfig, allowing it to use more than just strings. 

removed the Port from the config as it wasn't actually being used anywhere.


cardinal/config.go
```diff
@@ -1,49 +1,48 @@
...
+	DefaultRedisAddress              = "localhost:6379"
+	DefaultBaseShardSequencerAddress = ""
+	DefaultBaseShardQueryAddress     = ""
 )
 
 type WorldConfig struct {
-	RedisAddress      string
-	RedisPassword     string
-	CardinalNamespace string
-	CardinalPort      string
```
technicallyty, line 22: i removed port cause it wasn't actually being used anywhere.

technicallyty:
holding off on this PR til DF work is wrapped.

cardinal/world.go
```diff
@@ -163,6 +158,35 @@ func setLogLevel(levelStr string) error {
...
 }
 
+func applyProductionOptions(
+	cfg WorldConfig,
+	ecsOptions *[]ecs.Option,
+	serverOptions *[]server.Option,
+) error {
+	log.Logger.Info().Msg("Starting a new Cardinal world in production mode")
+	if cfg.RedisPassword == DefaultRedisPassword {
+		return eris.New("redis password is required in production")
```
jerargus, line 168: It would be helpful to include the environment variable with these error messages. e.g.

"in production, redis password must be set via the REDIS_PASSWORD environment variable"

Same with CardinalNamespace and BaseShard addresses below.

cardinal/config.go
jerargus: in cardinal/server/server.go the CARDINAL_PORT env variable is used to set the port. For consistency, I'd suggesting keeping CARDINAL_PORT here, and passing in the port to that package.

If that's not trivial to do here, you should make a linear ticket to capture that work.

jerargus (approved):

cardinal/world.go
technicallyty: [use env var keys in error message](https://github.com/Argus-Labs/world-engine/pull/511/commits/8e34f3b94295e11fe5cf1dfdc8168db46227ed5b)

cardinal/config.go
technicallyty: ill make a new ticket

technicallyty: https://linear.app/arguslabs/issue/WORLD-689/remove-automatic-port-instantiation-in-cardinalserver

namrapatel (approved):

PR #512, opened 2023-12-15
test(nakama): use mockery for unit tests in nakama multi commits

jerargus:
Closes: WORLD-659

## Overview

Add some sample tests that use mockery to mock the Nakama interfaces. This required bumping the nakama image to the latest 3.19. 3.16 was incompatible with the stretcher/testify module. This also puts nakama on version 1.21 of go (up from 1.20).

## Brief Changelog

- Create a .mockery.yaml config that will generate mocks for ALL Nakama interfaces.
- Change the docker-compose.yml health check to use a build-in nakama command line util, instead of curl.
- Update the Nakama base container from 3.16 to 3.19.

## Testing and Verifying

from world-engine/relay/nakama you can run `go test`


relay/nakama/allowlist_test.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+func (a *AllowListTestSuite) TestUserIDRequired() {
+	t := a.T()
+	// This context does not have a user ID.
+	ctx := context.Background()
+
+	_, err := claimKeyRPC(ctx, noopLogger(t), nil, nil, "")
+	// claimKeyRPC should fail because the userID cannot be found in the context
+	assert.Check(t, err != nil)
```
technicallyty, line 40: looks like the assert package has an `IsError` function that will do this behavior

technicallyty (approved):
LGTM, one small nit

relay/nakama/allowlist_test.go
jerargus: done

jerargus:
It's not clear why the coverage has gone down. I'm megging this anyway.

PR #514, opened 2023-12-18
test(nakama): Add more nakama unit tests using mocks

jerargus:
## Overview

Add more unit tests for the allowlist code using the mock framework.
Rename the checkVerified to isUserVerified to make it more explicit when a user is and is not verified. Now, an error is only returned if there was a problem accessing the user's data.

## Testing and Verifying

Unit tests added. Use `go test`


technicallyty (approved):
sweeeeet, very much needed these tests, thanks!!

PR #519, opened 2024-01-03
refactor(cardinal): Remove the ecs/log Logger struct and use the zerolog Logger instead

jerargus:
Closes: WORLD-660

## Overview

Instead of having a custom Logger struct that wraps the zerolog.Logger, use the zerolog.Logger directly.
Keep the methods that attach cardinal specific information to log events.

## Testing and Verifying

Existing unit test verify there are no functional changes


cardinal/ecs/log/log.go
namrapatel, line 1: Curious if we think this code is still necessary? Are these logs used somewhere? Not sure if they're so helpful for the user

namrapatel (approved):

cardinal/ecs/log/log.go
jerargus: I could go either way, but I want to leave that for another PR. Removing the ecslog struct is a self contained piece of work.

PR #520, opened 2024-01-03
fix(cardinal): Fix memory leak in ECB and remove CommitPending

jerargus:
Closes: WORLD-653

## Overview

There were two ways to commit pending ECB changes to redis. FinalizeTick and CommitPending. CommitPending would call "DiscardPending" after saving data to redis, meaning subsequent calls to CommitPending would only include changes that were made since the last call to CommitPending.

FinalizeTick, on the other hand, didn't call DiscardPending. Every time FinalizeTick was called, ALL previous changes would be re-submitted to redis. This includes data that wasn't actually modified.

This PR 1) Calls DiscardPending at the end of FinalizeTick and 2) Removes CommitPending completely (so there is only 1 way to save data to redis).

## Brief Changelog

- Remove CommitPending
- Update tests to use FinalizeTick instead of CommitPending
- Add a test to ensure FinalizeTick execution times and memory allocations do not grow abnormally.
 
## Testing and Verifying

A unit test was added.

### DIFF
--- a/CHANGELOG.MD
+++ b/CHANGELOG.MD
@@ -60,6 +60,8 @@ Ref: https://keepachangelog.com/en/1.0.0/
 
 ### Bug Fixes
 
+- (cardinal) #WORLD-653: Fix memory leak in ECB.
+
 - (nakama) #WORLD-643: Reject a persona tag request if it seems like cardinal didn't actually get the corresponding transaction.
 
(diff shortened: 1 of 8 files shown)


technicallyty (approved):
LGTM

namrapatel (approved):

PR #528, opened 2024-01-06
feat(cardinal): Change to delayed search results.

pyrofolium:
Closes: WORLD-699

Not a big change. 

Just a side note below (not required for reviewing the ticket):

_Typically map, foreach, filter and reduce are all implemented for iterators or collections. They are also usually implemented generically. Meaning usually you just have NewSearch return an iterator and your fine as foreach, map and reduce are already created generically in some generic library. 

We have multiple iterator types in our engine, none of them can utilize the generic implementation and we implement Each multiple times. 

foreach (aka Each) is a little too similar to a for loop, it usually exists as a "cheating" method for functional chains of map reduce processing. So someone can "cheat" and place it on the tail end of a map reduce chain: `results.filter(func(x) {x < 5}).map.filter.foreach(func(x) {iocall(x)})` It's used in place of map because it doesn't return a value so the only way to do something with it is to use side effects or free variables. That's the main purpose of forEach. We aren't using it this way because we don't have the other methods like Filter, reduce or map. 

If forEach exists by itself it's redundant to a for loop so it's a bit pointless. Go has dedicated keywords "for" to handle looping and side effects so why make a method that does the exact same thing?  

Additionally it's also impossible to implement map, filter, and reduce as chained methods in golang (to my knowledge) and doing so would also be sort of against the C-programming style philosophy that golang seems to follow. So what we end up with is sort of a broken feature set that's not very go-like. 

I recommend we should back pedal on Each. just have the user use it in a for loop. But that's another conversation. We can leave it as is, but this section of code is not "elegant" anyway so focusing too much on the design here is pointless imo._


cardinal/ecs/search.go
```diff
@@ -34,27 +35,42 @@ func NewSearch(filter filter.ComponentFilter) *Search {
...
 
+func (q *Search) Commit() error {
+	var err error
+	for _, callback := range q.callbacks {
+		err = callback()
+		if err != nil {
+			break
+		}
+	}
+	q.callbacks = q.callbacks[:0]
```
pyrofolium, line 46: There's alternative behavior here to make the value "consumed". Once committed it will always return an error. lmk if you prefer that, I don't really care.

jerargus:
The goal of this change was to allow the following syntax:

```err := worldCtx.NewSearch(cardinal.Exact(Foo{})).Each(worldCtx, func(id cardinal.EntityID) bool {
        // the search stuff
    }).Commit()
```
Instead of having to check the error from NewSearch. I wasn't expecting this PR to contain the chaining of Each calls together.

You put it well in this PR comment; Trying to chain more general purpose map/reduce functionally shouldn't be a goal of cardinal's search capabilities.

pyrofolium:
> You put it well in this PR comment; Trying to chain more general purpose map/reduce functionally shouldn't be a goal of cardinal's search capabilities.

In my PR comment I'm saying that this is the ideal design goal as that's the typical use case for Each on iterators.

I'm saying this syntax:

```
     for _, entityid := range worldCtx.NewSearch(cardinal.Exact(Foo{})) {
         ....
     }
```

Is more inline with go's philosophy then this:

```
    worldCtx.NewSearch(cardinal.Exact(Foo{})).Each(func(id cardinal.EntityID) bool)
```
Map, reduce, filter, and foreach (aka each) are all methods typically implemented on a generic Iterator interface. The existing thing that we have is both incomplete as it's missing all the other methods, it's also nonstandard and more towards a functional programming style then go-style code. 

I was just mentioning this, but it's not a big deal that it's not this way, but the ideally would be to travel in this direction. We have several implementations of iterators in our code. 

Practical implimentations are not always inline with like the "ideal" but I see that this team places a lot of emphasis on good design, so I thought I'd mention it.  

> The goal of this change was to allow the following syntax:
> 
> ```
> err := worldCtx.NewSearch(cardinal.Exact(Foo{})).Each(worldCtx, func(id cardinal.EntityID) bool {
>         // the search stuff
>     }).Commit()
> ```
> 
> Instead of having to check the error from NewSearch. I wasn't expecting this PR to contain the chaining of Each calls together.
> 
 
When I came up with this alternative interface in the meeting, this was not what I was saying. I Never Said move the Each method onto filters. I said Change the return Value of the each method to delay the result. That's all I said. As far as I know, Moving the Each method onto the filters was explicitly never discussed. 

Anyway I disagree with this interface you're proposing here. This is how the process currently works: 

1. Filters are supposed to compile down into a query. 
2. The query, when run is supposed to return results in the form of entity ids. 
3. The Each method is suppose to operate on entity ids. 

These filters are AST language primitives representing a query; inserting callbacks via Each into these filters is highly nonstandard and atypical. Why would user methods on how to operate on entities be part of the query that fetches the entities? Do you see the disconnect here?

Additionally Chaining of each calls is just harmless extra functionality. You're not liking this I assume:

Each(..).Each(...).Commit()

 It's a feature that exists because the Type checking enforces it. It's one of those features that exists not by some gut level feeling design philosophy, but it has to exist by theoretical necessity. Design via calculation if you will. If Each is to return a SearchResult and not an error this has to be done. Think of it as the functional equivalent of running a for loop on a list twice... there's nothing wrong with two for loops why should there be on two Each calls? 

We can disable the chaining though, I don't care for this as I mentioned in my other comment. If each is called more than once it can become "consumed" and on commit it will return an error. 

I think we're both pretty locked into our opinions here @jerargus let's just throw it to a vote.

jerargus:
Per our discussion, you'll try to match this pattern:

```
err := worldCtx.NewSearch(cardinal.Exact(Foo{})).Each(worldCtx, func(id cardinal.EntityID) bool {
        // the search stuff
    })
```
NewSearch will only return a search object and Each will return an Error.

pyrofolium:
I didn't mar

> Per our discussion, you'll try to match this pattern:
> 
> ```
> err := worldCtx.NewSearch(cardinal.Exact(Foo{})).Each(worldCtx, func(id cardinal.EntityID) bool {
>         // the search stuff
>     })
> ```
> 
> NewSearch will only return a search object and Each will return an Error.

We had an external discussion I believe this ticket was closed and I just went with a new PR here.

PR #531, opened 2024-01-09
feat(cardinal): Get rid of LazySearch

pyrofolium:
Closes: WORLD-703

@smsunarto 

I believe this is closer to what you're looking for. It removes LazySearch and Cardinal.Search just does the "delayed" result instead.


cardinal/lazycontainer.go
```diff
@@ -1,4 +1,4 @@
-package ecs
+package cardinal
```
technicallyty, line 1: i think the lazy container type and constructor can both be made private? looks like this is only used internally

smsunarto:
Closing this since it's no longer needed

PR #532, opened 2024-01-09
refactor(cardinal): change comp registration to hold additional data.

pyrofolium:
Closes: WORLD-705

Register an instance of a component during component registration. For various reasons Scott wants a specific implementation of `NewSearch` The change he wants necessitates that CQL should be able to compile into `Filterable` components. So modifications to the parser and all related unit tests are needed. The first part for the parser change is that we have to register components which is required to build the FIlterable object. 

Most of the code below involves a rename. The main change is the addition of a `map[string]component.Component` on the `ecs.world` type (now called `engine`)


smsunarto:
based on the completed refactor PR, this is not needed

smsunarto:
Closing this since it's no longer needed

PR #533, opened 2024-01-10
refactor(cardinal): make ComponentFilter use component.Component

smsunarto:
Closes: WORLD-707

## Overview

This is PR 1 out of 2 to refactor the Search system. This PR is focused on refactoring
`filter.ComponentFilter` to work wih `component.Component` instead of `component.ComponentMetadata`.

By doing this, we will be able to make `filter.ComponentFilter` the main way to do filter throughout
the codebase. 

## Brief Changelog

- Refactor ComponentFilter to take in component.ComponentMetadata
- Minor adjustment to CQL to accomodate this new change
- Minor adjustment to tests to accomodate this new change

## Testing and Verifying

- Existing unit tests have been adjusted to accomodate this change

PR #534, opened 2024-01-10
refactor(cardinal): make NewSearch use filter.ComponentFilter and replace LazySearch

smsunarto:
Closes: WORLD-703

## Overview

This is PR 2 out of 2 and completes the Search refactor. 

Since we have refactored `filter.ComponentFilter` to accept `component.Component`, there
is no longer a need for Filterable (filter.go). Therefore, this PR eliminates it altogether.

Search now directly uses `filter.ComponentFilter`, which allows us to eliminate `ConvertToComponentFilter`
call that was previously returning an error. As a result, we can drop the LazyContainer altogether.

There is now a clean **single** combo of filter/search. 

One note here: a side-effect of using ConvertToComponentFilter call is that it validates whether a component
is registered with the `Engine` before doing archetype matching. This is a feature that we might want to preserve
but it is technically not required because `evaluateSearch()` (ecs/search.go#107) will just treat filter with unregistered
component like any other filter (it will just not find a match, etc). If we want this validation, we can create a ticket for this check
to be added in `evaluateSearch` prior to running `sm.SearchFrom`.

## Brief Changelog

- Remove filterable and adjust `cardinal/filter.go` to use `filter.ComponentFilter`
- Remove LazyContainer/LazyUpdate
- `cardinal.NewSearch` now just uses a vanilla `ecs.NewSearch`
- Replace all uses of Filterable with `filter.ComponentFilter`
- Replace all usage of search such as within `/evm` and `/ecs/persona.go` to use the
  the new search. Eliminating error returns from NewSearch both from internal use and external use. 
- Adjust tests to accomodate the changes. 

## Testing and Verifying

- Unit tests are updated to accommodate the changes


smsunarto:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/Argus-Labs/world-engine/534?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* `main`
  * **PR #533** <a href="https://app.graphite.dev/github/pr/Argus-Labs/world-engine/533?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #534** <a href="https://app.graphite.dev/github/pr/Argus-Labs/world-engine/534?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/Argus-Labs/world-engine/534?utm_source=stack-comment).

pyrofolium (approved):
This is good. You were right, this is definitively better. I want to comment on one thing though. It's not that developers are just moving down the happy path, it's just we're biased because we've been in the weeds too long building it. I definitely did not Imagine `ComponentMetaData` as an aspect of a `Component` simply because the intention of the design was not for it to be like that in the first place. We made it to be to separate things and by luck a `ComponentMetaData` can be thought of as a component so we never thought of using polymorphism here. 

I think it's good to have fresh unbiased eyes on it so we can get past our preconceptions. I definitely appreciate your insight here, you were ultimately right in all your points.

One thing about your code is the polymorphism can make things a bit more confusing. I suggest changing the definition of ComponentMetaData to utilize the Components interface explicitly to make a bit more clear:


`	ComponentMetadata interface { //revive:disable-line:exported
		// SetID sets the ID of this component. It must only be set once
		SetID(TypeID) error
		// ID returns the ID of the component.
		ID() TypeID
		// New returns the marshaled bytes of the default value for the component struct.
		New() ([]byte, error)

		Encode(any) ([]byte, error)
		Decode([]byte) (any, error)
		GetSchema() []byte
		Component
	}`

technicallyty (approved):

smsunarto:
> This is good. You were right, this is definitively better. I want to comment on one thing though. It's not that developers are just moving down the happy path, it's just we're biased because we've been in the weeds too long building it. I definitely did not Imagine `ComponentMetaData` as an aspect of a `Component` simply because the intention of the design was not for it to be like that in the first place. We made it to be to separate things and by luck a `ComponentMetaData` can be thought of as a `Component` so we never thought of using polymorphism here.
> 
> I think it's good to have fresh unbiased eyes on it so we can get past our preconceptions. I definitely appreciate your insight here, you were ultimately right in all your points.
> 
> One thing about your code is the polymorphism can make things a bit more confusing. I suggest changing the definition of ComponentMetaData to utilize the Components interface explicitly to make a bit more clear:
> 
> ```
> ComponentMetadata interface { 
> 		SetID(TypeID) error
> 		ID() TypeID
> 		New() ([]byte, error)
> 
> 		Encode(any) ([]byte, error)
> 		Decode([]byte) (any, error)
> 		GetSchema() []byte
>                  //Remove Name method add Component interface. 
> 		Component
> 	}
> ```

good idea, just did this

smsunarto:
### Merge activity

* **Jan 10, 4:03 PM**: [Graphite](https://app.graphite.dev/github/pr/Argus-Labs/world-engine/534) rebased this pull request as part of a merge.

PR #537, opened 2024-01-10
refactor(cardinal): batch send transactions at end of tick

technicallyty:
## Overview

updates cardinal to send the txmap to the base shard at the end of the tick.

also some auxiliary changes to appease the linter. main one was the engine.Tick function being long, so tried to tighten it up.


cardinal/option_test.go
```diff
@@ -1,29 +1,12 @@
...
-
-func (d *DummyAdapter) QueryTransactions(_ context.Context, _ *types.QueryTransactionsRequest,
-) (*types.QueryTransactionsResponse, error) {
-	return nil, errors.New("this function should never get called")
-}
-
 func TestOptionFunctionSignatures(_ *testing.T) {
 	// This test is designed to keep API compatibility. If a compile error happens here it means a function signature to
 	// public facing functions was changed.
-	WithAdapter(&DummyAdapter{})
```
technicallyty, line 26: since adapters are internal, i don't think we need to ensure it's own fn signatures change. so removing this.

technicallyty: nvm just gonna change it to nil

cardinal/ecs/engine.go
```diff
@@ -467,32 +471,36 @@ func (e *Engine) Tick(ctx context.Context) error {
...
 		}
 	}
-	statsd.EmitTickStat(allSystemsStartTime, "all_systems")
+	statsd.EmitTickStat(startTime, "all_systems")
 	if e.eventHub != nil {
-		flushEventHubStartTime := time.Now()
 		// engine can be optionally loaded with or without an eventHub. If there is one, on every tick it must flush events.
 		e.eventHub.FlushEvents()
-		statsd.EmitTickStat(flushEventHubStartTime, "flush_events")
+		statsd.EmitTickStat(time.Now(), "flush_events")
```
jerargus, line 489: EmitTickStat takes the difference between the given time and time.Now and emits the delta as a metric. This is just going to cause every "flush_events" metric to be 0. Same with "finalize" below.

jerargus (approved):

namrapatel (approved):

cardinal/ecs/engine.go
technicallyty: ah okay, will revert. linter was complaining this function is too long. ill find another way to fix

PR #538, opened 2024-01-11
test(cardinal): in tests, run cardinal servers on different ports

jerargus:
Closes: WORLD-664

## Overview

In unit tests, run cardinal servers on separate ports to fix flaky tests.

## Testing and Verifying

In the testutils package, random ports are assigned instead of the default 4040.


technicallyty (approved):
LGTM

PR #545, opened 2024-01-19
revert(nakama): Remove notifications based on match state; notifications using the notifications system is still available

jerargus:
## Overview

Our first method of sending tx receipts to users used the a singleton match to transmit receipts via the game state. This is not the intended use of nakama matches or game state, so this PR removes this feature.

Transaction receipts are still sent to users via nakama's notification system (see notifications.go). The game state version of tx receipts stuck around because DF was using it. Now that DF is past, this feature can be removed.

## Summary by CodeRabbit

- **Refactor**
	- Simplified initialization process by integrating receipt matching logic into the main initialization function.
	
- **Documentation**
	- Corrected a typo in code comments for improved clarity.


pyrofolium (approved):
lgtm

PR #559, opened 2024-01-25
refactor(cardinal): make events work with fiber

pyrofolium:
Closes: WORLD-693

The creation of testfixture caused several bugs to surface related to the creation of multiple eventhubs.


cardinal/events/events.go
```diff
@@ -223,58 +222,29 @@ Loop:
 	eh.running.Store(false)
 }
 
-type webSocketHandler struct {
-	internalServe func(*websocket.Conn) error
-	path          string
-	parentHandler http.Handler
-	upgrader      websocket.Upgrader
-}
+func CreateWebSocketEventHandler(hub EventHub) func(conn *websocket.Conn) {
```
technicallyty, line 225: lets use `New<Type>` for constructors

cardinal/world.go
```diff
@@ -326,9 +327,10 @@ func (w *World) StartGame() error {
 		return err
 	}
 	if !w.instance.DoesEngineHaveAnEventHub() {
-		w.instance.SetEventHub(events.NewWebSocketEventHub())
+		eventHub := events.NewWebSocketEventHub()
+		w.instance.SetEventHub(eventHub)
```
technicallyty, line 331: would it be possible to make it so engine just always has an event hub, so `World` doesn't have to be concnered with that?

cardinal/server/handler/events.go
```diff
@@ -0,0 +1,11 @@
+package handler
+
+import (
+	"github.com/gofiber/contrib/websocket"
+	"github.com/gofiber/fiber/v2"
+	"pkg.world.dev/world-engine/cardinal/events"
+)
+
+func WebSocketEvents(hub events.EventHub) (func(*fiber.Ctx) error, func(*fiber.Ctx) error) {
```
namrapatel, line 9: This is a simple function, it can be read, and understood after some investigation, but I still think we should rename it to something better. Looks like it returns an upgrader/handler combo?

cardinal/server/server.go
```diff
@@ -144,6 +146,11 @@ func setupRoutes(app *fiber.App, engine *ecs.Engine, cfg Config) {
 		msgs[msg.Group()][msg.Name()] = msg
 	}
 
+	// Route: /events/
+	websocketUpgrader, websocketHandler := handler.WebSocketEvents(eventHub)
+	app.Use("/events", websocketUpgrader)
```
namrapatel, line 151: For the sake of my understanding, is basically saying that we use the upgrader for everything other than GET requests to `/events`?

cardinal/events/events.go
```diff
@@ -295,12 +265,11 @@ func WebSocketEchoHandler(ws *websocket.Conn) error {
 	}
 }
 
-func sendError(w http.ResponseWriter, erisError error) error {
-	w.Header().Set("Content-Type", "application/json")
-	w.WriteHeader(http.StatusInternalServerError)
-	err := json.NewEncoder(w).Encode(eris.ToJSON(erisError, true))
-	if err != nil {
-		return eris.Wrap(err, "")
+func FiberWebSocketUpgrader(c *fiber.Ctx) error {
```
namrapatel, line 268: nit: I prefer `WebSocketUpgrader` to `FiberWebSocketUpgrader`

namrapatel: nit: Also would be good to rename `WebSocketEchoHandler` to `WebSocketHandler` while we're at it?

cardinal/world.go
pyrofolium: No. You have to emit events in a system through world context. So engine must know what an event hub is. 

Then the server layer emits the events to the users through fiber. So Server needs to know about eventhub. 

eventHub has to live as a 3rd entity: Server, Engine, EventHub. It really depends on how you organize it. If Server, owns engine, then it doesn't explicitly need an event hub it just uses engine and vice versa.

As a package anything can own events. You can put it under server or engine or even outside of cardinal.

technicallyty: What im saying is - is it possible for `engine` to _always_ have an event hub, so the world doesn't have to configure one for it?

cardinal/server/handler/events.go
pyrofolium: I would personally name it `NewWebSocketUpgraderAndHandlerFunctions` but I named it like the above to cater to the convention here.

cardinal/server/server.go
pyrofolium: The api is not clear here. This is what they did in the docs. 

Basically there's two parts of a websocket connection. The http connection and then an upgrade to websockets. The way fiber does it is via context.Next(). Within the upgrader it will pass it along to the websocket handler. Why it's Use and Get is because the api designers decided to do it this way. It's an arbitrary decision from them. Likely done to make it fit the shape of existing methods and type checks.

cardinal/events/events.go
pyrofolium: The echo handler is the term for echoing a message. I can get rid of Fiber, I don't care.

cardinal/server/handler/events.go
pyrofolium: This function isn't explicitly required either I can just pass the functions from the library directly. It's done in attempt to mimic the existing patterns.

cardinal/world.go
pyrofolium: sure but this is a new design direction. I would say it shouldn't be related to this PR. 

I am slightly against this though because eventHub is an interface. There can be multiple implementations and thus it's required to be "injected" in. If it "always" has it it will "always" have one implementation.

technicallyty: no, because you still run the options, allowing different event hubs to be passed in. if no option is passed in, default to websockets, exactly as you have above. but instead of making the `World` object responsible for the default, just do it in `Engine`.

```go
	for _, opt := range opts {
		opt(e)
	}
	if e.receiptHistory == nil {
		e.receiptHistory = receipt.NewHistory(e.CurrentTick(), defaultReceiptHistorySize)
	}
	if e.eventHub == nil {
		e.eventHub = events.NewWebSocketEventHub()
	}
```

technicallyty: this will allow you to get rid of the `DoesEngineHaveAnEventHub` method as well

cardinal/events/events.go
namrapatel: Note the names that you are using on like 150 of `cardinal/server/server.go`, just a simple wsUpgrader and a wsHandler, I like those and think these fns should be the same imo

```
websocketUpgrader, websocketHandler := handler.WebSocketEvents(eventHub)
```

cardinal/world.go
namrapatel: I'm also in favour of what Tyler is suggesting, seems to make sense to me, indifferent which PR it happens in but seems like it is small enough to fit in this one if you want to

cardinal/server/handler/events.go
namrapatel: Read the code a bit more, I see what you're doing, seems like there are a couple other options but this is no biggie, I'll resolve the comment

cardinal/events/events.go
pyrofolium: websocketEchoHandler is different. The purpose is to handle a websocket connection by echoing an incoming message back to the sender. It's not used as part of cardinal. It was originally placed there as a test. It's used to test and maintain the semantics of the websocket.Connection api.

cardinal/world.go
pyrofolium: Let's move it to another PR. These function were in there originally.

technicallyty: https://linear.app/arguslabs/issue/WORLD-750/make-engine-responsible-for-defaulting-eventhub

pyrofolium: > I'm also in favour of what Tyler is suggesting, seems to make sense to me, indifferent which PR it happens in but seems like it is small enough to fit in this one if you want to

SetEventHub and DoesEngineHaveAnEventHub and the whole thing was designed to have Dependency injection. These were all originally there and unrelated to this PR. We should move away from piggy backing changes into every PR because that's what leads to stuff like this server refactor side branch where just too much stuff gets into a single PR that it becomes buggy and nobody really does a thorough review of it. 

This PR is a PR into another giant feature branch that 3 people picked up and applied changes to. Then on top of that we have even more changes concurrently going into main as we speak. it's huge enough already.

PR #561, opened 2024-01-25
refactor(cardinal): make Engine responsible for defaulting the EventHub

technicallyty:
## Overview

previously, World would be responsible for setting up the event hub if engine didn't have one. this PR moves the responsibility to engine.


pyrofolium (approved):
lgtm

namrapatel (approved):

PR #562, opened 2024-01-26
refactor(cardinal): stronger assertion of disallowing entity creation before loading game state

technicallyty:
## Overview

Closes: WORLD-752

before this PR, we had a delayed error of disallowing entity creation before loading game state.

this PR makes that assertion stronger, by erroring IMMEDIATELY if the state is not loaded. 

also removes an unused function.


namrapatel (approved):

pyrofolium (approved):
lgtm

PR #564, opened 2024-01-26
fix(cardinal): Make sure the endpoints declared in swagger are actually created

jerargus:
## Overview

Our swagger.yml file describes the endpoints our cardinal server should expose. Some of those endpoints have variables in them (e.g. custom queries and message), but some endpoints should ALWAYS exist.

This newly added test verifies that the non-variable endpoints are actually created, have the correct method attached to them, and fixes some errors in our code base.

The biggest change here is going from "/debug/state" -> "/query/debug/state". If we're OK with this change, I need to also update the relevant docs.

## Testing and Verifying

Newly added unit tests verifies the swagger endpoints actually get created.

### DIFF
--- a/cardinal/ecs/engine.go
+++ b/cardinal/ecs/engine.go
@@ -20,6 +20,7 @@ import (
 	"github.com/rotisserie/eris"
 	"github.com/rs/zerolog"
 	"github.com/rs/zerolog/log"
+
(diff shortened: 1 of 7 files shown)


technicallyty (approved):

pyrofolium (approved):
lgtm

PR #567, opened 2024-01-30
refactor(nakama): structure and clean RPC handlers

namrapatel:
Closes: WORLD-739

## Brief Changelog
1. RPC registration is isolated in `rpc.go`
2. Handlers are isolated in `handlers.go` and simply unmarshal the msg, call a helper function (3), handle result/error, and return marshaled response
3. "Business logic" for all handlers are in respective subfiles/subpackages such as `allowlist.go` and `persona.go`
4. Some returned errors have been changed. Note in `testsuite_test.go` there are several locations where an expected `409` is now expecting `500` (changed from `FailedPrecondition` to `InternalServerError`, this may be undesirable and can be reverted)

## Testing and Verifying

No new tests required.


relay/nakama/allowlist/allowlist.go
```diff
@@ -3,54 +3,33 @@ package allowlist
...
 
 var (
-	allowlistEnabledEnvVar = "ENABLE_ALLOWLIST"
-	allowlistEnabled       = false
-	allowlistKeyCollection = "allowlist_keys_collection"
-	allowedUsers           = "allowed_users"
+	EnabledEnvVar = "ENABLE_ALLOWLIST"
+	Enabled       = false
+	KeyCollection = "allowlist_keys_collection"
+	AllowedUsers  = "allowed_users"
```
jerargus, line 21: Is there any particular reason why these need to be public variables?

relay/nakama/handlers.go
```diff
@@ -0,0 +1,255 @@
...
+	"context"
+	"database/sql"
+	"encoding/json"
+	errors2 "errors"
+	"github.com/heroiclabs/nakama-common/runtime"
+	"github.com/rotisserie/eris"
+	"io"
+	"net/http"
+	"pkg.world.dev/world-engine/relay/nakama/allowlist"
+	"pkg.world.dev/world-engine/relay/nakama/errors"
```
jerargus, line 13: nit: the standard "errors" package should keep its name, and this should be nakamaerrors.

relay/nakama/errors/errors.go
technicallyty, line 1: on second thought, i think we can remove this package. we should just use the codes from `"google.golang.org/grpc/codes"`instead of re-defining them here.

the errors below can just be defined in the places they're used.

relay/nakama/handlers.go
```diff
@@ -0,0 +1,255 @@
...
+			return utils.LogErrorWithMessageAndCode(logger, err, errors.NotFound, "persona tag storage object not found")
+		case errors2.Is(err, persona.ErrPersonaTagEmpty):
+			return utils.LogErrorWithMessageAndCode(
+				logger,
+				err,
+				errors.InvalidArgument,
+				"claim persona tag request must have personaTag field",
+			)
+		}
+		return utils.LogError(logger, err, errors.Internal)
```
technicallyty, line 64: we actually don't want to use `Internal` as it ends up mapping to `http 500` and crashes webGL clients

namrapatel: All right, looks like (4) was right on the ball

relay/nakama/errors/errors.go
jerargus: +1 to this

relay/nakama/handlers.go
```diff
@@ -0,0 +1,255 @@
...
+	return func(
+		ctx context.Context,
+		logger runtime.Logger,
+		db *sql.DB,
+		nk runtime.NakamaModule,
+		payload string,
+	) (string, error) {
+		ptr := &persona.StorageObj{}
+		if err := json.Unmarshal([]byte(payload), ptr); err != nil {
+			return utils.LogErrorWithMessageAndCode(
```
jerargus, line 31: We have a lot of different utils.Log... helpers. I think they should be consolidated into 1 or 2 helpers that are easier to use. This is for another PR though. Here's the tracking ticket: https://linear.app/arguslabs/issue/WORLD-784/review-utilslog-helpers

relay/nakama/persona/persona.go
```diff
@@ -4,29 +4,114 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	errors2 "errors"
```
jerargus, line 7: Remove this rename.

relay/nakama/allowlist/allowlist.go
namrapatel: They are used in `rpc.go` and `rpc_test.go`, and the last two are used in `allowlist.go` as well

relay/nakama/persona/persona.go
```diff
@@ -103,51 +188,32 @@ func CardinalCreatePersona(
...
-		utils.MakeHTTPURL(readPersonaSignerEndpoint, cardinalAddr),
-		bytes.NewReader(buf),
-	)
+	nk runtime.NakamaModule,
+	globalCardinalAddress string,
+) (res *StorageObj, err error) {
+	personaStorageObj, err := LoadPersonaTagStorageObj(ctx, nk)
 	if err != nil {
-		return "", eris.Wrap(err, "")
+		if eris.Is(eris.Cause(err), ErrPersonaTagStorageObjNotFound) {
```
jerargus, line 198: Is this extra if needed? It looks like the handler is checking for this error so it can return NotFound, but it should still find the error even if it's wrapped.

relay/nakama/persona/persona.go
```diff
@@ -103,51 +188,32 @@ func CardinalCreatePersona(
...
-	if err = json.NewDecoder(httpResp.Body).Decode(&resp); err != nil {
-		return "", eris.Wrap(err, "")
-	}
-	if resp.Status == readPersonaSignerStatusUnknown {
-		return "", eris.Wrap(nakamaerrors.ErrPersonaSignerUnknown, "")
-	} else if resp.Status == readPersonaSignerStatusAvailable {
-		return "", eris.Wrap(nakamaerrors.ErrPersonaSignerAvailable, "")
+// setPersonaTagAssignment attempts to associate a given persona tag with the given user ID, and returns
+// true if the attempt was successful or false if it failed. This method is safe for concurrent access.
+func setPersonaTagAssignment(personaTag, userID string, globalPersonaTagAssignment *sync.Map) (ok bool) {
```
jerargus, line 212: s/globalPersonaTagAssignment/personaTagAssignment/

relay/nakama/persona/persona.go
namrapatel: you're right, killed it

jerargus (approved):

jerargus:
I say just force merge this PR. Coverage has gone down, but coverage wasn't great to begin with. We'll improve test coverage as part of this general refactor/cleanup work.

PR #569, opened 2024-01-30
chore(cardinal): make server.Config unexported

technicallyty:
## Overview

quick pr after i noticed config was exported when it didn't need to be


pyrofolium (approved):
lgtm

PR #570, opened 2024-01-30
refactor(cardinal): consolidate Engine into World

smsunarto:
Closes: WORLD-720

## Overview

This PR contains significant rearchitecture of Engine & World with the following goals:

- Completely eliminate `Engine` in favor of upstreaming the functionalities to `World`, therefore eliminating the extra layer of indirection.
- Reduce cognitive complexity of Engine by organizing it into smaller loosely coupled structs
  - Currently, only SystemManager and MessageManager is implemented. 
- Eliminate tight coupling that is causing circular dependencies

## Brief Changelog

- Extract EngineContext into its own separate module, it's now called`engine.Context`. This is done to eliminate the tight coupling. This is aliased as `cardinal.WorldContext` for external consumption.
- `EngineContext.NewSearch` is moved to `cardinal.NewSearch` this eliminates tight coupling/circular dependency also make the API consistent with the other cardinal operations i.e. `cardinal.CreateMany`, etc
- Introduce `EngineState` to replace `isStateLoaded`, etc to make it clear what is the order of stages of the engine.
- Create `systems.Manager` and `messages.Manager` to extract the Engine systems and messages state/business logic into a compartmentalized loosely coupled struct
- Extract persona components, messages, systems, and queries into a proto-plugin system.
- .... etc

## Additional Work
- I would say that this refactor is incomplete and there are additional work that can be done to clean up the codebase, but I figured it's better to just get this in first so that subsequent work is not blocked.

## Testing and Verifying

- Changes was made to the test to support the refactor, but other than that all tests run.


cardinal/ecs/systems/manager.go
```diff
@@ -0,0 +1,175 @@
...
+
+func (m *Manager) IsSystemsRegistered() bool {
+	return len(m.registeredSystems) > 0
+}
+
+func (m *Manager) GetSystemNames() []string {
+	return m.registeredSystems
+}
+
+func (m *Manager) GetCurrentSystem() string {
```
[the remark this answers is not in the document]
smsunarto: It's used in `Engine.Tick`

cardinal/ecs/messages/manager.go
```diff
@@ -0,0 +1,110 @@
...
+
+// RegisterMessages registers the list of message.Messages with the Manager. Returns an error if any of the
+// messages have duplicate names.
+func (m *Manager) RegisterMessages(msgs ...message.Message) error {
+	// We check for duplicate message names within the slice and against the map of registered messages.
+	// This ensures that we are registering all messages in the slice, or none of them.
+	var seenNames []string
+	for _, msg := range msgs {
+		// Check for duplicate message names against within the slice
+		if slices.Contains(seenNames, msg.Name()) {
```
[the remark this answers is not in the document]
smsunarto: no reason in particular, used slice because `msgs` is already a slice. non-intuitively i think it's also faster at small number of messages because you can preallocate it and it's contiguous memory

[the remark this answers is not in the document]
smsunarto: also it's more 1:1 to the system manager to use slice (and the system manager specifically wants a slice because it cares about ordering, which maps are not able to provide due to non deterministic key iter)

cardinal/ecs/messages/manager.go
```diff
@@ -0,0 +1,110 @@
...
+			return eris.Wrapf(ErrDuplicateMessageName, "duplicate tx %q", msg.Name())
+		}
+
+		// Check for duplicate message names against the map of registere messages
+		err, done := m.isNotDuplicate(msg)
```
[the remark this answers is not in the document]
smsunarto: I prefer not to inline because the method name helps with readability

cardinal/ecs/systems/manager.go
```diff
@@ -0,0 +1,175 @@
...
+		m.currentSystem = &systemName
+
+		// Inject the system name into the logger
+		eCtx.SetLogger(eCtx.Logger().With().Str("system", systemName).Logger())
+
+		// Executes the system function that the user registered
+		systemStartTime := time.Now()
+		err := m.systemFn[systemName](eCtx)
+		if err != nil {
+			m.currentSystem = nil
```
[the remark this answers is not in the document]
smsunarto: unfortunately this will break the expected behavior. The current system is currently used to indicate what system is last running when a panic (from within the system) occurs. If we add a defer, then it will always be nil if there's a panic.

namrapatel (commented):
I think the goals for our refactor should be to:
(1) completely remove the cardinal -> ecs indirection that we have
(2) reduce LOC as a result 
(3) introduce a few new ideas such as the Managers that you've introduced that help tie together the codebase after the indirection has been removed

In this current PR the indirection still exists, but it just moved from where it used to be, for example `entity.go` and `search.go` simply reach into the `ecs` package and do what I believe world context or world used to do before. On top of that, LOC is increased rather than decreased, and new ideas such as the Managers, the refactored EngineContext, etc. don't seem to decrease complexity by decreasing lines of code or number of objects/concepts in our codebase.

The way I think we can actually remove the indirection and decrease LOC is by collapsing `ecs` into `cardinal`. This would probably create some circular dependencies and a meaningful amount of code would have to be rewritten because certain files like `message.go` exist in `ecs` and in `cardinal` (the problem, imo) and collapsing them requires merging them manually.

In order to see if this actually works I'd need to spend a day or two minimum trying to see if I can collapse a single concept such as `messages` from `ecs` into `cardinal` and learn whether that can in practice be done for all the other concepts in `ecs`. Regardless, my stance is that if our major cardinal/ecs refactor does not remove indirection and decrease LOC then it is a lateral move and we should **at least attempt** a PR that aims at removing the indirection and decreasing the LOC. 

Hopefully I've not misunderstood anything that's happened in this PR, I tried to give examples that demonstrate the existence of indirection so as to ensure I'm not confused, but let me know if there is something I'm not seeing. I'm also interested in thoughts from @jerargus @pyrofolium @technicallyty on if they agree and if they think something like this is worth trying.

smsunarto:
> (1) completely remove the cardinal -> ecs indirection that we have

it's a work in progress, this will happen.

> On top of that, LOC is increased rather than decreased, and new ideas such as the Managers, the refactored EngineContext, etc. don't seem to decrease complexity by decreasing lines of code or number of objects/concepts in our codebase.

Reasoning about code complexity based on LOC is the wrong way to think about it, if that's the case then dumping everything in a single giant struct would produce the lowest amount of code but produce tight coupling, make state mutation unclear, and cause circular dependencies because Engine needs to be injected everywhere. 

Splitting the Engine into manager creates cleaner separation of concerns and make state dependencies explicit, therefore reducing cognitive complexity.

cardinal/world_persona.go
```diff
@@ -0,0 +1,43 @@
...
+	wCtx := NewReadOnlyWorldContext(w)
+	s := NewSearch(wCtx, filter.Exact(component.SignerComponent{}))
+	err = s.Each(
+		func(id types.EntityID) bool {
+			sc, err := GetComponent[component.SignerComponent](wCtx, id)
+			if err != nil {
+				errs = append(errs, err)
+			}
+			if sc.PersonaTag == personaTag {
+				addr = sc.SignerAddress
```
technicallyty, line 32: potential nil deref here: error is checked but allows logic to continue when `sc` could be nil

cardinal/system/system_manager.go
```diff
@@ -0,0 +1,149 @@
...
+		systemFn:          make(map[string]System),
+		currentSystem:     nil,
+	}
+}
+
+// RegisterSystems registers multiple systems with the system manager.
+// There can only be one system with a given name, which is derived from the function name.
+// If there is a duplicate system name, an error will be returned and none of the systems will be registered.
+func (m *Manager) RegisterSystems(systems ...System) error {
+	return m.registerSystems(&m.registeredSystems, systems...)
```
technicallyty, line 44: we don't need to pass in `m.registeredSystems` in a method owned by `m`

technicallyty:
going to add review comments as i roam around the code to fix e2e. apologies for not doing a proper review. will do a grouped review when i approach this more formally

smsunarto:
@coderabbitai review

cardinal/persona/query/query_persona_signer.go
```diff
@@ -0,0 +1,56 @@
...
+	SignerAddress string `json:"signerAddress"`
+}
+
+func PersonaSignerQuery(wCtx engine.Context, req *PersonaSignerQueryRequest) (*PersonaSignerQueryResponse, error) {
+	var status string
+
+	addr, err := wCtx.GetSignerForPersonaTag(req.PersonaTag, req.Tick)
+	if err != nil {
+		//nolint:gocritic // cant switch case this.
+		if errors.Is(err, ErrPersonaTagHasNoSigner) {
```
technicallyty, line 40: this check is no longer going to work, since `GetSignerForPersonaTag` defines its own `ErrPersonaTagHasNoSigner`. While the error string may be the same, Go will still treat the errors differently, and this comparison will fail.

e2e/testclients/e2etestclient/testsuite_test.go
```diff
@@ -339,13 +339,12 @@ func waitForAcceptedPersonaTag(c *clientutils.NakamaClient) error {
...
-		if status == "accepted" {
-			break
-		} else if status != "pending" {
-			return fmt.Errorf("bad status %q while waiting for persona tag to be accepted", status)
+		if err == nil {
+			if status == "accepted" {
+				break
+			} else if status != "pending" {
+				return fmt.Errorf("bad status %q while waiting for persona tag to be accepted", status)
+			}
```
technicallyty, line 347: made this change since the error was saying the tick hadn't processed the persona tag yet. so the first error encountered was stopping the retry mechanism from looping.

cardinal/persona/query/query_persona_signer.go
```diff
@@ -37,7 +38,7 @@ func PersonaSignerQuery(wCtx engine.Context, req *PersonaSignerQueryRequest) (*P
 	addr, err := wCtx.GetSignerForPersonaTag(req.PersonaTag, req.Tick)
 	if err != nil {
 		//nolint:gocritic // cant switch case this.
-		if errors.Is(err, ErrPersonaTagHasNoSigner) {
+		if strings.Contains(err.Error(), ErrPersonaTagHasNoSigner.Error()) {
```
technicallyty, line 41: made this change just to get e2e to pass. I'd recommend updating the code to not have duplicate `ErrPersonaTagHasNoSigner` errors. We should define one, and import it where needed.

cardinal/world_persona.go
```diff
@@ -0,0 +1,43 @@
...
+
+var (
+	ErrPersonaTagHasNoSigner        = errors.New("persona tag does not have a signer")
+	ErrCreatePersonaTxsNotProcessed = errors.New("create persona txs have not been processed for the given tick")
+)
+
+// GetSignerForPersonaTag returns the signer address that has been registered for the given persona tag after the
+// given tick. If the engine's tick is less than or equal to the given tick, ErrorCreatePersonaTXsNotProcessed is
+// returned. If the given personaTag has no signer address, ErrPersonaTagHasNoSigner is returned.
+func (w *World) GetSignerForPersonaTag(personaTag string, tick uint64) (addr string, err error) {
```
technicallyty, line 18: i feel like this logic should be moved to the persona plugin.

cardinal/cardinal.go
```diff
@@ -0,0 +1,347 @@
...
+	"strconv"
+)
+
+var (
+	ErrEntityMutationOnReadOnly          = errors.New("cannot modify state with read only context")
+	ErrEntityDoesNotExist                = iterators.ErrEntityDoesNotExist
+	ErrEntityMustHaveAtLeastOneComponent = iterators.ErrEntityMustHaveAtLeastOneComponent
+	ErrComponentNotOnEntity              = iterators.ErrComponentNotOnEntity
+	ErrComponentAlreadyOnEntity          = iterators.ErrComponentAlreadyOnEntity
+	ErrComponentNotRegistered            = iterators.ErrMustRegisterComponent
```
technicallyty, line 20: any reason for aliasing the errors? why not just use them directly?

cardinal/system/system_manager.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	allSystemStartTime := time.Now()
+	for _, systemName := range systemsToRun {
+		// Explicit memory aliasing
+		sysName := systemName
+		m.currentSystem = &sysName
+
+		// Inject the system name into the logger
+		wCtx.SetLogger(wCtx.Logger().With().Str("system", systemName).Logger())
```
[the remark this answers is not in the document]
smsunarto: @coderabbitai can you provide an example

cardinal/persona/utils/utils.go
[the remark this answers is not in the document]
smsunarto: CleanShot 2024-02-12 at 12 37 07

it's used multiple times. i'll just throw it into the `persona` package

cardinal/cardinal.go
smsunarto: mostly devex reason, we wanted to make all functions, errors, etc that is relevant to game devs available in `cardinal`.

cardinal/world_persona.go
smsunarto: this is ideally what i wanted to do, but the server code depends on this via WorldContext, so moving it to plugin would cause circular dependency.

cardinal/persona/utils/utils.go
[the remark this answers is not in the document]
technicallyty: eh? that screenshot just shows its referenced in a comment, then its test, and then your actual usage of it.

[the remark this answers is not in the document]
technicallyty: which now that i see that, that test needs to be moved

jerargus:
I need to spend some time going through the guts of this PR for a proper review, but I have a high level comments that I want to voice first. 

In summary, this PR complicates the user-facing package, which will make it more difficult to use. I personally don't agree with this tradeoff, but I recognize this as a style opinion and not a fundamental flaw with the PR. This opinion is shaped by my affection for godoc pages. Godoc pages are often my first stop when encountering a new library or framework.

The 3 main deficiencies I see with the new cardinal package are:

1. The methods on the WorldContext/engine.Context interface has grown by a lot (6 methods to 20). This is the main object used in Systems and Queries, so it should be as lean as possible. If we do go with this refactor, the methods in engine.Context should be split into "user" methods and "internal" methods, with a comment clearly delineating those groups, and proper documentation for all the "user" methods. Related to this, but more minor the WorldContext->engine.Context type alias makes discovering the methods on WorldContext slightly harder.
2. The cardinal.World object exposes a lot more functionality (11 methods to 32). It will be more difficult to group these methods into "user" methods and "internal" methods, since godoc automatically sorts the methods alphabetically. A user of cardinal should never be calling things like ListMessages, InjectLogger, UseNonce, or WaitForNextTick. Unlike the engine.Context interface above, I don't think there's an easy way to group these methods.
3. More minor: We're exposing QueryType and ComponentMetadata, which are also meant to be for internal use only.

cardinal/world_persona.go
technicallyty: we should move plugins to their own go.mod, as plugin authors would do to write their own

ill make a ticket for it

smsunarto:
> The methods on the WorldContext/engine.Context interface has grown by a lot (6 methods to 20). This is the main object used in Systems and Queries, so it should be as lean as possible. If we do go with this refactor, the methods in engine.Context should be split into "user" methods and "internal" methods, with a comment clearly delineating those groups, and proper documentation for all the "user" methods. Related to this, but more minor the WorldContext->engine.Context type alias makes discovering the methods on WorldContext slightly harder.

The correct solution to this is to just create a separate interface for user facing (i.e. `cardinal.WorldContext`) and internal facing (i.e. `engine.Context`), I just use the alias because it's the simplest. I'm already doing too much in this PR and I would rather do this subsequent PR with smaller diffs so people can give more pinpoint feedback.

> The cardinal.World object exposes a lot more functionality (11 methods to 32). It will be more difficult to group these methods into "user" methods and "internal" methods, since godoc automatically sorts the methods alphabetically. A user of cardinal should never be calling things like ListMessages, InjectLogger, UseNonce, or WaitForNextTick. Unlike the engine.Context interface above, I don't think there's an easy way to group these methods.

I agree that this is an issue, but the clean solution here would be to compartmentalized functionalities better using manager structs and dependency injecting those internally. The root cause of this issue is that `ecs.Engine` used to be the omnibus "do everything" struct. This makes it difficult to understand dependencies on the codebase (for ex, maybe you only need to know about query and message, but you are dependency injecting the entire engine via world context) and it also creates a lot of circular dependency.

cardinal/world_persona.go
smsunarto: wouldn't that also cause circular dependencies? 

`cardinal` depends on `arbitrary_plugin`
`arbitrary_plugin` depends on `cardinal`

cardinal/search.go
```diff
@@ -1,47 +1,119 @@
...
+
+// Each iterates over all entities that match the search.
+// If you would like to stop the iteration, return false to the callback. To continue iterating, return true.
+func (s *Search) Each(callback CallbackFn) error {
+	result := s.evaluateSearch()
+	iter := iterators2.NewEntityIterator(0, s.reader, result)
+	for iter.HasNext() {
+		entities, err := iter.Next()
+		if err != nil {
+			return err
```
jerargus, line 51: Each needs an IsReadOnly check, as well as a logAndPanic wrapper.

If Each encounters an error in a System, cardinal needs to panic. If Each encounters an error in a Query, it should be returned to the user.

cardinal/cardinal.go
```diff
@@ -0,0 +1,346 @@
...
+	if wCtx.IsReadOnly() {
+		return nil, ErrEntityMutationOnReadOnly
+	}
+
+	// Get all component metadata for the given components
+	acc := make([]types.ComponentMetadata, 0, len(components))
+	for _, comp := range components {
+		c, err := wCtx.GetComponentByName(comp.Name())
+		if err != nil {
+			return nil, logAndPanic(wCtx, ErrComponentNotRegistered)
```
jerargus, line 155: This refactor will make keeping track of these non-deterministic errors harder. Any modifications to these state-accessing methods need to make sure logAndPanic is used.

jerargus (changes requested):
Not done yet, but I'm leaving some preliminary comments.

cardinal/world.go
```diff
@@ -173,115 +214,89 @@ func NewMockWorld(opts ...WorldOption) (*World, error) {
...
-	return logAndPanic(wCtx, err)
-}
+// Tick performs one game tick. This consists of taking a snapshot of all pending transactions, then calling
+// each System in turn with the snapshot of transactions.
+func (w *World) Tick(ctx context.Context) error {
+	// Check the current world state and handle accordingly
+	switch w.WorldState {
+	case WorldStateInit:
+		// If the engine is not ready, we don't want to tick the engine.
+		// Instead, make we have recovered the state of the world firstreturn eris.New("invalid world state to tick")
```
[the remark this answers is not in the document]
smsunarto: whoops, accidentally commented the return

cardinal/search.go
smsunarto: fyi, the the original Each also doesn't have an `isReadOnly` check. I would prefer this to be a separate PR if you want this to be changed.

cardinal/cardinal.go
```diff
@@ -0,0 +1,346 @@
...
+	world.nameToQuery[q.Name()] = q
+
+	return nil
+}
+
+// Create creates a single entity in the world, and returns the id of the newly created entity.
+// At least 1 component must be provided.
+func Create(wCtx engine.Context, components ...types.Component) (types.EntityID, error) {
+	// Error if the context is read only
+	if wCtx.IsReadOnly() {
```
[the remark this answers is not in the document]
smsunarto: fixed. on a side note, i personally feel the Create -> CreateMany(.., 1, ...) counter-intuitive. CreateMany calling Create in a for loop seems much more intuitive.

cardinal/types/component.go
```diff
@@ -0,0 +1,66 @@
...
+func IsComponentValid(component Component, jsonSchemaBytes []byte) (bool, error) {
+	componentSchema := jsonschema.Reflect(component)
+	componentSchemaBytes, err := componentSchema.MarshalJSON()
+	if err != nil {
+		return false, eris.Wrap(err, "")
+	}
+	return IsSchemaValid(componentSchemaBytes, jsonSchemaBytes)
+}
+
+func IsSchemaValid(jsonSchemaBytes1 []byte, jsonSchemaBytes2 []byte) (bool, error) {
```
technicallyty, line 51: this can be lowercase, it isn't used outside this package.

cardinal/world.go
```diff
@@ -173,115 +214,89 @@ func NewMockWorld(opts ...WorldOption) (*World, error) {
...
+	case WorldStateReady:
+		// If the world have loaded game state, we can set the it WorldStateRunning
+		w.WorldState = WorldStateRunning
+	case WorldStateRecovering:
+		// If the world is recovering, proceed
+	case WorldStateRunning:
+		// If the world is already running, proceed
+	}
```
technicallyty, line 236: nit: cases that do the same thing can be grouped:

```suggestion
	case WorldStateRecovering, WorldStateRunning:
		// If the world is running or recovering, proceed
	}
```

cardinal/world.go
```diff
@@ -173,115 +214,89 @@ func NewMockWorld(opts ...WorldOption) (*World, error) {
...
-	err := ecs.AddComponentTo[T](wCtx.Engine(), id)
-	if wCtx.Engine().IsReadOnly() || err == nil {
-		return err
-	}
-	if eris.Is(err, ErrEntityDoesNotExist) ||
-		eris.Is(err, ErrComponentAlreadyOnEntity) {
+	// Create the engine context to inject into systems
+	wCtx := NewWorldContextForTick(w, txQueue)
+
+	// Run all registered systems.\
```
technicallyty, line 264:
```suggestion
	// Run all registered systems.
```

cardinal/world.go
```diff
@@ -173,115 +214,89 @@ func NewMockWorld(opts ...WorldOption) (*World, error) {
...
+	// This will run the registsred init systems if the current tick is 0
```
technicallyty, line 265:
```suggestion
	// This will run the registered init systems if the current tick is 0
```

cardinal/world.go
```diff
@@ -173,115 +214,89 @@ func NewMockWorld(opts ...WorldOption) (*World, error) {
...
+	if err := w.systemManager.RunSystems(wCtx); err != nil {
 		return err
 	}
 
-	return logAndPanic(wCtx, err)
-}
+	if w.eventHub != nil {
```
technicallyty, line 270: i believe this case is always true now

cardinal/world.go
```diff
@@ -295,27 +310,31 @@ func (w *World) StartGame() error {
...
 	}
 
-	if err := w.instance.LoadGameState(); err != nil {
-		if errors.Is(err, ecs.ErrEntitiesCreatedBeforeLoadingGameState) {
+	if err := w.loadGameState(); err != nil {
+		if errors.Is(err, ErrEntitiesCreatedBeforeLoadGameState) {
 			return eris.Wrap(ErrEntitiesCreatedBeforeStartGame, "")
 		}
 		return err
 	}
```
technicallyty, line 318: actually, i think we can remove ErrEntitiesCreatedBeforeLoadGameState. there are guards on the entity creation functions now

cardinal/world.go
```diff
@@ -342,11 +433,11 @@ func (w *World) Shutdown() error {
...
-	ok := w.gameSequenceStage.CompareAndSwap(gamestage.StageRunning, gamestage.StageShuttingDown)
-	if !ok {
-		// Either the world hasn't been started, or we've already shut down.
-		return nil
-	}
+	// ok := w.gameSequenceStage.CompareAndSwap(gamestage.StageRunning, gamestage.StageShuttingDown)
+	// if !ok {
+	//	// Either the world hasn't been started, or we've already shut down.
+	//	return nil
+	// }
```
technicallyty, line 440: dead code?

cardinal/world.go
```diff
@@ -357,104 +448,325 @@ func (w *World) Shutdown() error {
...
 
-func (w *World) Tick(ctx context.Context) error {
-	return w.instance.Tick(ctx)
+func (w *World) GetMessageByName(name string) (types.Message, bool) {
+	for _, msg := range w.msgManager.GetRegisteredMessages() {
+		if msg.Name() == name {
+			return msg, true
+		}
+	}
+	return nil, false
```
technicallyty, line 665: feels like this logic should be moved to the manager. the messages are already in a map over there, so should be a very smol method

technicallyty (commented):
first pass. let me know if i hit anything unrelated to the PR. hard to tell with some of the diffs

cardinal/world_persona.go
technicallyty: nope - circular dependency happens for packages, not go modules

cardinal/search.go
smsunarto: > Each needs an IsReadOnly check

What do you mean by `isReadOnly` check? Search _is_ read only, so it shouldn't error on when the context is read only?

pyrofolium (approved):
lgtm see comment.

cardinal/server/handler/tx.go
```diff
@@ -78,33 +78,34 @@ func PostTransaction(
 		var signerAddress string
 		// TODO(scott): this should be refactored; I don't see why getting signer address needs to be different here,
 		//  both of them should just derive the signer address using ecrecover from signature
-		if msgType.Name() == ecs.CreatePersonaMsg.Name() {
+		// TODO(scott): don't hardcode this
```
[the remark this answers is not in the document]
smsunarto: nope, dont want to scope creep it

cardinal/system/system_manager.go
```diff
@@ -0,0 +1,149 @@
...
+		currentSystem:     nil,
+	}
+}
+
+// RegisterSystems registers multiple systems with the system manager.
+// There can only be one system with a given name, which is derived from the function name.
+// If there is a duplicate system name, an error will be returned and none of the systems will be registered.
+func (m *Manager) RegisterSystems(systems ...System) error {
+	return m.registerSystems(systems...)
+}
```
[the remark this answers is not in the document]
smsunarto: i dont think there is a clear usecase where someone would want a system to be executed multiple times and i would prefer to err on the side of safety.

if there's a usecase for it, i'm happy to change it.

cardinal/search.go
jerargus: We're using IsReadOnly to tell when Each is being called from a System and when it's being called from a Query. 

In Queries, all search errors (e.g. a transient DB error) will be returned to the user's custom Query code, and they can fail query, or make up data, or do whatever they want. 

In Systems, that same transient DB error should NOT be returned to the user's custom System code. If they make a decision to update state or not based on that error, they will have a non-deterministic system.

In summary: nil errors can always safely be returned to the caller. "Deterministic" errors can always safely be returned to the caller.

For all other errors:
IsReadOnly == true? return it to the caller
IsReadOnly == false? logAndPanic

cardinal/server/handler/tx.go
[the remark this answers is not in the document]
technicallyty: do we have tickets created for the TODOs in this PR?

cardinal/cardinal.go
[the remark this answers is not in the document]
jerargus: Either way works for me, as long as the validation happens once.

cardinal/message/message_manager.go
```diff
@@ -0,0 +1,92 @@
...
+		m.registeredMessages[msg.Name()] = msg
+		m.nextMessageID++
+	}
+
+	return nil
+}
+
+// IsMessagesRegistered returns true if any messages have been registered with the MessageManager.
+func (m *Manager) IsMessagesRegistered() bool {
+	return len(m.registeredMessages) > 0
```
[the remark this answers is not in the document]
smsunarto: i removed this method in favor of just letting world check the length of list of registered messages, did the same for system.

cardinal/server/handler/health.go
```diff
@@ -17,11 +16,12 @@ type GetHealthResponse struct {
...
 //	@Success		200	{object}	GetHealthResponse
 //	@Router			/health [get]
-func GetHealth(engine *ecs.Engine) func(c *fiber.Ctx) error {
+func GetHealth() func(c *fiber.Ctx) error {
 	return func(ctx *fiber.Ctx) error {
 		return ctx.JSON(GetHealthResponse{
-			IsServerRunning:   true,
-			IsGameLoopRunning: engine.IsGameLoopRunning(),
+			IsServerRunning: true,
+			// TODO(scott): dep inject this value
```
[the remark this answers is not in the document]
smsunarto: im going to just let this here for now. if i understand correctly, the server wont be running if world is not started so this is kind of wonky

cardinal/server/handler/tx.go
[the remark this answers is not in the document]
smsunarto: i'll write them up

cardinal/server/server_test.go
```diff
@@ -270,8 +271,8 @@ func (s *ServerTestSuite) readBody(body io.ReadCloser) string {
 	return string(buf)
 }
 
-// createRandomPersona creates a random persona and returns it as a string
-func (s *ServerTestSuite) createRandomPersona() string {
+// cardinal.CreateRandomPersona cardinal.Creates a random persona and returns it as a string
```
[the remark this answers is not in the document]
smsunarto: thx for catching this, goland being annoying

cardinal/server/server_test.go
```diff
@@ -181,28 +182,28 @@ func (s *ServerTestSuite) TestQueryCustomGroup() {
...
+	err = s.world.Tick(context.Background())
 	s.Require().NoError(err)
 	s.nonce++
 }
 
-// creates a persona with the specified tag.
-func (s *ServerTestSuite) createPersona(personaTag string) {
-	createPersonaTx := ecs.CreatePersona{
+// Creates a persona with the specified tag.
+func (s *ServerTestSuite) CreatePersona(personaTag string) {
```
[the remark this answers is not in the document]
smsunarto: nope, i think its an automatic goland thing. nice catch.

cardinal/world.go
smsunarto: you are correct, good catch.

cardinal/world.go
smsunarto: i uncommented this and also just use store and compare and swap because there is no guarantee that the shutdown is only triggered after the game is at the running stage.

cardinal/filter/filter_test.go
```diff
@@ -273,10 +284,18 @@ func TestCanGetArchetypeFromEntity(t *testing.T) {
...
 	queryString += ")"
 
-	sameQuery, err := cql.Parse(queryString, engine.GetComponentByName)
+	getComponentByName := func(name string) (types.Component, error) {
+		comp, err := world.GetComponentByName(name)
+		if err != nil {
+			return nil, err
+		}
+		return comp, nil
+	}
```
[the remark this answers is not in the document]
smsunarto: prefer to pun this to a separate PR if we want it.

cardinal/cardinal_test.go
```diff
@@ -161,23 +160,25 @@ func TestShutdownViaSignal(t *testing.T) {
...
+	// wsBaseURL := "ws://" + addr
+	// conn, _, err := websocket.DefaultDialer.Dial(wsBaseURL+"/events", nil)
+	// assert.NilError(t, err)
+	// var wg sync.WaitGroup
+	// wg.Add(1)
+	// go func() {
+	// 	_, _, err := conn.ReadMessage()
+	// 	assert.Assert(t, websocket.IsCloseError(err, websocket.CloseAbnormalClosure))
+	//	wg.Done()
+	// }()
```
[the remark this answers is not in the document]
smsunarto: works again now, uncommented

cardinal/server/handler/health.go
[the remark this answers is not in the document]
technicallyty: yeah, now that i think of it - im not sure there'll ever be a case where the server would be running in a state with the game loop not running 🤔

cardinal/cardinal.go
smsunarto: refactor-ed error handling to user defer handler to make this more  straight forward

cardinal/world.go
```diff
@@ -3,72 +3,113 @@ package cardinal
...
 	endStartGame      chan bool
-}
 
-type (
-	// EntityID represents a single entity in the World. An EntityID is tied to
-	// one or more components.
-	EntityID = entity.ID
-	TxHash   = message.TxHash
-	Receipt  = receipt.Receipt
+	WorldState    WorldStateType
```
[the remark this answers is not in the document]
smsunarto: done

cardinal/world.go
```diff
@@ -334,6 +353,78 @@ func (w *World) StartGame() error {
 	return err
 }
 
+func (w *World) startGameLoop(ctx context.Context, tickStart <-chan time.Time, tickDone chan<- uint64) {
+	w.Logger.Info().Msg("Game loop started")
+	ecslog.World(w.Logger, w, zerolog.InfoLevel)
+	w.emitResourcesWarnings()
+
+	go func() {
+		ok := w.isGameLoopRunning.CompareAndSwap(false, true)
```
jerargus, line 375: To reduce the stuff in the World object, this atomic.Bool should be replaced with the single var that tracks the game state/stage. This can happen in a separate PR: https://linear.app/arguslabs/issue/WORLD-820/isgamelooprunning-should-be-removed-in-favor-of-the-stage-tracker

cardinal/world.go
```diff
@@ -357,104 +448,325 @@ func (w *World) Shutdown() error {
...
+}
+
+func (w *World) AddEVMTransaction(
+	id types.MessageID,
+	v any,
+	sig *sign.Transaction,
+	evmTxHash string,
+) (
+	tick uint64, txHash types.TxHash,
+) {
```
jerargus, line 597: I like this formatting for long function signatures.

cardinal/types/engine/context.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"github.com/rs/zerolog"
+	"pkg.world.dev/world-engine/cardinal/gamestate"
+	"pkg.world.dev/world-engine/cardinal/receipt"
+	"pkg.world.dev/world-engine/cardinal/txpool"
+	"pkg.world.dev/world-engine/cardinal/types"
+	"pkg.world.dev/world-engine/sign"
+)
+
+type Context interface {
```
jerargus, line 12: A user  will end up looking at this interface when trying to figure out what they are allowed to do in a System and Query. The User appropriate methods should be placed at the top with comments, and the internal stuff should be at the bottom.

cardinal/system/system_manager.go
```diff
@@ -0,0 +1,149 @@
...
+// RegisterInitSystems registers multiple init systems that is only executed once at tick 0 with the system manager.
+// There can only be one system with a given name, which is derived from the function name.
+// If there is a duplicate system name, an error will be returned and none of the systems will be registered.
+func (m *Manager) RegisterInitSystems(systems ...System) error {
+	return m.registerSystems(systems...)
+}
+
+func (m *Manager) registerSystems(systems ...System) error {
+	// Iterate through all the systems and check if they are already registered.
+	// This is done before registering any of the systems to ensure that all are registered or none of them are.
```
jerargus: I don't understand why it's important to guarantee that either all or none of the systems are registered. This guarantee is also made in the message manager. 

Errors here will likely end up as a log.Fatal. Do you have a use case in mind where the user would depend on this all-or-nothing functionality? 

If it's not needed, this method can be simplified to a single loop over `systems`, no extra systemNames slice, and no slices.Contains call. (same goes for RegisterMessages)

jerargus:
Final thoughts: this PR is too big for me to really evaluate it. I can't tell if it's an improvement, or if it will make working on the code base easier in the future. 

There are obvious good piece in here: Breaking out dedicated system/message managers, unifying the WorldState sequence under a single variable, and moving the creation of the read-only world context entirely out of the server come to mind.

But a lot of the PR is opaque to me. The biggest drawback is the new size of the cardinal API. You've outlined a path froward to fixing that in a previous comment, but I can't see that far ahead.

I'm going to approve, with the assumption that you'll address the final few comments I made, but I'd still like to have an in-person conversation about this tomorrow.

jerargus (approved):

cardinal/system/system_manager.go
smsunarto: It's just defensive programming. I prefer that my program is never in an invalid state without making assumptions of external error handling.

cardinal/world.go
smsunarto: will just do as separate pr

PR #572, opened 2024-01-31
refactor(cardinal): router package

technicallyty:
## Overview

This consolidates the shard/evm package into a single, new package `router`. 

The funcitonalities of each of those packages has been combined into a singular object, `router`, which handles submitting txs to the base shard, handling evm -> cardinal txs,  evm -> cardinal queries, and cardinal -> base shard historical transaction queries.

## Brief Changelog

- added `HandleEVMQuery` to engine to support router queries to avoid depending on ECS.
- added `ConsumeEVMMsgResult` to engine with destructured  return values.
- added `GetMessageByName` to engine.
- added `GetPersonaForEVMAddress` to engine.
- added `SetRouter` method to engine to allow engine to own the router object.
- added `RunRouter` method to engine to allow `World` to run the router without owning it.
- added more robust tests for router implementation via mockgen.
- removed optional for setting up router. its automatic in production mode.

## Testing and Verifying

unit tests


cardinal/router/router.go
technicallyty, line 1: 99.99% of this code is copied. would prefer not to nitpick its implementation in this PR 🤓

pyrofolium: nit: please leave comment indicating code is generated.

pyrofolium (approved):
lgtm

cardinal/router/router.go
technicallyty: oh meant copied as in, i wrote it in the evm/shard packages, and copied the impl back into `router`. so not generated, just not changed from my previous impl

cardinal/ecs/engine.go
```diff
@@ -92,6 +93,78 @@ type Engine struct {
...
+func (e *Engine) GetMessageByName(s string) (message.Message, bool) {
+	for _, msg := range e.registeredMessages {
+		if msg.Name() == s {
+			return msg, true
+		}
+	}
+	return nil, false
+}
+
+func (e *Engine) GetPersonaForEVMAddress(addr string) (string, error) {
```
smsunarto, line 132: Can we consolidate it with this? https://github.com/Argus-Labs/world-engine/blob/main/cardinal/ecs/persona.go#L255-L283

Both of them is EVM addresses, it's on the same curve and same signing scheme.

cardinal/router/provider.go
```diff
@@ -0,0 +1,18 @@
+package router
+
+import (
+	"pkg.world.dev/world-engine/cardinal/types/message"
+	"pkg.world.dev/world-engine/sign"
+)
+
+//go:generate mockgen -source=provider.go -package mocks -destination=mocks/provider.go
+type Provider interface {
```
smsunarto, line 9: Would appreciate a comment here on what this interface is used for

namrapatel (approved):

cardinal/router/router.go
```diff
@@ -0,0 +1,238 @@
...
+	shard "pkg.world.dev/world-engine/rift/shard/v2"
+	"pkg.world.dev/world-engine/sign"
+)
+
+const (
+	defaultPort = "9020"
+)
+
+//go:generate mockgen -source=router.go -package mocks -destination=mocks/router.go
+type Router interface {
```
smsunarto, line 24: Add a comment on what Router is responsible for

cardinal/router/doc.go
```diff
@@ -0,0 +1,6 @@
+// Package router provides functionality for Cardinal to interact with the EVM Base Shard.
```
smsunarto, line 1: hm, i would prefer to have this comment direclty on the struct definition or constructor so you get an explanation on it when you hover over the type def.

cardinal/router/router.go
```diff
@@ -0,0 +1,238 @@
...
+	Submit(
```
smsunarto, line 25: a comment on what Submit does

cardinal/router/router.go
```diff
@@ -0,0 +1,238 @@
...
+		ctx context.Context,
+		processedTxs txpool.TxMap,
+		namespace string,
+		epoch,
+		unixTimestamp uint64,
+	) error
+
+	QueryTransactions(ctx context.Context, req *shardtypes.QueryTransactionsRequest) (
```
smsunarto, line 33: a comment on what this does

cardinal/router/router.go
```diff
@@ -0,0 +1,238 @@
...
+	CodeTxFailed
+	CodeNoResult
+	CodeServerUnresponsive
+	CodeUnauthorized
+	CodeUnsupportedMessage
+	CodeInvalidFormat
+)
+
+// SendMessage is the server impl that receives SendMessage requests from the base shard client.
+func (r *routerImpl) SendMessage(_ context.Context, req *routerv1.SendMessageRequest,
```
smsunarto, line 108: nit: call this HandleShardMessage. When I see SendMessage i thought this is the metohd that is responsible in sending messages to the base shard, which I think is not the case?

cardinal/router/router.go
```diff
@@ -0,0 +1,238 @@
...
+	return &routerv1.SendMessageResponse{
+		Errs:      errStr,
+		Result:    result,
+		EvmTxHash: evmTxHash,
+		Code:      code,
+	}, nil
+}
+
+// QueryShard is the server impl that answers query requests from the base shard client.
+func (r *routerImpl) QueryShard(_ context.Context, req *routerv1.QueryShardRequest) (
```
smsunarto, line 185: nit: same as above, would call this HandleRouterQuery

cardinal/router/router.go
```diff
@@ -0,0 +1,238 @@
...
+	req := shard.SubmitTransactionsRequest{
+		Epoch:         epoch,
+		UnixTimestamp: unixTimestamp,
+		Namespace:     namespace,
+		Transactions:  messageIDtoTxs,
+	}
+	_, err := r.ShardSequencer.Submit(ctx, &req)
+	return eris.Wrap(err, "")
+}
+func (r *routerImpl) QueryTransactions(ctx context.Context, req *shardtypes.QueryTransactionsRequest) (
```
smsunarto, line 222: comment

cardinal/router/router.go
```diff
@@ -0,0 +1,238 @@
...
+	*shardtypes.QueryTransactionsResponse,
+	error,
+) {
+	res, err := r.ShardQuerier.Transactions(ctx, req)
+	return res, eris.Wrap(err, "")
+}
+
+func transactionToProto(sp *sign.Transaction) *shard.Transaction {
```
smsunarto, line 230: would move this router/codec

cardinal/router/router.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+	req := shard.SubmitTransactionsRequest{
+		Epoch:         epoch,
+		UnixTimestamp: unixTimestamp,
+		Namespace:     namespace,
+		Transactions:  messageIDtoTxs,
+	}
+	_, err := r.ShardSequencer.Submit(ctx, &req)
+	return eris.Wrap(err, "")
+}
```
smsunarto, line 221: I'm guessing this is for transmitting transaction data to base shard? Should rename this to `SubmitTxBlob`

cardinal/router/router.go
```diff
@@ -0,0 +1,238 @@
...
+	reply, err := r.provider.HandleEVMQuery(req.Resource, req.Request)
+	if err != nil {
+		zerolog.Logger.Error().Err(err).Msg("failed to handle query")
+		return nil, err
+	}
+	zerolog.Logger.Debug().Msgf("sending back reply: %v", reply)
+	return &routerv1.QueryShardResponse{Response: reply}, nil
+}
+
+func (r *routerImpl) Submit(
```
smsunarto, line 198: I'm guessing this is for transmitting transaction data to base shard? Should rename this to `SubmitTxBlob`

cardinal/router/router.go
```diff
@@ -0,0 +1,238 @@
...
+	// we don't need secure comms for this connection, cause we're just querying cosmos public RPC endpoints.
+	conn2, err := grpc.Dial(baseShardQueryAddr, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		return nil, eris.Wrapf(err, "error dialing evm base shard address at %q", baseShardQueryAddr)
+	}
+	rtr.ShardQuerier = shardtypes.NewQueryClient(conn2)
+	return rtr, nil
+}
+
+const (
```
smsunarto, line 97: nit: type/const declaration should always be on top of the file

cardinal/router/router.go
```diff
@@ -0,0 +1,238 @@
...
+	)
+
+	Shutdown()
+	Run() error
+}
+
+var _ routerv1.MsgServer = (*routerImpl)(nil)
+var _ Router = (*routerImpl)(nil)
+
+type routerImpl struct {
```
smsunarto, line 45: would just call this router

cardinal/router/router.go
```diff
@@ -0,0 +1,238 @@
...
+	shutdown func()
+}
+
+func (r *routerImpl) Shutdown() {
+	if r.shutdown != nil {
+		r.shutdown()
+	}
+}
+
+func (r *routerImpl) Run() error {
```
smsunarto, line 62: nit: call this Start. I'm trying to make it consistent across the codebase

cardinal/router/router.go
```diff
@@ -0,0 +1,238 @@
...
+		err = eris.Wrap(server.Serve(listener), "error serving server")
+		if err != nil {
+			zerolog.Fatal().Err(err).Msg(eris.ToString(err, true))
+		}
+	}()
+	r.shutdown = server.GracefulStop
+	return nil
+}
+
+func New(sequencerAddr, baseShardQueryAddr string, provider Provider) (Router, error) {
```
smsunarto, line 79: the constructor function should be the first function right below the type router struct {}. the methods should go after it in the order they are declared in the interface

cardinal/router/router.go
```diff
@@ -0,0 +1,238 @@
...
+	if err != nil {
+		return eris.Wrapf(err, "error listening to port %s", r.port)
+	}
+	go func() {
+		err = eris.Wrap(server.Serve(listener), "error serving server")
+		if err != nil {
+			zerolog.Fatal().Err(err).Msg(eris.ToString(err, true))
+		}
+	}()
+	r.shutdown = server.GracefulStop
```
smsunarto, line 75: i would just inline this in shutdown

cardinal/router/router.go
```diff
@@ -0,0 +1,238 @@
...
+	return nil
+}
+
+func New(sequencerAddr, baseShardQueryAddr string, provider Provider) (Router, error) {
+	rtr := &routerImpl{port: defaultPort, provider: provider}
```
smsunarto, line 80: i dont see a way to change the port, might as well inline it?

cardinal/ecs/engine.go
```diff
@@ -92,6 +93,78 @@ type Engine struct {
 	shutdownMutex sync.Mutex
 }
 
+func (e *Engine) ConsumeEVMMsgResult(evmTxHash string) ([]byte, []error, string, bool) {
+	rcpt, exists := e.consumeEVMMsgResult(evmTxHash)
+	return rcpt.ABIResult, rcpt.Errs, rcpt.EVMTxHash, exists
+}
+
+func (e *Engine) HandleEVMQuery(name string, abiRequest []byte) ([]byte, error) {
```
smsunarto, line 101: i generally dont want to differentiate EVM/non-EVM once it has made its way to engine. the responsibility of managing the EVM specifics should be router's responsibility.

smsunarto (changes requested):
above

cardinal/router/provider.go
smsunarto: oh, i understand the point of this now. would still love the comment, but yeah i think this is a great idea. we should probably do this for server and event hub as well.

cardinal/router/router.go
```diff
@@ -0,0 +1,238 @@
...
+	CodeUnauthorized
+	CodeUnsupportedMessage
+	CodeInvalidFormat
+)
+
+// SendMessage is the server impl that receives SendMessage requests from the base shard client.
+func (r *routerImpl) SendMessage(_ context.Context, req *routerv1.SendMessageRequest,
+) (*routerv1.SendMessageResponse, error) {
+	// first we check if we can extract the transaction associated with the id
+	msgType, exists := r.provider.GetMessageByName(req.MessageId)
```
smsunarto, line 111: i dont really mind either way, but it's probably better to just get a list of messages in New() and create an index/map that stores it in the router struct. Talking through an interface is slow (Go doesn't have zero cost abstraction) and you are copying this object again and again for every message that comes in here.

cardinal/ecs/engine.go
technicallyty: the function you linked is for matching the `Signer` <> `PersonaTag` pair. that is different than what is going on here, where im looking for the persona that has this EVM address under its authorized address list. so not really anything to do with the signing/curve stuff, moreso with what im looking for in the `SignerComponent`. 

However looking here again, we should probably require `Router` callers from EVM smart contracts to specify the PersonaTag they want to send the message as, and then check if they are actually under that tag's `AuthorizedAddress` list. but that will be a much larger refactor, and shouldn't be done in this PR.

cardinal/router/router.go
technicallyty: added

cardinal/router/provider.go
technicallyty: added

cardinal/router/doc.go
technicallyty: removed to typedef

cardinal/router/router.go
technicallyty: done

cardinal/router/router.go
technicallyty: this is the Server implementation of the following protobuf defition. `https://github.com/Argus-Labs/world-engine/blob/main/rift/proto/router/v1/router.proto#L12-L25`

there are no external callers of this method (besides the gRPC runtime), hence why it was not included in the `Router` interface. 

i can rename if you'd like, but that should be another PR as its unrelated to this PR (this code was copied from previous EVM package).

cardinal/router/router.go
technicallyty: same as above.

cardinal/router/router.go
technicallyty: comments will go on the `Router` interface

cardinal/router/router.go
technicallyty: this is a bit too specific to `Router`s code to go in something like codec IMO. also this is just converting an object to another, so not really encoded related here.

cardinal/router/router.go
technicallyty: done

cardinal/router/router.go
technicallyty: actually, i think im going to move the evm grpc server specific code to it's own file. my intent here was to put the const next to the only methods that use it.

cardinal/router/router.go
technicallyty: done

cardinal/router/router.go
technicallyty: done

cardinal/router/router.go
technicallyty: done

cardinal/router/router.go
technicallyty: i think we want to eventually allow for configuration of this port. i'll open a ticket.

cardinal/ecs/engine.go
technicallyty: It's a completely different data format though. Requests come through as ABI encoded bytes. Router has no idea what to do with those bytes unless it has it's query object, which would require importing ECS.

cardinal/router/router.go
technicallyty: I don't think this is copying objects? I either get the message from my own index, or i get it from Engine's. The tradeoff here has to do with whether we want to store the index twice (both engine and router store) or store it once, and get it through the provider interface.

i don't think talking through an interface to get the message is going to be a significant enough performance hit.

cardinal/router/router.go
smsunarto: just rename it then

cardinal/router/router.go
smsunarto: imo, it's cleaner to just do `grpc.NewServer()` and `routerv1.RegisterMsgServer(server, r)` on `New()` to store it in a router.server field. Better separation of concern between creation/start/shutdown.

cardinal/router/router.go
technicallyty: this method isn't really a codec type of thing though. its just object conversion.

cardinal/router/router.go
smsunarto: `r.provider.GetMessageByName()` returns a copy of the message object by value instead of a reference, so the object `msgType` is copied from engine and deallocated once the function terminates. 

Object allocation is expensive in Go and SendMessage is a hot function (it will be called a lot) in contrast to `New()` that is only called once.

cardinal/router/router.go
technicallyty: changed here: [make grpc server a field on router](https://github.com/Argus-Labs/world-engine/pull/572/commits/e9b5cb0c6fa58653aec9ea862f837d8463c60f43)

cardinal/router/router.go
smsunarto: i don't really mind what you call, goal here is to move it out of router.go

cardinal/router/router.go
technicallyty: the message is an interface. interfaces are always reference/pointer types, not values.

i'm not sure how getting the message interface from engine or from a mapping in `router` changes object allocation. an object is allocated for the message type no matter where i get it from

cardinal/ecs/engine.go
smsunarto: I moved out the Query interface out of cardinal/ecs in my PR, can do it later, but the same way you are doing `GetMessageByName` there should be nothing stopping us from just doing that for query too.

cardinal/router/router.go
technicallyty: i don't agree that its clutter though. its one function.

technicallyty: ill just inline it instead.

cardinal/ecs/engine.go
technicallyty: yup, i would have done the same pattern had i not needed to import ECS to deal with queries. ill open a ticket to do that once your PR is in.

cardinal/router/router.go
smsunarto: now that you mention it, wouldn't it be cleaner if there is a separate gRPC server struct that implements the expected method?

```
type routerServer struct {} 

func newServer() routerServer {
    // all the current gRPC setup logic like dial and gRPC.NewServer(), etc goes here
}
```


```
type router struct {
    server RouterServer
    ...
}

func New() {
     return router{
          server: newServer()
     }
}
```

all the

technicallyty: [make evm server its own object](https://github.com/Argus-Labs/world-engine/pull/572/commits/5abdb5b87a265c5110f7851fcbd41cad56110aed) done

smsunarto (approved):

PR #581, opened 2024-02-01
feat(nakama): Create Signer interface and include option to sign transactions using KMS

jerargus:
Closes: WORLD-1

## Overview

Add the ability to sign transactions with Google's KMS. To enable, the GCP_KMS_CREDENTIALS_FILE and  GCP_KMS_KEY_NAME and environment variables must be set.

The KMS signature doesn't include ethereum's recovery ID, so the recovery ID needs to be derived after the KMS call.

## Brief Changelog

- Create a Signer interface that will sign transactions. 
- Hide the current pattern of storing the private key in the Nakama DB behind the Signer interface
- Add a KMS signer that can sign transactions using GCP's KMS service
- Move nonce management into dedicated struct (both the KMS signer and the Nakama signer use Nakama to store the nonce)
- Add environment variables to enable KMS signing

## Testing and Verifying

Unit tests were added. The input data for the unit tests were generated using actual calls to the KMS service.


jerargus:
I still plan on adding some unit test to check failure cases, but the main logic of this PR is ready for review.

pyrofolium (approved):

relay/nakama/signer/kms.go
```diff
@@ -0,0 +1,241 @@
...
+
+	if !result.VerifiedDigestCrc32C {
+		return nil, fmt.Errorf("AsymmetricSign: request corrupted in-transit")
+	}
+	if result.Name != req.Name {
+		return nil, fmt.Errorf("AsymmetricSign: request corrupted in-transit")
+	}
+	if int64(crc32c(result.Signature)) != result.SignatureCrc32C.Value {
+		return nil, fmt.Errorf("AsymmetricSign: response corrupted in-transit")
+	}
```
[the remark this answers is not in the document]
jerargus: Nope, this was based on some google sample code. Updated.

relay/nakama/signer/kms.go
```diff
@@ -0,0 +1,241 @@
...
+	//	publicKey, err := x509.ParsePKIXPublicKey(block.Bytes)
+	//	if err != nil {
+	//		return "", eris.Wrap(err, "failed to parse public key")
+	//	}
+	//	pubKey, ok := publicKey.(*ecdsa.PublicKey)
+	//	if !ok {
+	// 		return "", eris.Wrap(err, "public key is not elliptic curve")
+	//	}
+	//
+	// But the above code does not seem to support BLA FINISH THIS COMMENT
```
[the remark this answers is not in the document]
jerargus: fixed.

relay/nakama/signer/signer.go
```diff
@@ -0,0 +1,20 @@
...
+import (
+	"context"
+
+	"pkg.world.dev/world-engine/sign"
+)
+
+type Signer interface {
+	SignTx(ctx context.Context, personaTag string, namespace string, data any) (*sign.Transaction, error)
+	SignSystemTx(ctx context.Context, namespace string, data any) (*sign.Transaction, error)
+	SignerAddress() string
```
namrapatel, line 15: Is it necessary to provide SignerAddress via a getter? Looks like we could do this by making it a public var in NakamaSigner and TxSigner if we wanted to, but I'm guessing this means that the interface will not require a public var named `SignerAddress`, but maybe that's fine? 

No big deal since it's only one time and cuz there is a relevant interface though.

namrapatel (approved):

relay/nakama/signer/signer.go
jerargus: There's no way to include a public var in an interface; interfaces can only contain methods. 

There's also no straightforward way to reach through an interface to the underlying implementation to get a specific field. You would have to do a type assertion.

technicallyty, line 15: is the `SignerAddress` just the public key? maybe this method should be called `PublicKey` instead

technicallyty (approved):
LGTM

relay/nakama/signer/signer.go
jerargus: It's not just the public key; It's generated using [crypto.PubkeyToAddress](https://pkg.go.dev/github.com/ethereum/go-ethereum/crypto#PubkeyToAddress)

PR #588, opened 2024-02-06
feat(cardinal): base shard transaction query iterator

technicallyty:
## Overview

adds a transaction iterator to the `router` package. this is done to reduce the amount of responsibilities the `RecoverFromChain` method has to deal with. Now, RecoverFromChain just supplies a closure that defines actions to take on the queried data.


cardinal/router/iterator/iterator.go
```diff
@@ -0,0 +1,138 @@
...
+	"google.golang.org/protobuf/proto"
+	"pkg.world.dev/world-engine/cardinal/types/message"
+	shardtypes "pkg.world.dev/world-engine/evm/x/shard/types"
+	shard "pkg.world.dev/world-engine/rift/shard/v2"
+	"pkg.world.dev/world-engine/sign"
+)
+
+// Iterator provides functionality to iterate over transactions stored onchain.
+type Iterator interface {
+	// Each calls `fn` for each tick of transactions it queries. An optional "ranges" may be given which will control
```
pyrofolium, line 18: We have this concept in several places in our code. Maybe it's time to break it out into it's own fundamental package. This will make this PR more complex though as it will become an interface with Generics and change the parameter signature.

Also Iterable is the more appropriate term for an interface. For example a interface for things that are number like is often called Numeric opposed to Number. 

All optional.

pyrofolium (approved):
lgtm

cardinal/router/iterator/iterator.go
technicallyty: not rly sure if this can be generic-ized. 

i think iterable would make sense if this was an open ended interface meant to be implemented by many different types, but i'm only doing interface here so that it can be mocked easily in tests.

pyrofolium: It can, most languages have this concept generic-ized. I left it as optional because it would detract from this PR due to the complexity. 

```
type Iterable[T any] interface {
	Each(func(x T))
}
```

the above is the generic interface. One would need to adjust the implementation such that all the parameters and data are collapsed into T. Then this would need to be done for all other things that implement Each. 

Either way I don't really care that much so as I said it's optional. I already approved.

PR #592, opened 2024-02-07
feat(cardinal): allow shutdown to be called from within a system.

pyrofolium:
Closes: WORLD-801

- Changed old shutdown function to private `shutdownEngine`. 
- Created new nonblocking public Shutdown function. 
- Created test to see if this new `Shutdown` function can be called from within systems. 

This was broken anyway and developers can cause a deadlock via `eCtx.GetEngine().Engine().Shutdown()`

What necessitated the need for this feature is the fact that profiling with golang is extremely finicky. Shutting down world, engine and cardinal must all be done properly without any panics, fatals and even a sleeping go routine that shutsdown the engine in X seconds would corrupt the profiling results. 

So basically I just need to create a system that will cleanly Shutdown world.Engine after X amount of ticks in order to get profiling to work.

### DIFF
--- a/cardinal/ecs/engine.go
+++ b/cardinal/ecs/engine.go
@@ -6,12 +6,13 @@ import (
 	"errors"
 	"fmt"
 	"path/filepath"
-	"pkg.world.dev/world-engine/cardinal/router"
 	"reflect"
 	"runtime"
 	"sync"
 	"sync/atomic"
 	"time"
+
+	"pkg.world.dev/world-engine/cardinal/router"
 
 	"google.golang.org/protobuf/proto"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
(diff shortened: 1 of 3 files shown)


cardinal/ecs/engine.go
```diff
@@ -759,7 +760,16 @@ func (e *Engine) IsGameLoopRunning() bool {
 	return e.isGameLoopRunning.Load()
 }
 
-func (e *Engine) Shutdown() error {
+func (e *Engine) Shutdown() {
```
jerargus, line 763: Shutdown should initiate a graceful shutdown that doesn't return until everything is off, so spinning off a goroutine here isn't appropriate. 

If this works for profiling, and need to be called from a system, why not just call `go e.Shutdown()` from the System?

pyrofolium: The problem here is if you call Shutdown from within a system. Then Shutdown will block the entire Tick because it's in the system which is in the tick itself. 

Profiling is the thing that necessitated this feature but it's a bit orthogonal.

pyrofolium: We discussed this with Scott, Jeremy and Tyler. I just won't do this because eCtx is going to be removed.

PR #595, opened 2024-02-08
chore(cardinal): making benchmarking cleaner.

pyrofolium:
Closes: WORLD-791

The other PRs related to profiling had big issues when attempting for the benchmarker to self terminate. The only way I've found to get a clean and uncorrupted benchmarking file is to manually close the docker-compose with ctrl-C. 

So this PR is to just fully clean up the benchmark. It cannot be automated as of now. 

Again the commands are:

`make e2e-benchmark`

Let it run for about 100 ticks

`Ctrl-C`

Then run: `go tool pprof -http=:8080 profiles/cpu.prof` from project folder to view the call graph or flamegraph.

You can mess with `e2e/testgames/gamebenchmark/main.go` if you want to see if you can find a way to automatically close the benchmark and still retrieve a full profile.


e2e/testgames/gamebenchmark/main.go
```diff
@@ -123,7 +71,31 @@ func main() {
 		panic(eris.ToString(err, true))
 	}
 
-	initsystems, systems = initializeSystems(testFlag, testKeys, initsystems, systems)
+	// TODO: Unused decorator somehow this breaks profiling (it ends up with missing pieces of the stack).
+	// TODO: Figure out how to get this to work.
```
namrapatel, line 75: I think at some point previously we had decided to remove uncommented code and put it in a Linear ticket for the task to deal with that feature? Can we do that here?

namrapatel (approved):
We should change the PR title to something more specific before merging

e2e/testgames/gamebenchmark/main.go
pyrofolium: sure.

jerargus (approved):

PR #596, opened 2024-02-08
refactor(nakama): break global vars into local vars and pass in as params

namrapatel:
Closes: WORLD-741

## Overview

There are several global vars at the top of our nakama plugin's `main.go` file. We create these global vars locally now and pass them in as params to the functions that need them instead.

## Testing and Verifying

- No new tests, existing tests passing.


technicallyty (approved):
LGTM

PR #597, opened 2024-02-09
test(nakama): Make a fake version of nakama module and update some tests

jerargus:
Closes: WORLD-808

## Overview

Create a "fake" version of NakamaModule and use it in some tests.

The fake implementation uses a map to store information. It's suitable for tests that write and read information to the DB and do not use any search functionality or versioning functionality.

This allows us to write tests that focus on public facing methods (i.e. rpc handlers). In addition, the sequence of database reads and writes don't need to be known ahead of time. Tests also look cleaner because the setup code only uses the public facing methods.

## Testing and Verifying

Unit tests have been modified to use the fake implementation. Some tests have been expanded to cover more edge cases.

### DIFF
--- a/relay/nakama/allowlist_test.go
+++ b/relay/nakama/allowlist_test.go
@@ -5,22 +5,27 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"testing"
+
+	"github.com/heroiclabs/nakama-common/runtime"
+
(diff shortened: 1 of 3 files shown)


technicallyty (approved):
LGTM

namrapatel (approved):

PR #598, opened 2024-02-12
docs: use correct function name for setting TickChannel WorldOption

technicallyty:
## Overview
updates the typo in the docs that uses the wrong option name for setting TickChannel.


namrapatel (approved):
Damn, I could barely wrap my head around the changes in this PR

PR #599, opened 2024-02-12
feat(cardinal): decouple redis from ecb with a new storage interface

pyrofolium:
Closes: WORLD-810

- changed storage under ecs to metastorage (let me know if you want to rename this, idc)
- created new storage interface under gamestate
- ecb now touches only gamestate.Storage and an instance of redis that impliments gamestate.Storage.

A pipeline and redis.Client were very similar concepts under redis. Under this interface a transaction and a storage are identical concepts. If you call `storage.StartTransaction` it will give you a new storage that represents the "transaction". 

When you call `storage.EndTransaction` it will end commit everything in the transaction and return to you the original storage object. 

This interface will of course be modified overtime as requirements change. 

EDIT:
This entire PR has been modified to take into account the massive refactor. I started from scratch and force pushed it. All comments are still relevant but refer to code in the wrong places.


cardinal/ecs/gamestate/redis.go
```diff
@@ -13,10 +13,121 @@ import (
 	"pkg.world.dev/world-engine/cardinal/types/component"
 )
 
+type RedisStorage struct {
+	OtherClient   redis.Cmdable
+	CurrentClient redis.Cmdable
```
technicallyty, line 18: these should be made private fields

should also comment what the differentiation of the clients is for. not sure if OtherClient is the best word either. if we could do something like `txClient` and `client` that might be a bit more clear

cardinal/ecs/gamestate/redis.go
```diff
@@ -13,10 +13,121 @@ import (
...
+func (r *RedisStorage) Close(ctx context.Context) error {
+	return eris.Wrap(r.CurrentClient.Shutdown(ctx).Err(), "")
+}
+
+func (r *RedisStorage) StartTransaction(_ context.Context) (Storage, error) {
+	pipeline := r.CurrentClient.TxPipeline()
+	redisTransaction := NewRedisStorage(pipeline)
+	redisTransaction.OtherClient = r.CurrentClient
+	return &redisTransaction, nil
+}
```
technicallyty, line 100: couldn't we just use the same `RedisStorage` to start a transaction, rather than creating & returning a new one? seems like we could just flip clients internally

cardinal/ecs/gamestate/redis.go
pyrofolium: Basically it swaps the transaction and the client back and forth as you enter transactions. cachedClient isn't appropriate either imo. I'll think of something.

cardinal/ecs/gamestate/redis.go
pyrofolium: Mutation is generally bad for coding. Should be avoided whenever possible as a rule of thumb. Better to create a new thing and the old thing retains the same behavior. This is my philosophy and it is how go-redis creates a transaction as well.

Specifically though for this use case there is an issue with out intuition on how storage is used: 

Imagine two references to the same RedisStorage, A and B. 
 ```
 A = &redisStorage.New()
 B = A
 
 process 1:
    B.write(3)
    
process 2: 
   A.startTransaction() // returns void transfroms B into transaction. 
   A.write(1)
   A.write(2)
   A.EndTransaction()
```

I expect 1 to be written after 2. So only these two outputs are possible:
1,2,3
3,1,2


But if A is changed to a transaction that could also change B and B can be part of the transaction leading to a potential race condition. We can now see 3 possible outputs: 
1,3,2
1,2,3
3,1,2

Which is against our intuition of how the transaction in process 2 should work. The transaction is supposed to block other writes. This mistake is prevented if startTransaction returns a new thing that represents a transaction. I'll re-emphasize that go-redis does this exact same pattern. I think the confusion here lies in the fact that I don't have a "Transaction" type I just reuse the RedisStorage type for the transaction. 

Type Aliasing could be used here if you want for clarity. I'm neutral about that. 
`type Transaction = Storage`

File: cardinal/ecs/gamestate/redis.go (hunk 1 above)
Comment (pyrofolium) [reply cardinal/ecs/gamestate/redis.go:18]: Actually you know what, you're right, there's no other name I can think of. Let's go with cached

File: cardinal/ecs/gamestate/redis.go (hunk 2 above)
Comment (pyrofolium) [reply cardinal/ecs/gamestate/redis.go]: As a side note. Rust semantics makes this bug I'm talking about here impossible from happening.

Comment (pyrofolium) [reply cardinal/ecs/gamestate/redis.go]: @technicallyty can you take a look when you get the chance also prob better to talk in person today if you want.

Comment (pyrofolium) [reply cardinal/ecs/gamestate/redis.go]: Sorry one other thing to add here. If we do it mutation style, there will have to be mutexes to make sure that mutating RedisStorage is thread safe. Doing it this way we just copy the reference address which is thread safe with no locking required.

File: cardinal/gamestate/redis.go [hunk 3]
```diff
@@ -12,10 +13,121 @@ import (
 	"github.com/rotisserie/eris"
 )
 
+type RedisStorage struct {
+	cachedClient  redis.Cmdable
+	currentClient redis.Cmdable
+}
```
jerargus, line 19: Add `var _ Storage = &RedisStorage{}` here to be explicit about what this struct is implementing.

cardinal/gamestate/Storage.go
```diff
@@ -0,0 +1,24 @@
+package gamestate
+
+import (
+	"context"
+)
+
+// Storage is the interface for all available stores related to the game loop
+// there is another store like interface for other logistical values located in `ecs.metastorage`
+type Storage interface {
```
jerargus, line 9: Renaming an existing struct/interface, so that the newly freed up name can be used in a new concept is really confusing when trying to review code. Call this "PrimitiveStorage" and revert the "MetaStroage" back to what it was originally.

After the new functionality is in, if we're unhappy with the names we can make a PR that only changes the names.

cardinal/gamestate/ecb.go
```diff
@@ -19,7 +20,7 @@ import (
 var _ Manager = &EntityCommandBuffer{}
 
 type EntityCommandBuffer struct {
-	client *redis.Client
+	storage Storage
```
jerargus, line 23: I'm super on board with removing the ECB's direct dependency on redis. 👍

cardinal/ecs/gamestate/redis.go
technicallyty: when i see `StartTransaction` i just don't really expect to get another copy of the storage object back. i think this would be nicer if we had a transactor interface or something. the whole client swapping seems a bit messy to me

jerargus (changes requested):
I'm not on board with introducing the term "meta" in this context. We already have "component" and "metacomponent" as a convention in our code base, but this use of meta feels different. I suggest undoing the file renames that introduces the metastore directory.

In general, I agree with removing ECB's dependency on redis and replacing it with an interface, but I haven't been able to fully wrap my head around the implementation details.

cardinal/ecs/gamestate/redis.go
pyrofolium: @technicallyty You are NOT getting a copy of Storage back. 

You are getting a transaction back. This is Exactly what go-redis does.  

It's just the Type of storage and type of transaction are identical. They have the same methods and the same interface this allows for things to operate on BOTH transaction and client in the future. We can use a type alias here if you want to separate concepts b

I can do the type alias here of:

`type Transaction = Storage`

this is actually deliberately crippling functions that accept storage as a parameter as those function can operate on both a transaction and a storage. It's solved by a quick typecast of `Transaction(storage)` or something of that nature so it's not a big deal to me. 

Would you prefer this?

cardinal/gamestate/Storage.go
pyrofolium: Hmmm I think we both agree that it Both "storages" need different names. But in my opinion the gamestate is the "primary" storage here and the other "storage" is more secondary. So I would just rename the other storage instead of the one in gamestate. I can do the opposite of what you suggest here, changing metastorage to primitivestorage. Would that work? 

Honestly though I don't care too much about the naming here, I can still use your suggestion here, but I present the one above as an alternative I'm a bit more in favor of, and if you agree we can go with that. If not I'll go with your initial suggestion, not a huge deal to me. Lmk.

jerargus: In this PR I'd prefer to leave any already-existing interfaces, strcuts, packages as-is, and make new names for whatever new things you're introducing. I think that's limited to the new interface in "Stroage.go" as well as the redis implementation in "cardinal/gamestate/redis.go". I suggest naming those PrimitiveStorage and RedisPrimitativeStorage and reverting the other names.

pyrofolium:
> I'm not on board with introducing the term "meta" in this context. We already have "component" and "metacomponent" as a convention in our code base, but this use of meta feels different. I suggest undoing the file renames that introduces the metastore directory.
> 
> In general, I agree with removing ECB's dependency on redis and replacing it with an interface, but I haven't been able to fully wrap my head around the implementation details.

I can explain the implementation. 

Basically StartTransaction() returns a new Transaction type similar to the pipe from go-redis. 

However I'm using the Storage interface and RedisStorage as a both the interface to a transaction and a client. 

In my implementation transaction and storage are identical concepts. Similar to how both a float and an int are identical. 1 + 1.232 works, and randomoperation(client) and randomoperation(transaction) also work. You can reference Tylers complaint up above if you want.

cardinal/ecs/gamestate/redis.go
technicallyty: right, i know its a transaction you get back, but it LOOKS like you're getting just another storage object. my two main gripes here:

1. client swapping looks messy and confusing to me
2. `StartTransaction` should be `BranchedClient`, `TxClient` or something that indicates you're getting a new object back. the word `StartXXX` doesn't really seem like a method name that would return another object - which is why i initially suggested struct mutation (which i see now isn't the best route).

technicallyty: also type aliasing won't necessarily cripple anything, because any functions that accept the `Storage` interface will still accept `Transaction` interface

pyrofolium: > also type aliasing won't necessarily cripple anything, because any functions that accept the `Storage` interface will still accept `Transaction` interface

You're right, I assumed there would have to be a cast here. But I just tried it, there's no way to type alias a generic type. Type aliasing is not a solution.

pyrofolium: No this works. @technicallyty I can do the above and I believe everything that accepts a transaction will accept a storage. Are you good with this?

pyrofolium: Also sorry, a later PR makes Storage generic. I was commenting on that.

technicallyty: yeah i think that might be better, if @jerargus could sign off on that id say go for it. only concern i have left after that is the client juggling stuff. but im not really sure how to make it cleaner

pyrofolium: @jerargus I talked to tyler about it in my one on one. He told me he's going to talk to you about it I think he's relatively neutral about your suggestion. I personally disagree with it. 

I think it's easier to discuss this in person, I'll wait for tuesday or monday.

technicallyty: talked with both @jerargus & @pyrofolium and it seems like the best path forward is:

- remove the double client business
- EndTransaction just returns an error. if someone needs the original storage, and the pipeline, they need to be explicit about it, and not expect a transaction to give them their OG client. (hopefully that makes sense).

I think aliasing the the storage interface for `Transaction` will still be a nice addition. just so everything is very clear-cut.

pyrofolium:
@technicallyty @jerargus All concerns resolved. Let me know if there's more.

jerargus (approved):

namrapatel (approved):

cardinal/ecs/gamestate/redis.go
pyrofolium: > My view: The receiver for the StartTransaction method and the thing returned are fundamentally different things and should not be treated the same. Yes, the redis package says they are both Cmdable, but under the hood it doesn’t work like that. e.g. doing a read operation from a TxPipeline will fail.
> 
> I think the write methods from PrimitiveStorage should be moved into a separate interface (PrimitiveTransaction) which has Set, Incr, Decr, Delete, and EndTransaction. PrimitiveStorage has all the Get* methods, as well as StartTransaction and Close. StartTransaction would return a PrimitiveTransaction.
> 
> This also means the implementations of the methods should be split across 2 structs. One struct only knows about a redis.Client, and the other only knows about a redis.Pipeliner.

I'm combing over old reviews to clarify situations because in a unscheduled meeting with Tyler he mentioned my PRs have bad design and it's not "golike" enough so I want to find out where and why and pinpoint concretely the issues here.

Just want to comment here the above design written by jermey is actually I believe a mistaken design. PrimitiveStorage cannot have Set, Incr and Decr removed because the PrimitiveStorage interface needs to write data outside of a transaction. It's not supposed to be ONLY reads outside of a transaction and ONLY writes inside of a transaction. Correct me if I'm mistaken here but this was not your intention @jerargus? It's sort of a harmless error here and you likely meant something else. Just documenting this here, feel free to ignore.

PR #603, opened 2024-02-14
fix(nakama): cardinal receipts not being routed to clients

namrapatel:
Closes: WORLD-849

Cardinal receipts are not being routed via Nakama to clients properly. This PR includes a few bug fixes to solve that, and an e2e test to verify that the happy path works. More e2e tests to come on this topic in upcoming sprints.

## Changelog
- Bugfix: `Verifier` is missing Persona creation receipts
- Bugfix: `Dispatcher` gets clogged if more than one receipt is added to it's input channel back to back
- `TestReceipts` test that attempts to receives cardinal receipts from Nakama's `Notifier`
- Get receipts alongside events in `ListNotifications` helper in `e2e`


e2e/testclients/clientutils/util.go
```diff
@@ -44,55 +44,70 @@ type NotificationItem struct {
...
 	}
 	err = json.Unmarshal(bodyData, &data)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 	c.notificationCursor = data.CacheableCursor
-	acc := make([]*Content, 0)
+	receipts := make([]*Receipt, 0)
+	events := make([]*Event, 0)
```
pyrofolium, line 91: Put an estimated number in here it prevents multiple allocation calls. It is a Trivial optimization and it doesn't always matter but good habit to have.

e2e/testclients/clientutils/util.go
```diff
@@ -111,7 +126,7 @@ func (c *NakamaClient) RegisterDevice(username, deviceID string) error {
 	if err != nil {
 		return err
 	}
-	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("Event-Type", "application/json")
 	// defaultkey is the default server key. See https://heroiclabs.com/docs/nakama/concepts/authentication/ for more
```
pyrofolium, line 130: Same, content-type should be the key, not event-type.

e2e/testclients/clientutils/util.go
```diff
@@ -44,55 +44,70 @@ type NotificationItem struct {
...
+}
+
 type NotificationCollection struct {
 	Notifications   []NotificationItem `json:"notifications"`
 	CacheableCursor string             `json:"cacheableCursor"`
 }
 
-func (c *NakamaClient) ListNotifications(k int) ([]*Content, error) {
+func (c *NakamaClient) ListNotifications(k int) ([]*Receipt, []*Event, error) {
 	path := "v2/notification"
```
pyrofolium, line 63: Function signature is a bit busy here. It returns three values. Better to split this into two functions. ListEvents and ListReceipts, with ListNotifications as an underlying base call and the filtering for subject line happening in ListEvents and ListReceipts.

relay/nakama/receipt/dispatcher.go
```diff
@@ -32,33 +32,34 @@ type Receipt struct {
 // Dispatcher continually polls Cardinal for transaction receipts and dispatches them to any subscribed
 // channels. The subscribed channels are stored in the sync.Map.
 type Dispatcher struct {
-	ch chan *Receipt
+	ch chan []*Receipt
```
pyrofolium, line 35: Rather then making a channel of an array of receipts you can make a buffered channel. A buffered channel will not block until the buffer is full. 

You mentioned "clog" in your PR description. Did you mean deadlock? Can you explain how the deadlock occurs? Looks like as long as Dispatch is running as a go routine it can never deadlock. 

It looks to me there can only be dropped messages. Is this what you mean by "clogged". What were you seeing here?

pyrofolium (changes requested):
see comments.

e2e/testclients/clientutils/util.go
```diff
@@ -144,7 +162,7 @@ func (c *NakamaClient) RPC(path string, body any) (*http.Response, error) {
 	if err != nil {
 		return nil, err
 	}
-	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("Event-Type", "application/json")
```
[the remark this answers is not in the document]
namrapatel: Good catch, this was caused by find and replace

relay/nakama/receipt/dispatcher.go
namrapatel: I can explain, Jeremy and I talked about this one and decided to go with batched (a slice of) receipts here because if you think of Cardinal as the producer, and the Dispatcher as the middle man, and the Notifier as the producer, what I was seeing was that the consumer was not always able to take off receipts from the channel as fast as the producer was putting them on.

Buffering them is what I first thought of but we cannot know how many receipts Cardinal is going to send over in a tick and if that number ever exceeds the buffered size we are going to drop receipts. That's why @jerargus and I agreed that using a slice of receipts in this case works better since no receipts will be dropped.

pyrofolium (changes requested):

relay/nakama/receipt/dispatcher.go
pyrofolium: @namrapatel I talked to jeremy. To summarize, 

He told me the main purpose of the slice as opposed to a buffered channel here was to use NotificationsSend as opposed NotificationSend (without the s) such that it can operate on a slice.

He also stated that the purpose for dispatcher for receipts purpose was to have multiple subscribers on the Producer for the future and he doesn't want one subscriber to potentially block it. 

I can agree with the two reasons above.

pyrofolium (approved):
lgtm.

relay/nakama/persona/verifier.go
```diff
@@ -88,56 +87,72 @@ func (p *Verifier) cleanupStaleEntries(now time.Time) {
...
+			// We're missing a success/failure message from cardinal or the initial request from the
+			// user to claim a persona tag.
+			return nil
+		}
+		// We have both a user ID and a success message. Save this success/failure to nakama's storage system
+		ctx := context.Background()
+		ctx = context.WithValue(ctx, runtime.RUNTIME_CTX_USER_ID, pending.userID) //nolint:staticcheck // its fine.
+		ptr, err := LoadPersonaTagStorageObj(ctx, p.nk)
+		if err != nil {
+			return eris.Wrap(err, "unable to get persona tag storage obj")
```
jerargus, line 139: Now that this loops over a number of hashes, this should accumulate the errors in a slice and return them with a errors.Join. With this current state of the code, a problematic early item in txHashes will cause the rest of the verifications to fail.

e.g. If txHashes has 10 items to verify, and there's a problem with the 5th, items 6, 7, 8, and 9 won't be verified.

jerargus (changes requested):

jerargus (approved):

PR #604, opened 2024-02-14
feat(cardinal): create storage interface for map replace compvalues with storage interface.

pyrofolium:
Closes: WORLD-815

- rename storage on ecb to dbStorage to indicate IO/database/persistence. 
- replace compValues with Storage interface
- implimented storage interface for a generic golang map. This is temporary, just for decoupling while maintaining unit test integrity. Eventually we will get rid of the map and replace it with something faster. After removal the implementation will always be there should we need it in the future for testing or anything else. 

This is the first of a couple PRs where all stateful primitives in ECB will be decoupled via Storage interfaces. This will allow for the replacing the storage with something else and/or modifying and upgrading the ECB against a storage interface rather then redis or maps.

There is a possibility of modifying this to work with go-cache but go-cache abstracts away transactions.


cardinal/gamestate/doc.go
```diff
@@ -65,7 +65,7 @@ value:  JSON serialized bytes that can be deserialized to a list of transactions
 processed in the last started tick. This data is only relevant when the START-TICK number does not match the END-TICK
 number.
 
-# In-memory storage model
+# In-memory dbStorage model
```
namrapatel, line 68: The changes from `storage` to `dbStorage` in this file should probably go back to `storage`. Looks like this comes from the refactor tool

cardinal/gamestate/map.go
```diff
@@ -0,0 +1,272 @@
+package gamestate
+
+import (
+	"context"
+
+	"github.com/rotisserie/eris"
+)
+
+type MapStorage[K comparable, V any] struct {
```
namrapatel, line 9: Looks like `MapStorage` is meant to adhere to the `PrimitiveStorage` interface, maybe we should add the following to ensure it:

`var _ MapStorage = &PrimitiveStorage{}`

cardinal/gamestate/map.go
```diff
@@ -0,0 +1,272 @@
...
+}
+
+func NewMapStorage[K comparable, V any]() *MapStorage[K, V] {
+	return &MapStorage[K, V]{
+		internalMap: make(map[K]V),
+	}
+}
+
+func (m *MapStorage[K, V]) Keys(_ context.Context) ([]K, error) {
+	res := make([]K, 0)
```
namrapatel, line 20: I think we should initialize `res` with size `len(m.internalMap)`

cardinal/gamestate/map.go
```diff
@@ -0,0 +1,272 @@
...
+	delete(m.internalMap, key)
+	return nil
+}
+
+// Does nothing for now
+func (m *MapStorage[K, V]) StartTransaction(_ context.Context) (Transaction[K], error) {
+	return m, nil
+}
+
+// Does nothing fow now
```
namrapatel, line 264: nit: typo, fow -> for

cardinal/gamestate/map.go
```diff
@@ -0,0 +1,272 @@
...
+	var res uint64
+	var resAny any
+	resRaw, ok := m.internalMap[key]
+	if !ok {
+		return 0, eris.New("does not exist in map")
+	}
+	resAny = resRaw
+	res, ok = resAny.(uint64)
+	if !ok {
+		return 0, eris.New("cannot convert to float")
```
namrapatel, line 80: Error messages don't seem to match the type they are converting to

cardinal/gamestate/map.go
```diff
@@ -0,0 +1,272 @@
...
+	var res float32
+	var resAny any
+	resRaw, ok := m.internalMap[key]
+	if !ok {
+		return 0, eris.New("does not exist in map")
+	}
+	resAny = resRaw
+	res, ok = resAny.(float32)
+	if !ok {
+		return 0, eris.New("cannot convert to float")
```
namrapatel, line 66: Maybe we should specific float32 in this error msg, given we also have a float64?

cardinal/gamestate/map.go
pyrofolium: Because both MapStorage and PrimitiveStorage are generic this can't be done. 

It compiles anyway so the compiler recognizes the genericness of it.

cardinal/gamestate/map.go
pyrofolium: The linter for some reason wants this at 0. Looking at the performance here there is the potential for multiple malloc calls if done this way though. But whatever the linter won't let this go through if I don't do it.

cardinal/gamestate/map.go
technicallyty: this works: `var _ PrimitiveStorage[string] = &MapStorage[string, any]{}`

cardinal/gamestate/map.go
technicallyty: it's because the first argument to `make` specifies size. if you append to it, you'll be adding an item at index len(m.internalMap) + 1. 

you need to change this to `res := make([]K, 0, len(m.internalMap)`

namrapatel (approved):

PR #605, opened 2024-02-15
chore(cardinal): rename TxQueue to TxPool and move it to types

technicallyty:
## Overview

rename TxQueue struct to TxPool

Moved it to `types` directory

renamed constructor to `New`


pyrofolium (approved):
lgtm. Plz review my code too.

namrapatel (approved):

smsunarto:
I prefer this to be `txpool` (geth also uses this term: https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-txpool)

Technically, we also don't (and shouldn't) enforce FIFO guarantees anyway.

technicallyty:
> I prefer this to be `txpool` (geth also uses this term: https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-txpool)
> 
> Technically, we also don't (and shouldn't) enforce FIFO guarantees anyway.



> I prefer this to be `txpool` (geth also uses this term: https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-txpool)
> 
> Technically, we also don't (and shouldn't) enforce FIFO guarantees anyway.

sure, however i want to highlight that their pool has both queued and pending transactions. ours just has queued transactions, which is why i went with queue. i don't really have a preference either way though.

technicallyty:
> I'm just saying this because a queue technically means that a transaction that comes first _must_ be processed first

[PriorityQueues](https://en.wikipedia.org/wiki/Priority_queue#:~:text=In%20computer%20science%2C%20a%20priority,before%20elements%20with%20low%20priority.) are a thing 😉.

ultimately its up to the ordering of their systems that determine when something gets processed, but i think that's a separate idea. queues are just about spitting out data thats given to them. what a process does with that, and how long it takes to do it, is not something that should be considered for whether something is a queue or not.

PR #608, opened 2024-02-16
DO NOT SUBMIT: demonstration of coverage limitation

jerargus:
DO NOT SUBMIT.

This PR is meant to test our coverage tool. The new code should have 100% coverage, but since package "internal" isn't being directly tested by "internal", I wonder if we'll actually see the full coverage.


smsunarto:
going to close this to not clutter the PR inbox

PR #609, opened 2024-02-18
refactor(cardinal): make NewWorldContextForTick private

smsunarto:
Closes: WORLD-XXX

## Overview

> Description of the overall background and high-level changes that this PR introduces

`NewWorldContextForTick` is a public function when it's intended only for internal use. This PR makes the function private.

## Brief Changelog

- Make NewWorldContextForTick private

## Testing and Verifying

- N/A

PR #610, opened 2024-02-19
feat(cardinal): recover using historical tick timestamp

smsunarto:
Closes: WORLD-809

## Overview

Currently, Cardinal re-executes systems with the current timestamp when recovering. This will cause non-deterministic behavior on system that relies on the tick's timestamp. 

This PR modifies the recovery behavior to re-execute the tick with the historical timestamp from the router recovery data.

## Brief Changelog

- Add `timestamp` parameter to `world.Tick()` that controls the timestamp of the tick
- Update `RecoverFromChain` to call `world.Tick()` with the  historical timestamp passed in via param
- Add a mock for `router/Iterator` to make it easier to test recovery behavior

## Testing and Verifying

- Added a BDD-style using [Goblin](https://github.com/franela/goblin) which is suitable for specifying complex behavior such as the recovery system.


smsunarto:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/Argus-Labs/world-engine/610?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* `main`
  * **PR #609** <a href="https://app.graphite.dev/github/pr/Argus-Labs/world-engine/609?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #610** <a href="https://app.graphite.dev/github/pr/Argus-Labs/world-engine/610?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/Argus-Labs/world-engine/610?utm_source=stack-comment).

smsunarto:
### Merge activity

* **Feb 19, 1:34 AM EST**: [Graphite](https://app.graphite.dev/github/pr/Argus-Labs/world-engine/610) rebased this pull request after merging its parent, because this pull request is set to merge when ready.

technicallyty:
i'd would suggest making a new function, TickWithTimestamp(..., timestamp uint64) and then move the impl that the current `Tick` to a private method `doTick` or something of that nature. That way you would only have to update the recover from chain's call to tick, and avoid having to update all the current calls to `Tick`.

this is a good explanation of that thought process:
https://go.dev/blog/module-compatibility

jerargus:
I agree with Tyler's suggestion.

smsunarto:
> i feel it would be better to make a new method: `TickWithTimestamp(..., timestamp uint64)` and then move the impl of `Tick` to a private method `doTick` or something of that nature that `Tick` and `TickWithTimestamp` would call into. That way you would only have to update the `RecoverFromChain`'s calls, and avoid having to update **all** the current calls to `Tick` with the new parameter.
> 
> ```go
> func (w *World) Tick(ctx context.Context) error {
>     return doTick(ctx, time.Now().UnixWhatever())
> }
> 
> func (w *World) TickWithTimestamp(ctx context.Context, timestamp uint64) error {
>     return doTick(ctx, timestamp)
> }
> 
> func (w *World) doTick(ctx context.Context, timestamp uint64) error { 
>     // ... impl
> }
> ```
> 
> this is a good explanation of that thought process: https://go.dev/blog/module-compatibility

I considered this, but decided against this for the following reason:
1. `world.Tick()` should be private to reduce the surface area of the public API and to avoid issues with external package (i.e. tests and user code) calling `world.Tick()` out of order. Ticking should be handled by the game loop goroutine and we can use the tick channel if we want to let user controller tick externally. This is tracked via https://linear.app/arguslabs/issue/WORLD-828/make-worldtick-private-method
2. There is already too many ambiguous methods related to tick(ing)/gameloop and i want to avoid making this even more ambiguous by introducing more methods: `tickTheEngine`, `StartGame`, `startGameLoop`,

jerargus (approved):
With this understanding that Tick is going to become private with WORLD-828, I'm OK with this PR.

PR #614, opened 2024-02-20
feat(cardinal): recover state from chain on startup

namrapatel:
Closes: WORLD-824

## Overview

- Call `RecoverFromChain()` from within `StartGame()` if there is a router set.
- Previously `worldstage` was moving from `Recovering` to `Ready` after recovering from chain, now we move to `Starting`
- Implement `FakeIterator` that allows us to pass in a slice of TxBatches and Tick pairs
- Fix `TestWorldRecovery` to not call `RecoverFromChain()` manually anymore since that is called in `StartGame()` now
- Fix `TestTransactionsSentToRouterAfterTick` to expect a new call to `TransactionIterator()`

## Tests
- Test recovery from chain using the new `FakeIterator` and passing in pre-defined TXs and asserting that the world moves to the expected state


cardinal/engine_test.go
```diff
@@ -344,6 +346,10 @@ func TestTransactionsSentToRouterAfterTick(t *testing.T) {
 	tx := &sign.Transaction{PersonaTag: "ty"}
 	_, txHash := world.AddEVMTransaction(fooMessage.ID(), msg, tx, evmTxHash)
 
+	fakeIterator := mocks.NewFakeIterator(nil)
+
+	rtr.EXPECT().TransactionIterator().Return(fakeIterator).Times(1)
+
```
technicallyty, line 352: i see this is added because the only requirement for running RecoverFromChain is if theres a router or not. maybe this should just be if the world is in production?

cardinal/engine_test.go
```diff
@@ -370,93 +376,106 @@ func TestTransactionsSentToRouterAfterTick(t *testing.T) {
...
-//
-//	err = world.RecoverFromChain(context.Background())
-//	assert.NilError(t, err)
-//
-//	assert.Equal(t, fooMessages, 2)
-//	assert.Equal(t, world.CurrentTick(), uint64(2))
-// }
+func TestRecoverFromChain(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
```
technicallyty, line 381: you no longer need to call ctrl.Finish(). hovering over the method shows this:

"New in go1.14+, if you are passing a *testing.T into NewController function you no longer need to call ctrl.Finish() in your test methods."

cardinal/engine_test.go
```diff
@@ -370,93 +376,106 @@ func TestTransactionsSentToRouterAfterTick(t *testing.T) {
...
+
+	rtr := mocks.NewMockRouter(ctrl)
+	rtr.EXPECT().Start()
```
technicallyty, line 384: its good to be very specific when u use mocks. set the return values (if any) and the amount of times it should be called.

cardinal/router/iterator/iterator.go
```diff
@@ -48,7 +48,8 @@ func New(
 	}
 }
 
-// Each iterates over txs
+// Each iterates over txs from the data availability layer. For each batch of transactions found in
+// each tick, it will apply the callback function to that batch and it's respective tick and timestamp.
```
technicallyty, line 52:
```suggestion
// Each iterates over txs from the base shard. For each batch of transactions found in
// each tick, it will apply the callback function to that batch and it's respective tick and timestamp.
```

technically doesn't do DA reading

cardinal/router/mocks/iterator.go
technicallyty, line 1: you should never add code to generated code files. once the mocks are re-generated, all this code will be lost.

I'd suggest just moving this to the test file itself.

cardinal/router/mocks/iterator.go
```diff
@@ -0,0 +1,55 @@
...
+				return fmt.Errorf("start tick %d is greater than stop tick %d", startTick, stopTick)
+			}
+		}
+	}
+
+	for _, val := range f.objects {
+		// Skip this iteration if the tick is before the startTick or after the stopTick (if stopTick is specified).
+		if val.Tick < startTick || (stopTick != 0 && val.Tick > stopTick) {
+			continue
+		}
```
technicallyty, line 46: we don't really need to add any of these checks. the goal of your mock here is to just pipe data into the `Each` closure, so that's all it should do.

technicallyty: eh, we then end up with the problem where anytime `Each` is updated, the mock will need to be updated as well. and all this logic is already tested in iterator_test.go.

cardinal/world.go
```diff
@@ -347,14 +347,19 @@ func (w *World) StartGame() error {
...
 	if err != nil {
 		return err
 	}
 
+	// If Cardinal is in Prod and Router is set, recover any old state of the engine from the chain
+	if getWorldConfig().CardinalMode == RunModeProd && w.router != nil {
+		if err := w.RecoverFromChain(context.Background()); err != nil {
+			return eris.Wrap(err, "failed to recover from chain")
+		}
+	}
```
technicallyty, line 361: i think i'd prefer to put `mode` as a field on World rather than read the config again

namrapatel: I agree, first thing I did was check world to see if it had a config! 🔥

technicallyty (approved):
looks alright to me... i guess...

PR #615, opened 2024-02-21
docs: typos on cql queries

rmrt1n:
Closes: WORLD-XXX

## Overview

Some of the examples used `CONTAIN` instead of `CONTAINS`.


pyrofolium (approved):
lgtm, maybe we should add a feature to allow both CONTAIN and CONTAINS.

PR #618, opened 2024-02-21
feat(cardinal): message api refactor, PART 1

pyrofolium:
Closes: WORLD-840

- Changed original `world.RegisterMessages` to `world.RegisterMessagesByName`
- Changed original `cardinal.RegisterMessages` to `cardinal.RegisterMessagesByName`
- created new property on MessageManager and renamed old property:
```golang
type Manager struct {
	registeredMessagesByName map[string]types.Message //renamed from registeredMessages
	registeredMessagesByType map[reflect.Type]types.Message // new property
	nextMessageID            types.MessageID
}
```
- In `func NewMessageType[In, Out any]` there is validation logic to make sure the generic parameters `In` and `Out` are structs. I extracted that logic into an external function called `func validateIfStruct[T any]() bool`
- Created new function `RegisterMessage[In any, Out any](world *World, name string, opts ...message.MessageOption[In, Out]) error` This function is what replaces the instantiation of the `MessageType` global
- Created new function `GetMessageFromWorld[In any, Out any](world *World) (*message.MessageType[In, Out], error)` this replaces usage of the `MessageType` global. This function is more for testing purposes as it needs the world to work. 
- Created new function `GetMessage[In any, Out any](wCtx engine.Context) (*message.MessageType[In, Out], error)` it also replaces usage of the `MessageType` globals. This function is more for use in systems as it needs the wCtx context. 
- Created accessor on world: `GetMessageManager` that allows access to private value MessageManager
- Created `GetMessageByType` on wCtx, MessageManager and world. 

Essentially the technique here is that the user NEVER creates a `MessageType`. All he does is Register the `In` and `Out` generic parameters. 

Internally MessageType is created and saved on two maps. One map is based on a string name, the other map is based on the type of `MessageType[In Out]`

Retrieval of this MessageType happens without needing any parameters other then generic parameters. It's done via reflection meaning that inside GetMessage[In, Out] a new MessageType[In, Out] is created and then with reflection the type is found and that type is fed into GetMessageByType(). Thus retrieval can be done without a string. 

Right now this change is an addition on top of the original way of using Messages. As a result it's possible for users to actually use the code incorrectly. For example the user can Register a message the old way and call the Message the new way and that will generate an error. 

This is done in order to make the change modular and more incremental. I will be doing this in several PRs that will eventually remove the usage of the global all together but wanted to get this up first so people can object before a major change is made. 

I changed one test to use the new way. See: `TestCannotWaitForNextTickAfterEngineIsShutDown`


cardinal/world.go
```diff
@@ -207,6 +208,41 @@ func NewMockWorld(opts ...WorldOption) (*World, error) {
 	return world, nil
 }
 
+func RegisterMessage[In any, Out any](world *World, name string, opts ...message.MessageOption[In, Out]) error {
+	msgType := message.NewMessageType[In, Out](name, opts...)
+	var msg types.Message = msgType
+	err := world.registerMessagesByName(msg)
+	if err != nil {
+		return eris.Wrap(err, "failed to register message")
+	}
```
technicallyty, line 217: no need for the extra `msg` variable here, the `msgType` _is_ `types.Message` already, so it can be passed it to methods that accept `types.Message`

cardinal/message/message.go
```diff
@@ -30,6 +31,15 @@ type MessageType[In, Out any] struct { //nolint:revive // this is fine for now.
 	outEVMType *ethereumAbi.Type
 }
 
+func validateIfStruct[T any]() bool {
```
technicallyty, line 34: can we just call this `isStruct`?

technicallyty, line 34: a unit test for this function would be nice

cardinal/world.go
```diff
@@ -207,6 +208,41 @@ func NewMockWorld(opts ...WorldOption) (*World, error) {
...
+	var ok bool
+	msg, ok = world.GetMessageManager().GetMessageByName(name)
+	if !ok {
+		return eris.Errorf("Failed to register message with name: %s", name)
+	}
```
technicallyty, line 222: this returns an error for the case where this message does not exist yet. needs to be the other way around

also, we should provide more context in the error message. i.e. `this message has already been registered`

cardinal/world.go
```diff
@@ -207,6 +208,41 @@ func NewMockWorld(opts ...WorldOption) (*World, error) {
...
+	typeValueOfMessageType := reflect.TypeOf(*msgType)
+	return world.GetMessageManager().RegisterMessageByType(typeValueOfMessageType, msg)
```
technicallyty, line 224: i'm confused on the purpose of registering by type this way. we are storing by type of msgType, which is `MessageType`, so that people can get `types.Message` by type right? but if they have the `MessageType` already, then they have a `types.Message`?

pyrofolium: type.Message can't be keyed by type because it's just an interface. It's the same type. 

However MessageType[K, V] is different for all possible Ks and Vs. That's why it can be used as a key value here. 

The purpose of registering the type this way is such that it can be retrieved via a generic function:

`GetMessage[K, V](wCtx)` this is effectively what eliminates the global.

cardinal/world.go
pyrofolium: I took this out. This logic was redundant.

cardinal/world.go
pyrofolium: Depending on the code, the other `map[string]types.Message` on the message manager will eventually be removed, once all dependencies and tests that utilize it are eliminated (if they can be at all).

technicallyty: got it, missed the GetMessage[K, V](wCtx) part

cardinal/message/message.go
```diff
@@ -31,7 +31,7 @@ type MessageType[In, Out any] struct { //nolint:revive // this is fine for now.
 	outEVMType *ethereumAbi.Type
 }
 
-func validateIfStruct[T any]() bool {
+func IsStruct[T any]() bool {
```
technicallyty, line 34: eh not really a fan of making something public just to test it. i'd say either:

- make a unit test file to test this directly
- indirectly test it via public method (can probs just add on this test: `TestReadTypeNotStructs` (not sure why that test has the wrong name actually) to ensure it works on both pointers to structs and non-ptr structs)

technicallyty (commented):
looks like the right path. i'd say we can proceed here

cardinal/message/message.go
pyrofolium: I'm not sure what you mean by make a unit_test to test this directly. But I went with that route because it seems the most straightforward. 

I made a file called isstruct_test.go that's part of the same message package. Let me know if that fits into what you were looking for.

cardinal/world.go
```diff
@@ -207,6 +208,35 @@ func NewMockWorld(opts ...WorldOption) (*World, error) {
 	return world, nil
 }
 
+func RegisterMessage[In any, Out any](world *World, name string, opts ...message.MessageOption[In, Out]) error {
```
smsunarto, line 211: This should be in `cardinal/cardinal.go` where we keep all of our other `Register<...>` functions at. The whole idea here is to have exactly 1 way to do something so the API surface area is as narrow as possible and you don't have random function lying around everywhere.

cardinal/message/message.go
pyrofolium: We talked about it on slack and agreed to go with the second route here.

cardinal/world.go
pyrofolium: Just note, right now there's multiple ways of registering messages. (The old way still works) I'm going to change this incrementally with multiple PRs to keep things readable and manageable. I will be slowly changing it across multiple PRs to the new way and once all dependencies are resolved I will remove everything associated with the "old way"

And yes I will move this to cardinal.go.

cardinal/message/manager.go
```diff
@@ -1,27 +1,31 @@
...
+		nextMessageID:            1,
 	}
 }
 
-// RegisterMessages registers multiple messages with the message manager
+// RegisterMessagesByName registers multiple messages with the message manager
 // There can only be one message with a given name, which is declared by the user by implementing the Name() method.
 // If there is a duplicate message name, an error will be returned and none of the messages will be registered.
-func (m *Manager) RegisterMessages(msgs ...types.Message) error {
+func (m *Manager) RegisterMessagesByName(msgs ...types.Message) error {
```
smsunarto, line 28: To be consistent with the rest of the manager, this should be `RegisterMessages`

cardinal/world_context.go
```diff
@@ -46,6 +50,21 @@ func NewReadOnlyWorldContext(world *World) engine.Context {
 	}
 }
 
+func GetMessage[In any, Out any](wCtx engine.Context) (*message.MessageType[In, Out], error) {
```
smsunarto, line 53: As per my other comment, MessageType is an internal type and this abstraction violation and will create tight coupling.

I also saw that this is only used once for a test, so I don't see a strong need for this either.

cardinal/world_context.go
```diff
@@ -62,6 +81,10 @@ func (ctx *worldContext) Logger() *zerolog.Logger {
 	return ctx.logger
 }
 
+func (ctx *worldContext) GetMessageByType(mType reflect.Type) (types.Message, bool) {
```
smsunarto, line 84: This is only used by the GetMessage in `world_context.go` that as I mention we shouldn't have, so we should remove this.

cardinal/message/manager.go
```diff
@@ -1,27 +1,31 @@
...
+
 	"github.com/rotisserie/eris"
 	"pkg.world.dev/world-engine/cardinal/types"
-	"slices"
 )
 
 type Manager struct {
-	registeredMessages map[string]types.Message
-	nextMessageID      types.MessageID
+	registeredMessagesByName map[string]types.Message
```
smsunarto, line 12: with the removal of `registeredMessagesByType` we can change this back to `registeredMessages` and keeping it consistent with the rest of the managers

cardinal/message/manager.go
```diff
@@ -50,7 +54,7 @@ func (m *Manager) RegisterMessages(msgs ...types.Message) error {
 		}
 
 		// Register message
-		m.registeredMessages[msg.Name()] = msg
+		m.registeredMessagesByName[msg.Name()] = msg
```
smsunarto, line 57: registeredMessageByType is not updated here, but it doesnt matter because we should remove registeredMessageByType (i left a comment directly on it on why we dont need it). Simple API -> one less thing to screw up

smsunarto (changes requested):

cardinal/world.go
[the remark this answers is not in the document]
pyrofolium: For some reason the previous reply didn't get under here, I'll paste it again:

Just note, @smsunarto, right now there's multiple ways of registering messages. (The old way still works) I'm going to change this incrementally with multiple PRs to keep things readable and manageable. Once all dependencies are resolved I will remove everything associated with the "old way" in a final PR.

And yes I will move this to cardinal.go.

cardinal/world.go
```diff
@@ -296,11 +326,15 @@ func (w *World) Tick(ctx context.Context) error {
 	return nil
 }
 
+func (w *World) GetMessageManager() *message.Manager {
```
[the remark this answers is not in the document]
pyrofolium: Yeah, I know. I'm fully aware of this trade off. Here's the thing, I can't think of any other way to make this work with this signature:

`func GetMessage[K, V](wCtx)`

The problem is it needs to access a private var but GetMessage is generic and it can't be a method so it can't get access to the internal var that I need. Thus the only way is to make certain things public. 

This PR is made small because I want people to see the tradeoff I made and discuss it before going further. Let me know what you think and if you have any ideas on how to approach this.

cardinal/world.go
```diff
@@ -207,6 +208,35 @@ func NewMockWorld(opts ...WorldOption) (*World, error) {
...
+	}
+	typeValueOfMessageType := reflect.TypeOf(*msgType)
+	return world.GetMessageManager().RegisterMessageByType(typeValueOfMessageType, msgType)
+}
+
+func (w *World) registerMessagesByName(msgs ...types.Message) error {
+	return w.msgManager.RegisterMessagesByName(msgs...)
+}
+
+func GetMessageFromWorld[In any, Out any](world *World) (*message.MessageType[In, Out], error) {
```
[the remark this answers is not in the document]
smsunarto: I'm confident that`GetMessageFromWorld` and `GetMessage` should not exist; this looks fine now, but then everyone wants to code and merge their PR quickly, people also review PR sloppily, and this sooner or later this will just be a nest of tech debt. 

I saw that you added this only for the `TestCannotWaitForNextTickAfterEngineIsShutDown` test, which honestly a good showcase the pile of tech debt the codebase is dealing with.

`AddEVMTransaction` violates abstraction boundaries by operating with `ID` that should have been for internal records only. If it just operates on the string name identifier, then you wouldn't have this issue in the first place because you name in advance what the name of the message is going to be.

Additionally, there is just a larger problem with tests in this codebase in general, because everyone wants to integration tests, every time they want to violate abstraction boundaries to access an internal state they add a getter to something that shouldn't exist in the first place. I say that if we want to keep this test, just turn into a unit test in package `cardinal` so it can do this abstraction boundaries violation it needs to in the first place and just take the L on it until we untangle the tight coupling.

cardinal/world.go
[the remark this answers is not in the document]
smsunarto: see comments i left above, imo there is just cascading complexity explosion here that will just make your life worse.

also, i'm pretty sure you do `wCtx.world.msgManager` since it's both in the same package.

cardinal/message/manager.go
pyrofolium: Sure I don't mind, I was just renaming it to keep it consistent witht he lower level method call.

cardinal/world.go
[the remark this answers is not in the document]
technicallyty: just want to add that this isn't exclusive to `AddEVMTransaction`. It's signature is 1:1 with `AddTransaction`, except it has an extra parameter for the EVM tx source hash.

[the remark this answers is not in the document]
smsunarto: yeah i think both is an issue, i'm just using it as an example becauss it's the first thing i saw in the test mentioned above

cardinal/message/manager.go
```diff
@@ -79,13 +83,27 @@ func (m *Manager) GetMessageByID(id types.MessageID) types.Message {
...
+	msg, ok := m.registeredMessagesByName[name]
+	return msg, ok
+}
+
+func (m *Manager) GetMessageByType(mType reflect.Type) (types.Message, bool) {
+	msg, ok := m.registeredMessagesByType[mType]
 	return msg, ok
 }
 
+func (m *Manager) RegisterMessageByType(mType reflect.Type, message types.Message) error {
```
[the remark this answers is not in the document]
pyrofolium: This is used by:

`RegisterMessage[In any, Out any](world *World, name string, opts ...message.MessageOption[In, Out]) error`

It's not used by GetMessage, that's another method, it's easy to get mixed up. The problem is similar to the one I described above. If I want to have a generic function to access a Registered value, that generic function cannot be a method by nature of being generic. 

If it can't be a method, then that means it can't access private variables so I have to sort of create a public route to the underlying method. 

As I stated previously I'm adapting the code with trade offs to allow for functions such as this: 
`RegisterMessage[In any, Out any](world *World, name string, opts ...message.MessageOption[In, Out]) error`
and this:
`GetMessage[In any, Out any](wCtx engine.Context) (*message.MessageType[In, Out], error)`

I'm definitely on board with changing the function signatures above or finding alternative ways to not have this access path. Very much open to your suggestions here.

cardinal/message/manager.go
pyrofolium: sure don't mind.

cardinal/world.go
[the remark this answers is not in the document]
pyrofolium: @smsunarto I agree with your comment on testing. But throwing in huge test refactors along with this change in a single PR makes it hard to read and really stressful for everyone. So my proposal is to do it pieces. Leave this `GetMessageFromWorld` so I can get the refactor in along with good tests to validate it then, remove GetMessageFromWorld and fix the tests in a later endeavor. I can create a ticket and put in a todo comment. 

I'm not sure I understand the part of GetMessage

GetMessage is the primary motivation behind this PR. It's so that users don't have to refer to a global. 

No global value: 
 
 `FooTx.Each(wCtx, ...)` 
 
 instead users do 
 
 `GetMessage[FooIn, FooOut](wCtx)`
 
 This was objectively the primary purpose I was trying to fulfill for this PR. Let me know if that wasn't what you had in mind. This was what I thought was the goal as explained in Tylers description here: https://linear.app/arguslabs/issue/WORLD-840/message-api-refactor

cardinal/message/manager.go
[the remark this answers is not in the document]
smsunarto: This is meant for `GetMessageByType` i believe. 

For the way to do this cleanly, see my code for RegisterComponent. 

If i understand correctly, the goal is to make this the one that is canonical once the refactor is finished? then this is the one who should be `RegisterMessage` and `RegisterMessageByName`/`RegisterMessages` above is the one that needs to be killed. Apologies, this whole "keeping both ways to keep the PR small" thing is confusing me on what i actually need to review, but you do you.

```
func RegisterComponent[T types.Component](w *World) error {
	if w.worldStage.Current() != worldstage.Init {
		return eris.Errorf(
			"engine state is %s, expected %s to register component",
			w.worldStage.Current(),
			worldstage.Init,
		)
	}

	compMetadata, err := component.NewComponentMetadata[T]()
	if err != nil {
		return err
	}

	err = w.componentManager.RegisterComponent(compMetadata)
	if err != nil {
		return err
	}

	return nil
}
```

cardinal/world.go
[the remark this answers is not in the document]
smsunarto: i'm not proposing a complete test refactor, you have exactly 1 function that uses GetMessageFormWorld, just rip that one test into a separate test file that is in the cardinal package.

cardinal/world_context.go
smsunarto: nvm, i understand what you are trying to do now, you still can't use `message.MessageType` though, I would make this private so you don't shoot yourself in the foot. Instead, you need to use `types.Message`.

I would also move this to cardinal.go together with `GetComponent`

cardinal/world.go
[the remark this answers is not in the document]
smsunarto: left follow up comment on GetMessage below. copying it here for clarity

> nvm, i understand what you are trying to do now, you still can't use message.MessageType though, I would make this private so you don't shoot yourself in the foot. Instead, you need to use types.Message.

I would also move this to cardinal.go together with GetComponent

cardinal/world.go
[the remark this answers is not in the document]
smsunarto: left follow up comment on GetMessage below. copying it here for clarity

> nvm, i understand what you are trying to do now, you still can't use message.MessageType though, I would make this private so you don't shoot yourself in the foot. Instead, you need to use types.Message.
>
>I would also move this to cardinal.go together with GetComponent

smsunarto (approved):

cardinal/world_context.go
pyrofolium: Hey so I tried this but it didn't work: 

MessageType is the current type that contains Each which is what needs to be exposed here. It's hard to put Each onto the Message interface because it contains Generic parameters. That's the issue, because it contains generic parameters it will actually pollute the Message interface if I put that function signature on there and it will then pollute everything else that uses the Message interface including Context. Etc. 

There's probably a fix around this, it will take a bit. I will keep this in mind. For now this will go up and I will change it later.

pyrofolium: I agree with this technical direction and I prefer it over the current method. For the record I won't be going with it due to time constraints. I think the current way actually works fine though not ideal. I talked with @jerargus (correct me if I'm wrong Jeremy) that you didn't want to block the ticket so you won't request changes. I am going to resolve the ticket and merge because of that. We can always roll back if you change your mind.

cardinal/world.go
pyrofolium: I'm reviewing old PRs to find out issues and complaints. 

@technicallyty In this review, You are incorrect. The logic here is 100% viable and you are making an incorrect assumption here. If you look above the code the message is indeed registered by name. Then in the subsequent code it grabs the same message that was registered. 

This was just defensive coding in case the registerByMessageName function was mutating the message. But since it's passed by reference it's redundant logic as side effects from register will effect the original msg regardless.

PR #619, opened 2024-02-21
refactor(cardional): decouple compValuesToDelete in ecb from redis.

pyrofolium:
Closes: WORLD-872

Pretty straightforward. All maps will be replaced with storage eventually.

### DIFF
--- a/cardinal/gamestate/ecb.go
+++ b/cardinal/gamestate/ecb.go
@@ -23,7 +23,7 @@ type EntityCommandBuffer struct {
 	dbStorage PrimitiveStorage[string]
 
 	compValues         PrimitiveStorage[compKey]
-	compValuesToDelete map[compKey]bool
+	compValuesToDelete PrimitiveStorage[compKey]
 	typeToComponent    map[types.ComponentID]types.ComponentMetadata
 
 	activeEntities map[types.ArchetypeID]activeEntities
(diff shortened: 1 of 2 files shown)


jerargus:
My view is we do not want each node in the linked list to use general purpose storage. We should have a series of ECBs that perform all operations using go  build-in types and concepts (which gives us nested transactions), and a final node that uses some persistent store (to allow us to recover from disk on server restarts). 

Generalizing the final node so that we can easily swap out the "persistent storage" portion of game state is worth doing. Trying to generalize layers of ECB so they can store things in different ways is not worth doing.

pyrofolium:
> My view is we do not want each node in the linked list to use general purpose storage. We should have a series of ECBs that perform all operations using go build-in types and concepts (which gives us nested transactions), and a final node that uses some persistent store (to allow us to recover from disk on server restarts).

I think this is definitely a viable direction. In fact this was my initial thoughts on it to use go built in types. @smsunarto has also been pushing for cacheing libraries instead of Go built in types. 

The reason why I'm putting an interface called Storage here is so that we can put any type of storage behind it, That includes Go built in types. I'm totally on board with your direction as well. Scotts library recommendation apparently has something that can work faster then a built in map so that's viable as well, and we can test and benchmark it. 

>Generalizing the final node so that we can easily swap out the "persistent storage" portion of game state is worth doing. Trying to generalize layers of ECB so they can store things in different ways is not worth doing. 

I think it's fine though. I already put a map behind storage and the effort is on me here. That and I'm trying to find middle ground here between your wishes and scotts suggestions. We can benchmark and test both ways here.  

I want to mention one last thing, If we use Storage on every node, then each node becomes identical in type. That is a more elegant design in my opinion. Generalizing the code to support multiple designs we can try out is better also because as a team we're not too familiar with game server design so this can help us try a variety of configurations. 

Anyway, let's talk in person on Thursday. I want to hear all your concerns with my design direction and I want to clarify what I'm doing here as well.

jerargus (approved):
LGTM.

We spoke about this pattern in person, and while I disagree with putting the pending ECB operations behind an interface at this time, I understand that it will let us try different libraries and potentially make things more performant in the future.

PR #621, opened 2024-02-22
feat(cardinal): enable submitting of empty TXs to chain

namrapatel:
Closes: WORLD-846

## Overview

- Remove prerequisite where the `TxPool` must have transactions in it, before submitting a Tx Blob.


technicallyty (changes requested):
test pl0x

PR #623, opened 2024-02-23
feat(cardinal): change api to get rid of creation of messagetype (PART 2)

pyrofolium:
Closes: WORLD-881

Probably a couple more PRs stacked on top of this one so this isn't completely final. Feel free to mention any changes you want in those subsequent PRs here. This one is pretty big as it touches every unit test. There were some trade offs I made here. Please read the bullet points below or ask me about it if you have questions. 

Overall this change is big and is close to completing the overall refactor. 

This one is unfortunately quite big because I changed a bunch of unit tests to take into account the new way of Registering Messages so that I could delete the old way of doing it. 

- manager now has two private methods, `registerMessagesByName` and `registerMessagesByType`
- There is a new function in the `manager.go` called `RegisterMessageOnManager[In, Out](manager *Manager, name string)` that will now handle the registration of messages on the manager. This function also handles instantiation of the `MessageType` AND registration via name and type. 
- `cardinal.GetMessageFromWorld` is used to fetch Messages outside of the system context. Unfortunately this function is still needed as of now as there are situations where this needs to be done. This does however get rid of the need of a global MessageType. _** I can look into trying to find a way around this in a future PR**_
- `func RegisterMessage[In, Out](world, *World, name string)` is the new way to create/register a message on cardinal
- `cardinal.GetMessage[Msg, Result](wCtx)` is used for getting a message in a system.
- RegisterMessage(s) is gone. Users cannot register multiple messages and they cannot instantiate a MessageType. 
- The original `NewMessageType[In, Out](name, opts)` is now private it cannot be used by the user. However there is an identical function made in testutils for usage in tests that require a `MessageType`
- All instances I can find where a MessageType is created and a world exists is now changed to the new way of registering messages. Basically the user does not need to instantiate a MessageType if a world exists. If a world doesn't exist (which only happens for tests) they can use `testutils.NewMessageType`
- On `MessageType` I created public functions SetName and SetGroup. This allows for me to create a  NewMessageType function in the testutils pacackage. With the existence of this new function I can then make the NewMessageType function in the `message` package private. 
- `message_test.go` is made to be part of the `message` package in order to test private functions. This caused a circular dependency for tests using `cardinal`. Tests involving `cardinal` were moved from `message_test.go` to `cardinal_test`, this makes sense as although those tests were testing "messages" they were testing them in a very high level way and using basically cardinal level functions/methods.


cardinal/cardinal_test.go
```diff
@@ -40,6 +44,634 @@ type Rawbodytx struct {
...
+	for i := 0; i < 10; i++ {
+		someMsg, err := cardinal.GetMessageFromWorld[SomeMsgRequest, SomeMsgResponse](world)
+		assert.NilError(t, err)
+		req := SomeMsgRequest{GenerateError: i%2 == 0}
+		txHash := tf.AddTransaction(someMsg.ID(), req, testutils.UniqueSignature())
+		knownTxHashes[txHash] = req
+	}
+
+	// Perform a engine tick
+	assert.NilError(t, world.Tick(context.Background(), uint64(time.Now().Unix())))
```
jerargus, line 97: nit: Ticking should be done with tf.DoTick() (any calls to world.Tick in this test file can be replaced with tf.DoTick())

jerargus (approved):
LGTM, a couple of small comments

cardinal/cardinal_test.go
pyrofolium: There's a deadlock on `TestAddToPoolDuringTickDoesNotTimeout` I commented it with a TODO and created a ticket. https://linear.app/arguslabs/issue/WORLD-883/implement-primitivestorage-for-typetocomponent

All the other ticks are changed.

cardinal/cardinal_test.go
```diff
@@ -40,6 +44,635 @@ type Rawbodytx struct {
...
+
+// TestAddToPoolDuringTickDoesNotTimeout verifies that we can add a transaction to the transaction
+// pool during a game tick, and the call does not block.
+func TestAddToPoolDuringTickDoesNotTimeout(t *testing.T) {
+	tf := testutils.NewTestFixture(t, nil)
+	world := tf.World
+
+	assert.NilError(t, cardinal.RegisterMessage[*ModifyScoreMsg, *EmptyMsgResult](world, "modify_Score"))
+
+	inSystemCh := make(chan struct{})
```
jerargus, line 253: This fixes the deadlock:

```
	inSystemCh := make(chan struct{})
	// Push a value to inSystemCh to verify we're in the middle of a tick. Close inSystemCh at the end of the test
	// to make sure the System exits so that post-test cleanup can take place.
	defer func() { close(inSystemCh) }()
	err := cardinal.RegisterSystems(
		world,
		func(engine.Context) error {
			<-inSystemCh
			<-inSystemCh
			return nil
		},
	)
```

pyrofolium: Still blocks. I say let's fix this in another ticket. It detracts from the current issue I'm dealing with here. I think the refactor of the "ticks" was a bit of detraction but I was fine with it because it's small but it looks like it's ballooning slowly into something a bit bigger.

PR #624, opened 2024-02-23
feat(cardinal): on ecb change typeToComponent to use PrimitiveStorage Interface.

pyrofolium:
Closes: WORLD-883

Straight forward. changes one section of the ecb to use PrimitiveStorage instead of Map. I'm just going to do this one here as it was already in flight before I had a talk with Scott about the alternative implementation. 

We may want to go with PrimitiveStorage[K, V] which is more type safe and involves a lot of changes so I'm going to wait until we have that discussion with everyone before doing any more of these PRs related to decoupling ECB from a concrete storage type.

### DIFF
--- a/cardinal/gamestate/ecb.go
+++ b/cardinal/gamestate/ecb.go
@@ -24,7 +24,7 @@ type EntityCommandBuffer struct {
 
 	compValues         PrimitiveStorage[compKey]
 	compValuesToDelete PrimitiveStorage[compKey]
-	typeToComponent    map[types.ComponentID]types.ComponentMetadata
+	typeToComponent    PrimitiveStorage[types.ComponentID]
 
 	activeEntities map[types.ArchetypeID]activeEntities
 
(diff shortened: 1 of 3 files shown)


cardinal/gamestate/ecb.go
```diff
@@ -24,7 +24,7 @@ type EntityCommandBuffer struct {
 
 	compValues         PrimitiveStorage[compKey]
 	compValuesToDelete PrimitiveStorage[compKey]
-	typeToComponent    map[types.ComponentID]types.ComponentMetadata
+	typeToComponent    PrimitiveStorage[types.ComponentID]
```
jerargus, line 27: I disagree with this. We want to keep type information here; we don't want future maintainers to have to know to cast this to the ComponentMetadata type.

The compValuesToDelete field was more ok because it's just a Set and the 'bool' didn't really matter. The compValues field was more ok because there actually is a mix of types that will be stored in the values, but this field has a specific job with a single type.

pyrofolium: The whole reason for these series of PRs was for replacing ALL maps with the Storage interface. Just to be clear about this. So it's not about whether your ok with one and not ok with the other, the approval should also include a wholistic opinion here, whether you approve ALL maps to be replaced with storage. 

Anyway this was more of a tentative PR, scott discussed a possibility of using a PrimitiveStore[K, V] rather then PrimitiveStore[K], so ALL of PrimitiveStorage might be completely replaced with a new design and that replacement is actually more inline with what you're talking about here (You did however explicitly disagree with the PrimitiveStorage[K, V] last week so I'm not sure about this). Anyway I'll just Draft this for now and we'll talk more when Scott gets back.

PR #629, opened 2024-02-27
feat(cardinal): MessageEach refactor finalization.

pyrofolium:
Closes: WORLD-888

- `GetMessage` is made private, and will after this be referred to as `getMessage`.
- `getMessage` is moved to `cardinal.go`
- `EachMessage` is used in unit tests where it is relevant
- `testutils.GetMessage` is created and it's a copy of `cardinal.getMessage`. `testutils.GetMessage` is to be used in unit tests.

- edit: getMessage has been removed. The logic is inlined into `EachMessage`. There is now only one `GetMessage` function and it lives in testutils.

### DIFF
--- a/cardinal/cardinal.go
+++ b/cardinal/cardinal.go
@@ -68,21 +68,6 @@ func RegisterComponent[T types.Component](w *World) error {
 	return nil
 }
 
-func getMessage[In any, Out any](wCtx engine.Context) (*message.MessageType[In, Out], error) {
-	var msg message.MessageType[In, Out]
-	msgType := reflect.TypeOf(msg)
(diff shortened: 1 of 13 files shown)


cardinal/cardinal.go
```diff
@@ -67,6 +68,21 @@ func RegisterComponent[T types.Component](w *World) error {
 	return nil
 }
 
+func getMessage[In any, Out any](wCtx engine.Context) (*message.MessageType[In, Out], error) {
```
smsunarto, line 71: nit: i would just inline this into EachMessage since it's only used once.

smsunarto (approved):

PR #632, opened 2024-02-27
feat(cardinal): register to base shard on startup

technicallyty:
## Overview

updates cardinal to automatically send a `Register` message over to the base shard when `StartGame` is called. 

## Testing and Verifying

i added an end to end test. to run this:

1. `make rollup` (wait for this to fully build and start producing blocks)
2. `GOWORK=off CARDINAL_MODE=production REDIS_PASSWORD=foo make game` (wait for this to start ticking)
3. uncomment the t.Skip from the test, and then run it as normal.


pyrofolium (approved):
lgtm

cardinal/router/router_test.go
```diff
@@ -203,6 +205,42 @@ func TestRouter_SendMessage_TxFailed(t *testing.T) {
...
+	f.req = in
+	return &shard.RegisterGameShardResponse{}, nil
+}
+
+func (f *fakeTxHandler) Submit(
+	_ context.Context,
+	_ *shard.SubmitTransactionsRequest,
+	_ ...grpc.CallOption,
+) (*shard.SubmitTransactionsResponse, error) {
+	panic("implement me")
```
pyrofolium, line 228: tbh, I don't mind stuff like this. But typically If I did it someone would block this in my PR, hard. Add a todo here at least.

technicallyty: yeah, its just there so it satisfies the interface so i can pass the object in. i changed the panic message: [change panic message in mock](https://github.com/Argus-Labs/world-engine/pull/632/commits/69c95a0b027c3510ba4f6338ef83f304245b0977)

docker-compose.yml
```diff
@@ -108,11 +108,12 @@ services:
     container_name: test_game
     environment:
       - REDIS_ADDRESS=redis:6379
+      - REDIS_PASSWORD=${REDIS_PASSWORD}
       - CARDINAL_NAMESPACE=TESTGAME
       - CARDINAL_LOG_LEVEL=${CARDINAL_LOG_LEVEL:-info}
-      - SHARD_SEQUENCER_ADDR=chain:9601
-      - BASE_SHARD_ADDR=chain:9090
-      - ENABLE_ADAPTER=${ENABLE_ADAPTER:-true}
+      - CARDINAL_MODE=${CARDINAL_MODE:-development}
```
jerargus, line 114: These env variables also exist in docker-compose.benchmark.yml

e2e/chain/e2e_test.go
```diff
@@ -65,3 +69,23 @@ func TestTransactionStoredOnChain(t *testing.T) {
 	assert.NilError(t, err)
 	assert.Check(t, len(txs.Epochs) != 0)
 }
+
+// TestNamespaceSaved ensures that when the stack is running, namespaces can be queried
+// and the address in the namespace can be supplied to a router msg client and send requests.
+func TestNamespaceSaved(t *testing.T) {
+	t.Skip("do not run this test unless the docker containers for base shard, game shard, nakama, and " +
+		"celestia are running. to run the stack, run `make game` and `make rollup`. wait for startup and run this test")
```
pyrofolium, line 77: The dev experience of manually reactivating this test because of an environmental change is not very good. 

Instead this test should be automatic. It should only activate itself if the environment it correct OR the test should be placed at the E2E level where the environment is mocked up. Putting this skip here will pretty much make it unlikely for this thing to ever run EVEN when the set up is correct.

technicallyty: yep, i agree. planning to address this here https://linear.app/arguslabs/issue/WORLD-900/make-e2e-tests-with-evm-part-of-the-ci

i would do it now, but launch. can't spend too much time making things pretty unfortunately

pyrofolium: I already approved. You can ignore all my comments, resolve and have it go up regardless. None of my suggestions should block you at all. 

I suggest putting a todo here as well, in case the ticket gets buried and all context is lost.

e2e/chain/e2e_test.go
```diff
@@ -65,3 +69,23 @@ func TestTransactionStoredOnChain(t *testing.T) {
 	assert.NilError(t, err)
 	assert.Check(t, len(txs.Epochs) != 0)
 }
+
+// TestNamespaceSaved ensures that when the stack is running, namespaces can be queried
+// and the address in the namespace can be supplied to a router msg client and send requests.
+func TestNamespaceSaved(t *testing.T) {
+	t.Skip("do not run this test unless the docker containers for base shard, game shard, nakama, and " +
```
jerargus, line 76: I tried following these instructions and I see the error:

```
   e2e_test.go:80: assertion failed: error is not nil: rpc error: code = Unknown desc = codespace sdk code 26: invalid height: WorldEngine is not ready; please wait for first block
```

e2e/chain/e2e_test.go
```diff
@@ -65,3 +69,23 @@ func TestTransactionStoredOnChain(t *testing.T) {
...
+		"celestia are running. to run the stack, run `make game` and `make rollup`. wait for startup and run this test")
+	chain := newChainClient(t)
+	res, err := chain.namespace.Namespaces(context.Background(), &namespacetypes.NamespacesRequest{})
+	assert.NilError(t, err)
+	assert.Equal(t, len(res.Namespaces), 1)
```
jerargus, line 81: I'm seeing this failure when I run it locally:

```
=== RUN   TestNamespaceSaved
    e2e_test.go:83: assertion failed: 0 (int) != 1 (int)
```

technicallyty: updated the `make` command needed to run the game stack properly.

jerargus: 👍 test is passing with the updated instructions.

PR #634, opened 2024-02-27
docs: remove explicit language about main.go in World docs

technicallyty:
## Overview

removes the explicitness around `main.go` in the `World/Overview` section of the docs. This is now only suggested by the example code, instead of implying requirement.

### DIFF
--- a/docs/cardinal/game/world/overview.mdx
+++ b/docs/cardinal/game/world/overview.mdx
@@ -3,11 +3,11 @@ title: 'Overview'
 description: 'How to create and manage your Cardinal world'
 ---
 
-In Cardinal, the `main.go` file is used as the entry point for your game shard implementation. This is where you will initialize your world and register your component, message, query, and system to it.
+In Cardinal, the `World` serves as the central hub for registering components, messages, queries, and systems. This is where you define and configure the elements that drive your game's functionality, as well as kick off the game's loop-driven runtime.
 
 ## Example
 
-A typical `main.go` file will look like this:
+Typical `World` usage looks like this:
 
 ```go main.go
 package main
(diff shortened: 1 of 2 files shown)


jerargus (approved):

smsunarto (changes requested):
I prefer to still keep the reference to main.go, I don't really understand what benefit replacing it with world outright and if any creates confusion for new devs on where to look.

technicallyty:
> I prefer to still keep the reference to main.go, I don't really understand what benefit replacing it with world outright and if any creates confusion for new devs on where to look.

It's the main bit of text that i dislike here:
```
In Cardinal, the main.go file is used as the entry point for your game shard implementation. This is where you will initialize your world and register your component, message, query, and system to it.
```

someone could register their components/messages/queries/systems each in their own file, so it feels odd to suggest this is the only place you can do that. imo, thats _more_ confusing to beginners as it gives off an incorrect implication.

imo, this is mixing documentation with how go works with Cardinal works, and i think it should be as concise as possible. i feel in this doc we should be just explaining what the world is, how to use it, and show an example usage.

pyrofolium (approved):
Looks fine to me. I think an average user coming in will understand what's going on here. What's left is polish and I'm good with either way.

smsunarto:
> someone could register their components/messages/queries/systems each in their own file, so it feels odd to suggest this is the only place you can do that. imo, thats more confusing to beginners as it gives off an incorrect implication.

My philosophy here is docs should be opinionated as it helps new users where to look. If they are smart enough to have their own opinion and register stuff in their own file, then they know enough that you don't have to do that in main.go. The probability of a user getting confused on which file they need to be registering their stuff is significantly higher than the odds of someone getting confused of whether they can register components outside of main.go

smsunarto:
Going to close this, if there are users who complains about this/ask about this, i'm happy to be proven wrong and revisit.

PR #636, opened 2024-02-28
fix(cardinal): Do not check signer address when signature verification is disabled

jerargus:
Closes: WORLD-894

## Overview

When signature verification is disabled, we should accept any and all transactions that come in. Previously, cardinal would complain if the persona tag of a transaction did not have a signer address. This PR modifies the code so that a persona tag's signer address is not checked at all when signature verification is disabled.

## Brief Changelog

- Add tests demonstrating problem
- Fix the tests

## Testing and Verifying

2 unit tests were added.

### DIFF
--- a/cardinal/server/handler/tx.go
+++ b/cardinal/server/handler/tx.go
@@ -3,12 +3,14 @@ package handler
 import (
 	"errors"
 	"fmt"
+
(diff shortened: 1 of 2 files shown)


pyrofolium (approved):
lgtm

technicallyty (approved):
lgtm

PR #639, opened 2024-02-29
fix(nakama): Make nakama notifications not persistent (they will no longer be save…

jerargus:
…d for disconnected clients)

## Overview

Nakama's notifications system allows for making notifications 'persistent'. Notifications that can't be delivered right away will be saved for the relevant user.

Notifications about receipts should only be delivered to connected clients, so this persistent flag should be false.

See https://heroiclabs.com/docs/nakama/tutorials/unity/pirate-panic/notifications/ for more details about notifications.

This initially broke out e2e test, which depended on persistent notifications. While Nakama doesn't have a dedicated go client, [this forum post](https://forum.heroiclabs.com/t/can-i-get-some-help-to-implement-a-go-realtime-client-api/473) has some details on how to get websockets up and running with Nakama in go.

## Brief Changelog

- Make receipt and event notificaitons not-persistent. 
- Update e2e Nakama client so use a websocket to read notifications


pyrofolium (approved):
lgtm

namrapatel (approved):

PR #640, opened 2024-02-29
fix(nakama): disable persistent nakama notifications for events/receipts

pyrofolium:
Closes: WORLD-901

This change does 3 things:

- previously cardinal passes a string to nakama, This has been changed to a serialized Struct called Event. 
- Event contains a message which is the original string and additional boolean property called `Persistence`
- Persistence is used to set Persistence on notifications in nakama. See: https://heroiclabs.com/docs/nakama/tutorials/unity/pirate-panic/notifications/

Mainly this was done to fix issues in nakama related to persistence. Users can now get rid of this problem by setting it to false. However, unit tests require persistence to be set to true.


smsunarto (changes requested):
i vote we just make this never persistent instead of letting the users choose. Notificaiton persistence is a Nakama concept, not Cardinal. marking things as persistence or not in cardinal violates separation of concern; we can revisit how we can architect this when there's a legitimate user need.

pyrofolium:
> i vote we just make this never persistent instead of letting the users choose. Notificaiton persistence is a Nakama concept, not Cardinal. marking things as persistence or not in cardinal violates separation of concern; we can revisit how we can architect this when there's a legitimate user need.

Yeah I was initially doing this. The problem is e2e tests require persistence to work. Basically to unblock mark he can manually set persistence to false. See here for more info: https://argus-labs.slack.com/archives/C04SKM9PAUB/p1709170261761939

Also I believe mark wants full control over persistence: https://argus-labs.slack.com/archives/C04SKM9PAUB/p1709169340885249

I agree with this overall direction though, but it will be a longer fix as it requires additional fixing to the e2e test suite. Maybe a subsequent PR? What do you think @smsunarto ?

smsunarto:
i prefer punting this and doing it correctly, we need to cut release and this is very API breaking.

pyrofolium:
> i prefer punting this and doing it correctly, we need to cut release and this is very API breaking.

Alright I'll reverse it.

PR #641, opened 2024-02-29
fix(cardinal): change default persistence to false for events.

pyrofolium:
Closes: WORLD-901

old PR closed: https://github.com/Argus-Labs/world-engine/pull/640 Please reference the conversation there for more details. 

This change defaults event notifications to have `persistence = false`. When persistence is true it causes a memory leak bug in nakama. 

In addition to this the Events test is temporarily disabled because it requires persistence = true in order for the e2e test to work. This will be punted to another issue: https://linear.app/arguslabs/issue/WORLD-902/fix-e2e-test-for-events

### DIFF
--- a/e2e/testclients/e2etestclient/testsuite_test.go
+++ b/e2e/testclients/e2etestclient/testsuite_test.go
@@ -17,6 +17,8 @@ import (
 )
 
 func TestEvents(t *testing.T) {
+	t.Skip("this test is skipped as it requires nakama notifications to be set to true " +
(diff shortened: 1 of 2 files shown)


jerargus:
It turns out getting our Nakama client to use websockets wasn't too bad: https://github.com/Argus-Labs/world-engine/pull/639

pyrofolium:
Closed because @jerargus fixed both events and receipts in another PR.

PR #642, opened 2024-02-29
refactor: require router requests to include a persona tag

technicallyty:
## Overview

This PR updates the Router's EVM -> Cardinal message router to include a persona tag. 

That is, when EVM code wants to call `SendMessage` and send a msg to Cardinal, it needs to include the Persona tag its attempting to act as.

This is needed because: previously, if two users authorized the same EVM contract address, the router request would automatically act as the first persona. 

More concretely:

- persona Foo authorizes 0x12345
- persona Bar authorizes 0x12345
- cardinal receives SendMessage request from 0x12345.

which persona should 0x12345 use here? this is what that PR solves.

## Brief Changelog

- update Rift proto package to include `persona_tag` in the `SendMessageRequest`
- update Router precompile to take in a `persona` string in `SendMessage` 
- updated Cardinal's SendMessage server implementation to get the `SignerComponent` component and check it's authorized address list for the sender's 0x address.

## Testing and Verifying

updated existing unit tests to reflect the new field in the system.


technicallyty:
@coderabbitai review

jerargus (approved):

PR #643, opened 2024-02-29
test(cardinal): add create persona test helper

technicallyty:
## Overview

adds a helper method to Cardinal's test fixture to facilitate the creation of persona tags.


jerargus:
Where did you imagine this would be used? For cardinal users writing tests? 

It looks like our `world-engine/cardinal/persona/persona_test.go` could make use of this helper. How about updating that test file to use this helper?

technicallyty:
> Where did you imagine this would be used? For cardinal users writing tests?
> 
> It looks like our `world-engine/cardinal/persona/persona_test.go` could make use of this helper. How about updating that test file to use this helper?

done

jerargus (approved):

PR #645, opened 2024-03-01
perf(cardinal): faster search.

pyrofolium:
Closes: WORLD-878

Search is bounded by this:
    # archetypes: A
    # components per archetype C
    # number of components in a filter query F
    O(ACF)
    if A == C == F == N then the upper bound is O(N^3)
    
This change mitigates the runtime to O(N^2) by saving it to a map. For small N this change is actually slower slightly. For large N this change is significantly faster. A future PR can have the system detect the length of the query and the length of the amount of components and dynamically decide on linear search or map. 

Overall though this is just a slight change as cacheing takes over after the first time Each is called on a search. This just decreases start up time and new searches that aren't cached. For a large amount of archetypes, components and large queries the slow down can hiccup and block the main loop. 

There is an alternative implementation here where on EXACT and CONTAINS, components can be stored on a map instead of an array. The problem with this is it moves a cost of O(N) to query instantiation as the map has to be built every time you call Exact or Contains. While doing this is more "elegant" this change will be unexpected to the user as they expect costly operations to happen on Each, not on constructing the query.

### DIFF
--- a/cardinal/component/component_test.go
+++ b/cardinal/component/component_test.go
@@ -1,11 +1,12 @@
 package component_test
 
 import (
+	"testing"
+
 	"github.com/alicebob/miniredis/v2"
 	"pkg.world.dev/world-engine/cardinal/filter"
 	"pkg.world.dev/world-engine/cardinal/iterators"
 	"pkg.world.dev/world-engine/cardinal/types"
-	"testing"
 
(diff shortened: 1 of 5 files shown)


technicallyty (approved):
lgtm, nice work

jerargus (approved):

PR #648, opened 2024-03-01
chore(e2e): clean up naming in e2e directory, speed up compose command

technicallyty:
## Overview

was having trouble navigating the e2e folder when i was doing chain related work. so this PR addresses the naming of the files/dirs to better convey what they are/whats inside. i also updated the e2e-nakama docker compose command to only spin up the containers it needs, which should speed up e2e testing, as the chain+celestia no longer runs with the suite.


makefiles/test.mk
```diff
@@ -12,7 +12,7 @@ e2e-nakama:
 		cd $(ROOT_DIR); \
 	)
 
-	@docker compose up --build --abort-on-container-exit --exit-code-from test_nakama --attach test_nakama
+	@docker compose up --build game nakama test_nakama --abort-on-container-exit --exit-code-from test_nakama --attach test_nakama
```
technicallyty, line 15: this speeds up the e2e by a TONNNNN. without this line, every single service defined in the compose file gets invoked - meaning the chain + celestia client runs as well 🙈

docker-compose.benchmark.yml
```diff
@@ -146,8 +146,8 @@ services:
   test_nakama_benchmark:
     container_name: test_nakama_benchmark
     build:
-      context: ./e2e/testclients/
-      dockerfile: benchmarktestclient/Dockerfile
+      context: ./e2e/tests/
+      dockerfile: bench/Dockerfile
```
pyrofolium, line 150: bench can mean a number of things. It's short and sweet but IMO, unclear. We should use the full word benchmark here to make the meaning crystal clear.

pyrofolium (changes requested):

pyrofolium:
I'm ok/neutral about all the other renamings.

docker-compose.benchmark.yml
technicallyty: i mean, when you run benchmark tests in Go, the command is `go test -bench=.`, so i don't think this would really be that unclear.

pyrofolium: Good point. You're right. Didn't think of this. If the go command does this then people should know. Approved.

PR #651, opened 2024-03-01
refactor(cardinal): Remove concurrent management fields in favor of the stage manager

jerargus:
Closes: WORLD-820

## Overview

Instead of using a collection of mutexes, channels, and atomic bools to manage the starting and stopping of different stages, just use the "worldstage" package.

## Brief Changelog

- Expand worldstage so that a goroutine can be notified when a specific stage has been reached
- Remove channels/atomics/mutexes from the World struct that were helping track game stage
- Fix some tests that were calling shutdown unnecessarily

## Testing and Verifying

All unit tests have remained generally the same and are still passing.


smsunarto (approved):
thank god, finally

technicallyty (approved):
LGTM

PR #655, opened 2024-03-05
fix(cardinal): Change the type of errors in receipts to strings to ensure the error …

jerargus:
…text is sent back to the client

Closes: WORLD-923

## Overview

In the receipt plugin, the slice of errors were not properly serialized because the "error" type is an interface, and isn't guaranteed to have a JSONMarshal implementation. In practice, I was just seeing empty objects in place of actual errors.

## Brief Changelog

- Change the Receipt errors type to a slice of strings
- Convert internal receipts to strings when querying receipts.
- Add unit test to ensure serialized receipts actually contain the error text.
- Add an end to end test to ensure serialized receipts actually contain the error text.

## Testing and Verifying

Unit and end-to-end tests added.


technicallyty (approved):
lgtm

PR #660, opened 2024-03-05
fix(e2e): clean all containers at the end of e2e test rounds

namrapatel:
Closes: WORLD-928

## Overview

This is a simple change that uses `docker compose down -v` to shutdown all the containers at the end of each nakama e2e test run and delete their volumes. This means each new run will start with clean state but also will not require any image redownloading.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,7 +4,7 @@ services:
     # Only use cockroachdb single-node clusters for non-production environment
     image: cockroachdb/cockroach:latest-v23.1
     command: start-single-node --insecure --store=attrs=ssd,path=/var/lib/cockroach/,size=20%
-    restart: "no"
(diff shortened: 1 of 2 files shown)


pyrofolium (approved):
lgtm

namrapatel:
Moving back to draft because I noticed that redis and cockroach db are not being shutdown and as a result their volumes are still alive. Needs more debugging, none of the commands are working to shut them down.

pyrofolium:
> Moving back to draft because I noticed that redis and cockroach db are not being shutdown and as a result their volumes are still alive. Needs more debugging, none of the commands are working to shut them down.

You can try `docker compose down --volumes -v`

technicallyty (approved):
gg

PR #661, opened 2024-03-06
feat(nakama): Add command line flag that allows for the adjustment of the outgoing …

jerargus:
…queue size

Closes: WORLD-922

## Overview

The [outgoing_queue_size](https://heroiclabs.com/docs/nakama/getting-started/configuration/#socket.outgoing_queue_size) config parameter specifies how many pending notifications are allowed for a connected client before Nakama decides the client is unresponsive (shutting down the connection). 

If many receipts/events are generated in a single tick, the burst of traffic could overload this queue even though both the client and Nakama can handle the traffic. 

This PR exposes the Nakama command line parameter so it can be modified via docker-compose/world.toml.


pyrofolium (approved):
lgtm

PR #662, opened 2024-03-07
feat(nakama): use websockets to stream receipts from cardinal to nakama

namrapatel:
Closes: WORLD-817

- [x] Refactor Cardinal Receipts to go thru that Cardinal Dispatcher, Cardinal Events already go thru here
- [ ] Rename Nakama’s EventHub to be a Nakama Dispatcher (in the same way the current one is) and read Events and Receipts from it’s web socket connection with the Cardinal Dispatcher
- [ ] Refactor Nakama’s Notifier to read from this new Nakama Dispatcher and call `nk.NotificationSendAll()` to forward Events and `nk.NotificationsSend()` to send Receipts
- [ ] Delete Nakama’s current dispatcher.go

### DIFF
--- a/cardinal/message/message.go
+++ b/cardinal/message/message.go
@@ -150,20 +150,10 @@ func (t *MessageType[In, Out]) Each(wCtx engine.Context, fn func(TxData[In]) (Ou
 			)
 			errs = append(errs, err)
 			t.AddError(wCtx, txData.Hash, err)
+			emitReceiptEvent(wCtx, txData.Hash, result, errs)
 		} else {
 			t.SetResult(wCtx, txData.Hash, result)
-			if err = wCtx.EmitEvent(
-				map[string]any{
-					"txHash": txData.Hash,
-					"result": result,
-					"errs":   errs,
-				},
-			); err != nil {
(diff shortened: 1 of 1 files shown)


cardinal/message/message.go
```diff
@@ -228,6 +232,21 @@ func (t *MessageType[In, Out]) GetInFieldInformation() map[string]any {
 	return types.GetFieldInformation(reflect.TypeOf(new(In)).Elem())
 }
 
+func emitReceiptEvent(wCtx engine.Context, txHash types.TxHash, result any, errs []error) {
+	if err := wCtx.EmitEvent(
+		map[string]any{
+			"txHash": txHash,
+			"result": result,
+			"errs":   errs,
+		},
```
pyrofolium, line 241: This is a draft so just leaving a comment here. For this we should expose EmitJSONEvent to engine.Context to avoid this extra step of map conversion. 

Ideally this exposure should be "private". When users themselves want to launch an event they can use EmitEvent for type safety across nakama as well. I'm a little undecided about this as EmitJSONEvent eliminates an extra conversion to map step, but it sacrifices static safety as a result so not sure which one (or both) to make public.

PR #665, opened 2024-03-07
feat: make world.Tick() private func

zulkhair:
Closes: WORLD-828

## Overview

Make `world.Tick()` private func

## Brief Changelog

- Change `world.Tick()` to `world.doTick()`
- Adjust all test to use `doTick()` func
- Modify `tickTheEngine()` to throw panic in every error on `doTick()`

## Testing and Verifying

- Adjusted unit test


zulkhair:
* **#665** <a href="https://app.graphite.dev/github/pr/Argus-Labs/world-engine/665?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @zulkhair and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

cardinal/world.go
```diff
@@ -213,15 +213,9 @@ func (w *World) CurrentTick() uint64 {
...
-//  This is a footgun because you want to make sure world is started using StartWorld() instead of calling Tick()
-//  directly. This is unfortunately used a lot in tests.
-//  One thing we need to do is to make the tickDone channel return the tick number AND error if any so it can be used
-//  to check if there are errors when trying to tick.
-
-// Tick performs one game tick. This consists of taking a snapshot of all pending transactions, then calling
+// ticks performs one game tick. This consists of taking a snapshot of all pending transactions, then calling
 // each System in turn with the snapshot of transactions.
-func (w *World) Tick(ctx context.Context, timestamp uint64) error {
+func (w *World) ticks(ctx context.Context, timestamp uint64) (err error) {
```
technicallyty, line 218: any reason for the plurality here?

smsunarto: Name collision with struct prop `tick`

zulkhair: No reason, i'm just haven't found another word to replace tick, so I added an s after it.
Yes tick already used by var tick. so it will collided.

we can change it to doTick() maybe?

cardinal/world.go
```diff
@@ -49,7 +49,7 @@ const (
 type World struct {
 	server          *server.Server
 	tickChannel     <-chan time.Time
-	tickDoneChannel chan<- uint64
+	tickDoneChannel chan types.TickResult
```
jerargus, line 52: nit: `chan<- types.TickResult` specifies a channel that can only have data pushed to it, which is how World should be using the tickDone channel.

cardinal/world.go
jerargus: +1 to doTick

cardinal/world.go
```diff
@@ -429,20 +432,21 @@ func (w *World) startGameLoop(ctx context.Context, tickStart <-chan time.Time, t
...
+	if err != nil {
 		bytes, err := json.Marshal(eris.ToJSON(err, true))
 		if err != nil {
 			panic(err)
 		}
 		w.Logger.Panic().Err(err).Str("tickError", "Error running Tick in Game Loop.").RawJSON("error", bytes)
 	}
 	if tickDone != nil {
-		tickDone <- currTick
+		tickDone <- types.TickResult{CurrentTick: currTick, Error: err}
```
jerargus, line 449: Since a non-nil error on line 441 always results in a panic, this Error field will always be nil.

We should either 1) reserve the tickDone channel for successfully completed ticks (reverting this channel's type back to a uint64 instead of a struct) or 2) Ensure errors actually make it out of this function by putting this if statement between lines 440 and 441. 

My weakly held opinion is option 1.

cardinal/world.go
zulkhair: I'm reading TODO comment from @smsunarto at line `219` before changes : 
> One thing we need to do is to make the tickDone channel return the tick number AND error if any so it can be used to check if there are errors when trying to tick.

But if there is no one using this error in the future i'll change it to send only channel `chan<- types.TickResultinstead` instead of bidirectional channel `chan types.TickResult`

cardinal/world.go
zulkhair: updated to doTick

cardinal/world.go
zulkhair: ohh, i think panic should be only called when `json.Marshal` returning error, so the code should be like this :
```
err := w.doTick(ctx, uint64(time.Now().Unix()))
if err != nil {
	bytes, errMarshal := json.Marshal(eris.ToJSON(err, true))
	if errMarshal != nil {
		panic(errMarshal)
	}
	w.Logger.Panic().Err(err).Str("tickError", "Error running Tick in Game Loop.").RawJSON("error", bytes)
}
```

The return of error in line 449 now is used by `DoTick` func in `TestFixture`, it's on `cardinal/testutils/world.go`

jerargus: Per our conversation: We DO want w.Logger.Panic()... to panic, so Msg needs to be attached to that line.

This way, any tick errors that are encountered will results in a panic, causing cardinal to shut down.

the tickDone channel doesn't need an error field in this case.

The test in tick_test.go "TestIfPanicMessageLogged" becomes a little more complicated since it no longer has access to the public Tick function. I suggest reverting this test back into its original code, and moving it to a new file called `world_test.go`, with a package of `cardinal`. This will give you access to the private doTick function.

In general we should try to put our tests in `<package>_test` to force us to only test public facing APIs, but this situations seems like an appropriate place to deviate from that rule.

cardinal/world.go
zulkhair: reverted to `chan<- uint64` because tick always throw panic if there is any error

jerargus (approved):

PR #666, opened 2024-03-07
feat(cardinal): persona tag character limit

jerargus:
Closes: WORLD-819

## Overview

Ensure persona tags are between 3 and 16 characters long.

## Brief Changelog

- Updated persona tag validation logic (Checking the length directly is more performant than including it in the regexp)
- Moves regexp pre-compilation to the global scope

## Testing and Verifying

tests added


technicallyty (approved):
LGTM

PR #677, opened 2024-03-12
feat(nakama): use websockets to stream receipts from cardinal to nakama

namrapatel:
Closes: WORLD-817

- [x]  Refactor Cardinal Receipts to go thru the Cardinal EventHub, Cardinal Events already go thru here
- [x] Batch `Tick`, `Events`, and `Receipts` into a `TickResults` object and push that thru the EventHub rather than sending individual events
- [x]  Refactor Nakama's EventHub to read events from the incoming `TickResults` and dispatch them
- [x] Refactor Nakama’s Notifier to read from Nakama's EventHub and call nk.NotificationSendAll() to forward Events and nk.NotificationsSend() to send Receipts
- [x] Fix e2e and unit tests
- [x] Delete Nakama’s current dispatcher.go

Note: 
- Delete eventQueue in Cardinal EventHub?


relay/nakama/events/events.go
```diff
@@ -73,24 +95,36 @@ func (eh *EventHub) Shutdown() {
...
 			continue
 		}
 		if messageType != websocket.TextMessage {
 			eh.Shutdown()
 			continue
 		}
+		receivedTickResults := TickResults{}
+		err = json.Unmarshal(message, &receivedTickResults)
+		if err != nil {
+			log.Error("unable to unmarshal message into TickResults: ", err)
```
jerargus, line 113: This still attempts to emit the empty receivedTickResults. I think you want a continue here.

relay/nakama/events/events.go
```diff
@@ -73,24 +95,36 @@ func (eh *EventHub) Shutdown() {
...
+
 		eh.channels.Range(func(_ any, value any) bool {
-			channel, ok := value.(chan []byte)
-			if !ok {
-				err = eris.New("not a channel")
-				eh.Shutdown()
-				return false
+			switch ch := value.(type) {
+			case chan []byte:
+				for i := 0; i < len(receivedTickResults.Events); i++ {
```
jerargus, line 119: nit: 
`for _, e := range receivedTickResults.Events {
  ch <- e
}`
Is more idiomatic.

jerargus (changes requested):

relay/nakama/events/events.go
```diff
@@ -45,8 +52,18 @@ func CreateEventHub(logger runtime.Logger, eventsEndpoint string, cardinalAddres
 	return &res, nil
 }
 
-func (eh *EventHub) Subscribe(session string) chan []byte {
-	channel := make(chan []byte)
+func (eh *EventHub) Subscribe(session string, channelType interface{}) interface{} {
+	var channel interface{}
+
+	switch channelType.(type) {
```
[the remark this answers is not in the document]
namrapatel: Done

relay/nakama/events/events.go
namrapatel: Done

relay/nakama/events/events.go
namrapatel: Nice catch! Done

pyrofolium (approved):
lgtm

edit: wait let me look at one more thing.

cardinal/events/tick_results.go
```diff
@@ -0,0 +1,44 @@
+package events
+
```
pyrofolium, line 2: This should live in cardinal imo. EventHub doesn't rely on it as a dependency. Cardinal uses it to aggregate events and receipts then it passes it to eventhub. Cardinal is the one that depends on both eventhub and TickResults, while eventHub does not depend on TickResults. 

I'll leave this as optional because it can be moved later.

cardinal/events/tick_results.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"encoding/json"
+	"github.com/rotisserie/eris"
+	"pkg.world.dev/world-engine/cardinal/receipt"
+)
+
+type TickResults struct {
+	Tick     uint64
+	Receipts []receipt.Receipt
+	Events   [][]byte
```
pyrofolium, line 12: Tick results is serialized into bytes, but events which is already bytes is serialized again. We should leave it as a concrete type here, (perhaps any or map[string]any)

pyrofolium (changes requested):
Only one change, and one optional change. The optional change you can resolve it if you don't feel like changing it.

cardinal/events/tick_results.go
namrapatel: Done, left as `any`

cardinal/events/tick_results.go
namrapatel: Moved!

jerargus (approved):

pyrofolium (approved):
lgtm

PR #678, opened 2024-03-13
feat(cardinal): Use sliding window to track nonces

jerargus:
Closes: WORLD-910

## Overview

Update our nonce management to keep track of a sliding window of valid nonces. The sliding window shifts each time a new highest nonce is seen. Nonces below the sliding window are rejected outright, so there is no need to keep track of those nonces in redis. As the sliding window shifts, nonces in redis will fall outside the sliding window. Periodically, these irrelevant nonces are removed from redis.

This pattern ensures:
- The memory dedicated to keeping track of used nonces does not grow without bounds.
- Nonces can be sent in slightly out of order
- Used nonces are persisted across Cardinal server restarts

[Design Doc](https://www.notion.so/arguslabs/Nonce-Management-7d88e8b7571243c7939f97b790e097e2)

## Testing and Verifying

Unit tests added


pyrofolium (approved):
lgtm

cardinal/storage/redis/nonce.go
```diff
@@ -3,35 +3,140 @@ package redis
...
+	removed, err := r.Client.ZRemRangeByScore(ctx, key, minScore, maxScore).Result()
+	if err != nil {
+		log.Err(err).Msg("failed to remove old nonces")
+		return
+	}
+	r.countNonce[key] -= int(removed)
+}
+
+// getMaxNonceForKey returns the highest used nonce for the given key.
+func (r *NonceStorage) getMaxNonceForKey(ctx context.Context, key string) (uint64, error) {
```
namrapatel, line 120: Should we be more specific and clarify that the `key` is in fact just a `signerAddress`? Seems like that is the only possible thing we would use here so being specific and clear makes sense as opposed to leaving it generic?

namrapatel (approved):

cardinal/storage/redis/nonce.go
```diff
@@ -3,35 +3,140 @@ package redis
...
+)
+
+var ErrNonceHasAlreadyBeenUsed = errors.New("nonce has already been used")
+
 type NonceStorage struct {
 	Client *redis.Client
+	// mutex locks the UseNonce function to make it safe for concurrent access. This is a single lock for all signer
+	// addresses. An improvement on NonceStorage would have a different lock for each signer addresses.
+	mutex      *sync.Mutex
+	maxNonce   map[string]uint64
```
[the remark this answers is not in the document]
jerargus: Added comments

cardinal/storage/redis/nonce.go
jerargus: Updated to signerAddressKey. It's not just the signer address, but also common prefix to help identify all these keys: `USED_NONCES_<some-signer-address>`

PR #679, opened 2024-03-14
feat(cardinal): fix flakey tests with features in eventhub. 

pyrofolium:
Closes: WORLD-939

- makes `RegisterConnection` and `UnregisterConnection` blocking. These functions will block until a websocket connection is finished and placed into eventHub
- Created blocking methods to read the amount of connections eventHub has and to read the amount of events in the eventHub queue. 
- Deleted ThreadSafeBuffer as it's unused. 
- Created new function NewTestFixtureWithPort. The reason why is because the random ports that is selected by the original Test fixture constructor could crash the test if run enough times. This is rare, I do not believe it happens often enough to show up on CI, but it did show up locally. 
- Fixed TestEvents
     - Two bugs. The testfixture would sometimes use a port that was unavailable. This was fixed with NewTestFixtureWithPort. This bug is probably so rare that it will never happen in CI. It only showed up as I was running tens of thousands of tests on my local computer. 
     - The second bug involved a race condition. websocket clients return the "ok" way to quickly while eventHub is still processing. This is an issue with the websocket on fiber. 
     
- Fixed TestEventsThroughSystems.
     - This is just the testfixture bug above

Description of the bug:

Basically when a connection is first established by eventHub we need to do register the connection and save it to a map inside eventHub. The connection request comes through fiber and is handed off to eventhub through RegisterEvent. From the client perspective all of this should happen and BLOCK until the connection is registered. This does not occur. 

The problem occurs because fiber gives the client the "OK" message that everything is fine before eventHub registers the connection. Essentially what fiber is doing is as soon as you enter into the websocket handler that has the websocket connection itself, the fiber server has already given the websocket client the ok to begin sending messages. So it does not wait for Registering the connection with eventHub in this handler. The framework doesn't really provide a way to block this (there could be a way but that would take much more work to find it). 

What the test was doing was it was moving faster than eventHub. It sent messages to eventHub when one of the connections wasn't finished registering.

The bug is minor and in production it will show up as events that are dropped on initialization if the client responds too quickly. For tests it can only be fixed by blocking the entire test until all connections are sensed to be registered. These are what the getters are for to see how many connections eventHub has. There is an unused method to sense how many messages are in the eventHub queue that was used in debugging but not in the final PR. I left that method in there because it could be useful in the future. 

Additional issues: 

- There is a potential race condition that was also found and fixed that is unrelated to the race condition above. 

Basically on eventHub you can:
1. RegisterConnection
2. EmitEvent(event)
3. the connection registered in 1. will not receive the event because it has not finished registering before the emitevent was called. I changed this function to blocking in order to eliminate the above race connection. 

- fiber will also crash on start up occasionally. I only encountered this once every 7000 tests or something like that. It's so rare and obscure that I didn't touch it.

### DIFF
--- a/cardinal/events/events_test.go
+++ b/cardinal/events/events_test.go
@@ -1,19 +1,19 @@
 package events_test
 
 import (
-	"bytes"
 	"encoding/json"
 	"fmt"
 	"sync"
 	"sync/atomic"
 	"testing"
+	"time"
 
+	"pkg.world.dev/world-engine/assert"
 	"pkg.world.dev/world-engine/cardinal"
+	"pkg.world.dev/world-engine/cardinal/testutils"
 	"pkg.world.dev/world-engine/cardinal/types/engine"
 
 	"github.com/gorilla/websocket"
-	"pkg.world.dev/world-engine/assert"
-	"pkg.world.dev/world-engine/cardinal/testutils"
 )
 
 func wsURL(addr, path string) string {
@@ -27,7 +27,7 @@ type Event struct {
 func TestEvents(t *testing.T) {
 	// broadcast 5 messages to 5 clients means 25 messages received.
 	numberToTest := 5
-	tf := testutils.NewTestFixture(t, nil, cardinal.WithDisableSignatureVerification())
+	tf := testutils.NewTestFixtureWithPort(t, nil, "1337", cardinal.WithDisableSignatureVerification())
 	addr := tf.BaseURL
 	tf.StartWorld()
 	url := wsURL(addr, "events")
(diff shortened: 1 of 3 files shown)


cardinal/events/events_test.go
```diff
@@ -28,7 +27,7 @@ type Event struct {
 func TestEvents(t *testing.T) {
 	// broadcast 5 messages to 5 clients means 25 messages received.
 	numberToTest := 5
-	tf := testutils.NewTestFixture(t, nil, cardinal.WithDisableSignatureVerification())
+	tf := testutils.NewTestFixtureWithPort(t, nil, "1337", cardinal.WithDisableSignatureVerification())
```
jerargus, line 30: I made a ticket for myself to improve how we do port selection: https://linear.app/arguslabs/issue/WORLD-960/update-test-fixture-code-to-make-port-selection-more-robust

jerargus (approved):

smsunarto (approved):

PR #680, opened 2024-03-14
chore: fix issue with 'make rollup' leaving celestia-devnet container running

heronimus:
Closes: WORLD-880

## Overview

- On `make rollup` action, the `celestia_devnet` container leaves running after the interrupt action (ctrl+c) because it started on a separate action (`scripts/start-celestia-devnet.sh`) using `--detach` mode.

## Brief Changelog

- makefiles: use bash `trap` command, to catch exit and run `docker compose down chain celestia-devnet` to properly remove the `celestia_devnet` container & compose networks.

## Testing and Verifying

- run: `make rollup`
- interrupt & exit the running process with ctrl+c
- check `docker ps` and verify there's no `celestia_devnet` container running

image
^^ last log shows it properly remove both evm_base_shard & celestia_devnet container and compose networks.


makefiles/build.mk
```diff
@@ -11,6 +13,7 @@ start-evm:
 
 rollup:
 	@. ${CURDIR}/evm/scripts/start-celestia-devnet.sh && \
+	trap 'docker compose down chain celestia-devnet' EXIT; \
```
jerargus, line 16: Running your test instructions didn't seem to work.

I see the error `unknown command "chain" for "docker compose down"`. I think the `docker-compose down` doesn't actually take the list of containers... which seems at odds with the documentation here: https://docs.docker.com/reference/cli/docker/compose/down/. 

I'm on `Docker Compose version v2.17.3`

I think you either want to 1) use `docker compose down` to stop all containers or 2) use `docker compose stop chain celestia-devnet`

heronimus: @jerargus Yes, I just found out that the feature to selectively 'compose down' a service was only introduced after Docker v2.19.x (https://github.com/docker/compose/pull/10552). Currently, I'm running v2.24.x on my machine. 

I think it's fine to just tear down all the services with `docker compose down` for broader compatibility with older Docker versions, it will have same behaviour since the container running on `make rollup` is only chain & celestia-devnet anyway. Also, I'm avoiding `docker compose stop` as it still leaves the container there in a "stopped" status.

heronimus: I've just push the update. Thanks for pointing it out!

technicallyty (approved):
LGTM - tested locally.

makefiles/build.mk
jerargus: 👍 I'll update docker on my end as well.

jerargus (approved):

PR #696, opened 2024-03-20
feat(cardinal): use uuid instead of ints for entity id. DRAFT proposal 

pyrofolium:
Closes: WORLD-798

Replaces entity ids with uuids. 

The problem is in `TestFinalizeTickPerformanceIsConsistent`

Performance profile will change. ints are just that much more performant. Please write your opinion here. 

Note this is not ready for merging. This draft is to demonstrate the consequences of a uuid change on the test mentioned above. 

link to relevant slack conversation with Jeremy's requested distribution graphs:

https://argus-labs.slack.com/archives/C03G859K5TQ/p1710962088523289

@jerargus please list your thoughts here as well ^^


pyrofolium:
image
Above is the histogram of memory allocation when running `createAndFinalizeEntities` 100 times in `TestFinalizeTickPerformanceIsConsistent`

image
Above is the histogram for the time duration when running `createAndFinalizeEntities` 100 times in `TestFinalizeTickPerformanceIsConsistent`

Basically the data aligns with the intuition and logic. Strings have more varied allocation profiles and the duration is slightly slower as a result. My recommendation and vote is to just move forward with it as the main cost is the duration 

@jerargus and anyone else please comment with

pyrofolium:
> i say we punt this

Yeah I'm fine with this.

jerargus:
I'm fine with skipping this for now. If/when we do implement it we should keep that "TestFinalizeTickPerformanceIsConsistent" test, but with relaxed performance requirements to reflect the variable nature of string allocations. 

I think the initial reason for the test was because a bug in our storage code was never actually freeing up resources between ticks, and we were not only saving newly created entities, but also every single  entity that was ever created since the start of the cardinal process. This concern doesn't go away with the switch from int ids to string ids.

pyrofolium:
> I'm fine with skipping this for now. If/when we do implement it we should keep that "TestFinalizeTickPerformanceIsConsistent" test, but with relaxed performance requirements to reflect the variable nature of string allocations.
> 
> I think the initial reason for the test was because a bug in our storage code was never actually freeing up resources between ticks, and we were not only saving newly created entities, but also every single entity that was ever created since the start of the cardinal process. This concern doesn't go away with the switch from int ids to string ids.

I'm ok with this too.

pyrofolium:
Looks like everyone is in agreement to skip this.

PR #697, opened 2024-03-20
fix(cardinal): messages from different groups can now have same name

technicallyty:
## Overview

fixed an issue where messages could not be registered if with the same name when they were from different groups.

## Brief Changelog

- added godoc to RegisterMessage
- added FullName to message interface, which is a combination of <group>.<name>
- manager now registers messages based on their full name.
- changed method names to reflect the above
- added some docs explaining groups, fully qualified names, etc


cardinal/cardinal.go
```diff
@@ -100,6 +100,9 @@ func EachMessage[In any, Out any](wCtx engine.Context, fn func(message.TxData[In
 	return nil
 }
 
+// RegisterMessage registers a message to the world. Cardinal will automatically set up HTTP routes that map to each
+// registered message. Message URLs are take the form of "group/name". A default group, "game", is used
```
smsunarto, line 104: We would want to make sure to restrict the name of the group and name to alphanumeric and dash ( - )  to make sure there are no issues with this URL registrations

cardinal/message/manager.go
```diff
@@ -24,20 +25,20 @@ func NewManager() *Manager {
 }
 
 func (m *Manager) RegisterMessage(msgType types.Message, msgReflectType reflect.Type) error {
-	name := msgType.Name()
```
smsunarto, line 27: What I think we should do here is make msgType.ID() be set to what you are currently defining as msgType.FullName(). Since FullName() is guaranteed to be unique, it's a valid substitute for ID() and is actually better we can also remove ID() altogether, therefore resolving the TODO I left below on msgType.SetID().

cardinal/message/manager.go
```diff
@@ -64,9 +65,9 @@ func (m *Manager) GetMessageByID(id types.MessageID) types.Message {
 	return nil
 }
 
-// GetMessageByName returns the message with the given name, if it exists.
-func (m *Manager) GetMessageByName(name string) (types.Message, bool) {
```
smsunarto, line 68: I would just make this `GetMessage(id MessageID)`

cardinal/message/manager.go
```diff
@@ -75,11 +76,11 @@ func (m *Manager) GetMessageByType(mType reflect.Type) (types.Message, bool) {
 	return msg, ok
 }
 
-// isMessageNameUnique checks if the message name already exist in messages map.
-func (m *Manager) isMessageNameUnique(msgName string) error {
```
smsunarto, line 79: Then, this would be `isMessageIDUnique`

docs/cardinal/shard/evm-to-cardinal.mdx
```diff
@@ -66,14 +66,14 @@ interface IRouter {
...
 
-The `sendMessage` method enables smart contracts to send messages to the game shard specified by the given namespace.
+The `sendMessage` method enables smart contracts to send messages to the game shard specified by the given namespace. The messageID must be the fully qualified name of the message. For Cardinal game shards, the fully qualified name is of the form "groupName.MessageName". For example, the `create-persona` message's fully qualified name is `persona.create-persona`. Messages without a specified custom group in Cardinal have a default group called "game".
 
 #### Parameters
 
 | Parameter   | Type     | Description                                                                                |
 |-------------|----------|--------------------------------------------------------------------------------------------|
 | `message`   | `bytes`  | ABI encoded message struct.                                                                |
-| `messageID` | `string` | Message identifier. Usually the name passed into Cardinal's `NewMessageTypeWithEVMSupport` |
```
smsunarto, line 76: We can do this as a separate PR, but I prefer if the function signature is like this instead.

`func sendMessage(message bytes, messageGroup string, messageName string, namespace string)`

Then Cardinal would be the one that turns group + name into the ID.

This is better IMO because 1) user doesn't need to know how to format the ID correctly, 2) less odds of fat finger, 3) if for whatever reason we want to change how we encode the ID internally within Cardinal, they don't have to switch the messageID manually.

The downside of this is that we are imposing an opinion on what the message identifier schema is. In a scenario where let's say we have a Cardinal alternative implementation that doesn't use groups, then it would be awkward. If we wanted to add version as a part of the identifier, then this precompile function signature would have to be updated too.

Curious to hear your thoughts

smsunarto (changes requested):

docs/cardinal/shard/evm-to-cardinal.mdx
technicallyty: i dont think the user should have to know the ID at all, tbh. nor should they have to worry about fat fingering keys.

the ideal usage of router is with helper smart contracts like this:

```solidity
contract DarkFrontierHelper {
    string sendEnergyID = "game.send-energy"
    struct SendEnergyInput {
        string To;
        string From;
        string PlanetID;
    }


    function SendEnergy(SendEnergyInput input, string namespace) {
        bz = abi.encode(input)
        router.sendMessage(bz, sendEnergyID, namespace)
    }
}
```

i would like to strive to keep router as slim and unopinionated as possible. this does have the downside that smart contracts would have to update if we want to change the schema, but i'm fine with the tradeoff if we get to keep router slim.

cardinal/cardinal.go
technicallyty: added alphanumeric with dash/underscores enforcement and added test cases. [enforce validity on message name and group, clean up message test file](https://github.com/Argus-Labs/world-engine/pull/697/commits/e98585e404df1a7be2a0be0382223ece33afe38b)

also cleaned up some stuff in that commit.

cardinal/message/manager.go
technicallyty: eh, would prefer to make the removal of ID a linear ticket. that has some cascading changes across tests and would prefer to keep this PR focused

cardinal/message/manager.go
technicallyty: i put the group / name validation in NewMessageType. i want the error to happen right away if they construct a message with a bad group and/or name. the isMessageNameUnique is more of a stateful check, whereas group/name validation is a stateless check.

smsunarto (approved):

cardinal/types/message.go
```diff
@@ -5,6 +5,9 @@ type Message interface {
 	Name() string
 	// Group returns the group of the message.
 	Group() string
+	// FullName returns the fully qualified name of the message. It is formed by <group_name>.<message_name>.
+	// example: persona.create-persona
+	FullName() string
```
jerargus, line 10: I've added a triage ticket to collapse FullName and ID into the same thing: https://linear.app/arguslabs/issue/WORLD-980/in-cardinaltypesmessagego-id-and-fullname-should-be-combined

jerargus (approved):

cardinal/types/message.go
technicallyty: thanks!

PR #698, opened 2024-03-20
ci: add swagger check

heronimus:
Closes: WORLD-925

## Overview

Do Swagger specs check to make sure it doesn't change.

## Brief Changelog

- Add `make swagger-check`: perform a diff check between the existing and newly generated swagger specs from the latest changes.
- Add `make swaggo-install`: install swag-cli to generate swagger specs files.
- Add CI action to execute swagger-check on PR event. 
- Fix swaggo annotation issue on function with multiple API operations (`cardinal/server/handler/tx.go`, `PostTransaction`) 
--> Swaggo supports multiple `@Router` on a single function, but multiple defined annotations like `@Summary`, and `@Accept`,`@Produce` with same value causing a double and overridden and invalid swagger specification format. 
(I put the details in the review comments below https://github.com/Argus-Labs/world-engine/pull/698#discussion_r1532703826)

## Testing and Verifying

- Run `make swagger-check` locally, it will create temporary generated swagger files at `.tmp/swagger` directory, then do a comparison vs existing swagger spec files on `cardinal/server/docs/swagger.json`.
- If no changes are detected, the command will exit and print `swagger-diff: no changes detected`.
- If changes are detected, it will return an error message indicating the difference.

--> Example when an endpoint is missing & request params is different:

image

### DIFF
--- a/cardinal/server/handler/tx.go
+++ b/cardinal/server/handler/tx.go
@@ -31,24 +31,21 @@ type Transaction = sign.Transaction
 // PostTransaction godoc
 //
 //	@Summary		Submit a transaction to Cardinal
-//	@Description	Submit a transaction to Cardinal
+//	@Description	Submit a transaction to Cardinal / Create a Persona transaction to Cardinal
 //	@Accept			application/json
 //	@Produce		application/json
-//	@Param			txType	path		string	true	"label of the transaction that wants to be submitted"
-//	@Success		200		{object}	PostTransactionResponse
-//	@Failure		400		{string}	string	"Invalid transaction request"
+//	@Success		200	{object}	PostTransactionResponse
+//	@Failure		400	{string}	string	"Invalid transaction request"
+//
+// PostTransaction with label godoc
+//
 //	@Router			/tx/game/{txType} [post]
+//	@Param			txType	path	string	true	"label of the transaction that wants to be submitted"
 //
 // PostTransaction with Persona godoc
 //
-//	@Summary		Create a Persona transaction to Cardinal
-//	@Description	Create a Persona transaction to Cardinal
-//	@Accept			application/json
-//	@Produce		application/json
-//	@Param			txBody	body		Transaction	true	"Transaction details"
-//	@Success		200		{object}	PostTransactionResponse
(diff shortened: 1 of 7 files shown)


cardinal/server/handler/tx.go
```diff
@@ -31,24 +31,21 @@ type Transaction = sign.Transaction
...
-//	@Summary		Create a Persona transaction to Cardinal
-//	@Description	Create a Persona transaction to Cardinal
-//	@Accept			application/json
-//	@Produce		application/json
-//	@Param			txBody	body		Transaction	true	"Transaction details"
-//	@Success		200		{object}	PostTransactionResponse
-//	@Failure		400		{string}	string	"Invalid transaction request"
 //	@Router			/tx/persona/create-persona [post]
+//	@Param			txBody	body	Transaction	true	"Transaction details"
 func PostTransaction(
```
heronimus, line 49: FYI: I made these changes here to fix issue this swaggo issue:

Swaggo supports multiple `@Router` on a single function like the `PostTransaction` function here 
but multiple defined annotations like `@Summary`, and `@Accept`, `@Produce` with same value causing a double, overridden, and invalid swagger specification format.

- The swagger-diff checker complains about these duplicated array values:
```
swagger.json is not a valid Swagger specification:
- The property '#/paths//tx/game/{txType}/post/produces' contained duplicated array values
- The property '#/paths//tx/game/{txType}/post/consumes' contained duplicated array values
```
image

- `@Sumarry` for `/tx/game/{txType}` router getting overridden by `/tx/persona/create-persona` 
example when importing it in Postman:
image


The proposed fix:
- remove duplicate `@Accept` `@Produce` annotation from `PostTransaction` function
- use generic `@Summary` values that can describe `/tx/game/{txType}`  and `/tx/persona/create-persona`, so it less confusing than being `/tx/game` described as create-persona.

**There's an alternative workaround by creating an additional function wrapper, but I don't think that necessary for this swagger use case --> https://github.com/swaggo/swag/issues/1095#issuecomment-1489594724

technicallyty (approved):
LGTM

namrapatel (approved):

PR #699, opened 2024-03-21
fix: nakama verifier reads only create persona results 

namrapatel:
Closes: WORLD-836

Fixes the issue where the Nakama Verifier was reading all receipts as CreatePersonaResults by reading the `success` key in the Result struct by adding a `type` key to the Result struct that can be used for all Cardinal Internal receipts.

e2e tests broken, debugging...


technicallyty:
closing as stale

PR #702, opened 2024-03-21
feat(cardinal): add property filtering capabilities to search.

pyrofolium:
Closes: WORLD-617

this allows users to use search like this with chained filters:

```go
amount, err = cardinal.NewSearch(worldCtx, filter.Exact(Alpha{})).
		FilterSelect(func(id types.EntityID) bool {
			comp, err := cardinal.GetComponent[Alpha](worldCtx, id)
			assert.NilError(t, err)
			return comp.Name1 == "BLAH"
		}).Count()
```

FilterSelects can be chained on top of a Search object an unlimited amount of times. This brings the search api more inline with functional programming paradigms where we have map, reduce, filter, and foreach. With golang, only filter and foreach are possible as chainable methods. Map/Reduce will require functions as they are generic. 

The secondary purpose for adding this feature is to pave the way for filters to be a compile target for High level CQL syntax that allows filtering components based off of properties. the new filterFuncs property contains target functions for compilation. While this CQL syntax doesn't exist yet, it could in the future. 

Typically this method `FilterSelect` is just called `Filter` in almost all programming languages and frameworks. However our ECS concept of `filters` collides with this name so I chose the term: `FilterSelect`. 

changes:
- added filterFuncs property to *Search
- modified Count() to take into account filterFuncs
- modified Each to take into account filterFuncs
- modified First to take into account filterFuncs
- added new FilterSelect method

### DIFF
--- a/cardinal/search/search.go
+++ b/cardinal/search/search.go
@@ -45,7 +45,7 @@ func NewSearch(wCtx engine.Context, filter filter.ComponentFilter) *Search {
 	}
 }
 
-func (s *Search) Filter(callback func(id types.EntityID) bool) *Search {
+func (s *Search) FilterSelect(callback func(id types.EntityID) bool) *Search {
 	s.filterFuncs = append(s.filterFuncs, callback)
 	return s
 }
--- a/cardinal/search/search.go
+++ b/cardinal/search/search.go
@@ -29,6 +29,7 @@ type Search struct {
 	namespace   string
 	reader      gamestate.Reader
 	wCtx        engine.Context
+	filterFuncs []func(id types.EntityID) bool
 }
 
 // NewSearch creates a new search.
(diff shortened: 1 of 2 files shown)


smsunarto:
looks cool, I wonder if we should call it `Where(...)` instead

```
amount, err = cardinal.NewSearch(worldCtx, filter.Exact(Alpha{})).
		Where(func(id types.EntityID) bool {
			comp, err := cardinal.GetComponent[Alpha](worldCtx, id)
			assert.NilError(t, err)
			return comp.Name1 == "BLAH"
		}).Count()
```

pyrofolium:
> looks cool, I wonder if we should call it `Where(...)` instead
> 
> ```go
> amount, err = cardinal.NewSearch(worldCtx, filter.Exact(Alpha{})).
> 		Where(func(id types.EntityID) bool {
> 			comp, err := cardinal.GetComponent[Alpha](worldCtx, id)
> 			assert.NilError(t, err)
> 			return comp.Name1 == "BLAH"
> 		}).Count()
> ```

Yeah that's probably better. I'll change it.

pyrofolium:
I talked with Scott about this. Looks like he wants a new design. I'll wait on his next spec. For this one if there's a lot of resistance or requested changes it to it I'll just close this PR otherwise it will go up. Main reason is because Search will be replaced by Scotts new design so there's no point in doing too many changes if it's going to be replaced anyway.

jerargus (approved):
Nice. I think this is a sensible addition to our search pattern.

smsunarto (approved):

PR #703, opened 2024-03-21
refactor: use tokens + custom interceptors instead of SSL

technicallyty:
## Overview

removes usage of SSL in favor of simple tokens. tokens can be generated like so: `openssl rand -hex 32`. they're just strings though, so any arbitrary token can be used. but using world-cli, we can enforce the usage of the above openssl command.

## Brief Changelog

- add RouterKey to config, required in prod mode
- added interceptors to all router clients/servers to inject/read keys before running the invoker/handler.

## Testing and Verifying

ran e2e-evm tests with and without router keys and saw the expected behavior.


evm/router/router.go
```diff
@@ -228,30 +223,22 @@ func (r *routerImpl) getConnectionForNamespace(ns string) (routerv1.MsgClient, e
...
-		RootCAs:    certPool,
-		MinVersion: tls.VersionTLS12,
-	}
+// intercepts grpc invocations and injects the router-routerKey.
+func (r *routerImpl) clientCallInterceptor(
+	ctx context.Context, method string, req, reply any, cc *grpc.ClientConn,
+	invoker grpc.UnaryInvoker, opts ...grpc.CallOption,
+) error {
+	md := metadata.New(map[string]string{"router-routerKey": r.routerKey})
+	ctx = metadata.NewOutgoingContext(ctx, md)
```
jerargus, line 241: This may be overly pedantic, but it would probably be safer to use:
```
ctx = metadata.AppendToOutgoingContext(ctx, "router-routerKey", r.routerKey)
```
To make sure you're not overwriting any existing metadata.

evm/sequencer/server.go
```diff
@@ -135,3 +111,27 @@ func (s *Sequencer) RegisterGameShard(
...
+	req any,
+	_ *grpc.UnaryServerInfo,
+	handler grpc.UnaryHandler,
+) (resp any, err error) {
+	md, ok := metadata.FromIncomingContext(ctx)
+	if !ok {
+		return nil, status.Errorf(codes.Unauthenticated, "missing metadata")
+	}
+
+	routerKey, ok := md["router-routerKey"]
```
jerargus, line 127: What about putting this "router-routerKey" as a constant in the evm/utils package that both the sequencer and router packages import?

evm/sequencer/server.go
```diff
@@ -135,3 +111,27 @@ func (s *Sequencer) RegisterGameShard(
...
+	if !ok || len(routerKey) == 0 {
```
jerargus, line 128: nit: the ok check is redundant here. If the key is missing, len(routerKey) will be 0.

jerargus (approved):
Left some optional changes.

cardinal/router/router.go
```diff
@@ -60,16 +61,17 @@ type router struct {
...
+func New(namespace, sequencerAddr, baseShardQueryAddr, routerKey string, provider Provider) (Router, error) {
+	rtr := &router{namespace: namespace, port: defaultPort, provider: provider, routerKey: routerKey}
 
-func New(namespace, sequencerAddr, baseShardQueryAddr string, provider Provider) (Router, error) {
-	rtr := &router{namespace: namespace, port: defaultPort, provider: provider}
-
-	conn, err := grpc.Dial(sequencerAddr, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	conn, err := grpc.Dial(
+		sequencerAddr,
+		grpc.WithUnaryInterceptor(rtr.clientCallInterceptor),
```
smsunarto, line 72: Use `grpc.WithPerRPCCredentials` instead. 

As mentioned above, we should have a package that is both used by EVM and Cardinal with a router key struct type that implements the `grpc.PerRPCCredentials` interface (https://pkg.go.dev/google.golang.org/grpc/credentials#PerRPCCredentials).

This avoids hardcoding the metadata key individually that might become a footgun when we need to make a change to that in the fture.

Example: https://github.com/kelseyhightower/grpc-hello-service/blob/master/credentials/jwt/jwt.go

cardinal/router/router.go
```diff
@@ -146,3 +152,14 @@ func (r *router) Start() error {
 	r.serverAddr = listener.Addr().String()
 	return nil
 }
+
+// intercepts grpc invocations and injects the router-key.
+func (r *router) clientCallInterceptor(
```
smsunarto, line 157: This can be completely removed if you use the grpc.WithPerRPCCredentials option mentioned above.

cardinal/router/router_evm_server.go
```diff
@@ -30,13 +33,44 @@ type evmServer struct {
...
+	ctx context.Context,
+	req any,
+	_ *grpc.UnaryServerInfo,
+	handler grpc.UnaryHandler,
+) (resp any, err error) {
+	// we only want to guard the SendMessage method. not the query shard method.
+	if _, ok := req.(*routerv1.SendMessageRequest); !ok {
+		return handler(ctx, req)
+	}
+	md, ok := metadata.FromIncomingContext(ctx)
```
smsunarto, line 59: You can use the router key package/type mentioned above to add `jwt.FromIncomingContext` to obtain the token from the request context without having to hardcode "router-key" here.

evm/router/router.go
```diff
@@ -228,30 +223,22 @@ func (r *routerImpl) getConnectionForNamespace(ns string) (routerv1.MsgClient, e
 	addr := res.Address
 	conn, err := grpc.Dial(
 		addr,
-		grpc.WithTransportCredentials(r.creds),
+		grpc.WithUnaryInterceptor(r.clientCallInterceptor),
```
smsunarto, line 226: can also use grpc.WithPerRPCCredentials here too. If you want it to only be injected for a specific rpc endpoint, you can inject `grpc.PerRPCCredentials` as the call option for that rpc endpoint.

cardinal/config.go
```diff
@@ -42,6 +42,8 @@ type WorldConfig struct {
 	CardinalLogLevel          string  `config:"CARDINAL_LOG_LEVEL"`
 	StatsdAddress             string  `config:"STATSD_ADDRESS"`
 	TraceAddress              string  `config:"TRACE_ADDRESS"`
+	// RouterKey is a JWT token used to secure communications between the game shard and the base shard.
```
smsunarto, line 45: This is technically not a JWT token, a JWT token would need to be signed by the server during an authentication process. This is more like a normal API key, which is fine, but the comment needs to be corrected.

smsunarto (changes requested):
Also nothing here that we should figure out a way to make this process of generating a JWT token and injecting it seamless via World CLI.

evm/router/router.go
technicallyty: whoops, that darn refactor tool!

technicallyty: will address that as well

cardinal/router/router.go
```diff
@@ -126,6 +128,10 @@ func (r *router) SubmitTxBlob(
 	return eris.Wrap(err, "")
 }
 
+func (r *router) TransactionIterator() iterator.Iterator {
```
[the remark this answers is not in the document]
technicallyty: unrelated to or, its just relocated method

evm/router/router.go
technicallyty: client call interceptors have been removed for PerRPCCredential option.

evm/sequencer/server.go
technicallyty: done. placed it in the rift package.

evm/sequencer/server.go
technicallyty: addressed

cardinal/config.go
```diff
@@ -58,6 +60,9 @@ func (w WorldConfig) Validate() error {
 		return eris.New("must supply BASE_SHARD_SEQUENCER_ADDRESS and BASE_SHARD_QUERY_ADDRESS for production " +
 			"mode Cardinal worlds")
 	}
+	if w.RouterKey == "" {
+		return eris.New("must supply ROUTER_KEY for production mode Cardinal worlds")
```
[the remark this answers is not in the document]
technicallyty: what do you suggest? i wasnt sure if i wanted to enforce any single format so people could generate any type of key they want.

cardinal/router/router.go
technicallyty: done

cardinal/config.go
technicallyty: removed `JWT` from the comment

cardinal/config.go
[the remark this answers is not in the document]
technicallyty: optimism uses length 32 openssl keys: https://docs.optimism.io/builders/node-operators/tutorials/testnet#create-a-jwt-secret

and it looks like thats what zul is doing in the CLI PR. should we just enforce a length 32 key of only alphanumerics?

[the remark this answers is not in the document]
smsunarto: sure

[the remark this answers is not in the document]
technicallyty: done [validation of key](https://github.com/Argus-Labs/world-engine/pull/703/commits/d91e18a7f62157541934de0a228381de95618a19)

technicallyty:
fixing some tests

technicallyty:
done

evm/app/plugins.go
```diff
@@ -1,35 +1,31 @@
...
-	certPath := os.Getenv("SERVER_CERT_PATH")
-	keyPath := os.Getenv("SERVER_KEY_PATH")
-	if certPath == "" || keyPath == "" {
-		logger.Info("running shard sequencer without SSL certs")
-		app.ShardSequencer = sequencer.NewShardSequencer()
+	routerKey := os.Getenv("ROUTER_KEY")
+	var sequencerOpts []sequencer.Option
+	var routerOpts []router.Option
+	if routerKey == "" {
+		app.Logger().Debug("WARNING: starting the EVM base shard in insecure mode. No ROUTER_KEY provided")
```
jerargus, line 19: Why not emit this at the warning level, instead of prefixing this "WARNING" string?

jerargus, line 19: Is this a zerolog? I think you need to end the line with "Msg" to actually emit the log.

jerargus, line 19: The previous version of this code used log.Logger instaed of app.Logger(). Any particular reason you're switching?

jerargus (approved):

evm/app/plugins.go
technicallyty: cosmos SDK for some reason doesn't have a warning level lol. its just INFO, ERROR, or DEBUG. 🤷🏻

technicallyty: it's cosmossdk.io/log

technicallyty: just better to use the cosmos sdk logger so all the logs show up as expected and in the same format

cardinal/router/router_evm_server.go
smsunarto: I don't think this is resolved yet. I don't think this is resolved yet.

You can just add `FromIncomingContext` to the credentials package so you don't have to do `routerKey[md.credentials.TokenKey]`

rift/credentials/creds.go
```diff
@@ -0,0 +1,44 @@
...
+
+var (
+	TokenKey = "router_key"
+
+	_ credentials.PerRPCCredentials = &simpleTokenCredential{}
+
+	routerKeyRegexp = regexp.MustCompile(`^[a-zA-Z0-9]{64}$`)
+)
+
+type simpleTokenCredential struct {
```
smsunarto, line 19: nit: just call it `tokenCredential`, I don't think "simple" really gives us anything here.

smsunarto (changes requested):

cardinal/router/router_evm_server.go
technicallyty: just to clarify, you want the credentials package to extract the key from the ctx and return it to the `serverCallInterceptor` method?

rift/credentials/creds.go
technicallyty: fixed

smsunarto (approved):

PR #705, opened 2024-03-22
feat(nakama): Add SIWE login method

jerargus:
Closes: WORLD-905

## Overview

Enable SIWE authentication for Nakama. See the [design doc](https://www.notion.so/arguslabs/Ethereum-Login-for-Nakama-c453e4ac301c41cba81cc2c8dedeadb0) for more context on how this is implemented.

Since SIWE isn't directly supported by Nakama, we use the AuthenticateCustom handler. This handler can be expanded in the future to support other login methods. The vars["type"] field specifies what login method to use.

The majority of the new code is in relay/nakama/siwe/siwe.go. The two main function of that package are to:
1. Generate a new SIWE message that can later be signed by a client and 
2. Validate a given SIWE message and signature. 

## Brief Changelog

- Move the AdminAccountID constant to the utils package
- Expand the e2e test client to also support SIWE
- Update the fake nakama storage helper to reject Value fields that are not JSON serialized (this matches the behavior of the runtime.NakamaModule).
- Update the fake nakama storage helper to better align with how Nakama handles storage object versions.

## Testing and Verifying

- Unit tests added for the siwe package
- And end to end test added to make sure SIWE can be used to authenticate a user.


jerargus:
I misjudged how difficult it would be to test the custom auth handlers. I'm going to split the HandleSIWE handler into two methods.

relay/nakama/custom_auth.go
smsunarto, line 1: It's probably cleaner to put this in a separate `auth` package where each auth handler (i.e. `handleSIWE`) have its own file.

relay/nakama/custom_auth.go
```diff
@@ -0,0 +1,98 @@
...
+	signInWithEthereumType = "siwe"
+)
+
+var (
+	ErrBadCustomAuthType = errors.New("bad custom auth type")
+)
+
+func initCustomAuthentication(initializer runtime.Initializer) error {
+	if err := initializer.RegisterBeforeAuthenticateCustom(handleCustomAuthentication); err != nil {
+		return eris.Wrap(err, "failed to init siwe authentication")
```
smsunarto, line 28: nit: while this is true for now, when we start adding more custom authentication this error message is no longer valid. It's better for this to be `failed to register custom authentication handlers` the `failed to init siwe authentication` can then be bubbled up from the `if authType == signInWithEthereumType` block by wrapping the error from handler.

relay/nakama/custom_auth.go
```diff
@@ -0,0 +1,98 @@
...
+	return nil
+}
+
+func handleCustomAuthentication(
+	ctx context.Context,
+	logger runtime.Logger,
+	_ *sql.DB,
+	nk runtime.NakamaModule,
+	in *api.AuthenticateCustomRequest) (*api.AuthenticateCustomRequest, error) {
+	authType := in.GetAccount().GetVars()["type"]
```
smsunarto, line 39: it's probably cleaner to have this be a utils function that also validates the authType and return an error if it's not a valid auth type.

relay/nakama/custom_auth.go
```diff
@@ -0,0 +1,98 @@
...
+
+func handleSIWE(
+	ctx context.Context,
+	logger runtime.Logger,
+	nk runtime.NakamaModule,
+	in *api.AuthenticateCustomRequest,
+) (
+	*api.AuthenticateCustomRequest, error,
+) {
+	signerAddress := in.GetAccount().GetId()
```
smsunarto, line 55: should this be extracted to a `validateAuthRequest`?

relay/nakama/custom_auth.go
```diff
@@ -0,0 +1,98 @@
...
+	signature := in.GetAccount().GetVars()["signature"]
+	message := in.GetAccount().GetVars()["message"]
+
+	if signerAddress == "" {
```
smsunarto, line 59: should probably just check the fields individually then errors.Join(...) it

relay/nakama/custom_auth.go
```diff
@@ -0,0 +1,98 @@
...
+			logger, siwe.ErrMissingSignature, codes.InvalidArgument, "missing field")
+		return nil, err
+	}
+	if signature != "" && message == "" {
+		_, err := utils.LogErrorWithMessageAndCode(
+			logger, siwe.ErrMissingMessage, codes.InvalidArgument, "missing field")
+		return nil, err
+	}
+
+	if signature != "" && message != "" {
```
smsunarto, line 75: trust your validator

relay/nakama/custom_auth.go
```diff
@@ -0,0 +1,98 @@
...
+		// The user has provided a signature and a message. Attempt to authenticate the user.
+		if err := siwe.ValidateSignature(ctx, nk, signerAddress, message, signature); err != nil {
+			_, err = utils.LogErrorWithMessageAndCode(
+				logger, siwe.ErrMissingMessage, codes.Unauthenticated, "authentication failed")
+			return nil, err
+		}
+		return in, nil
+	}
+
+	// The signature and message is empty. Generate a new SIWE message for the user.
```
smsunarto, line 85: This might be a personal preference, but I typically think error states like this should be handled first and terminates the function earlier before going with the happy path

relay/nakama/siwe/siwe.go
```diff
@@ -0,0 +1,276 @@
...
+// GenerateResult will be serialized to JSON and sent back to a client. The client must sign the message in the
+// SIWEMessage field, and resubmit their authentication request.
+type GenerateResult struct {
+	SIWEMessage string `json:"siwe_message"`
+}
+
+// nonceStorageObj is the struct that is serizlied to JSON and saved to the Nakama storage layer. Each signer address
+// maps to a different nonceStorageObj. A signature is valid if the nonce contained in the signed message does NOT
+// already have an entry in this object.
+type nonceStorageObj struct {
```
smsunarto, line 51: on a separate note, we should probably consolidate all Nakama storage stuff into a single `storage` package

relay/nakama/siwe/siwe.go
```diff
@@ -0,0 +1,276 @@
...
+func makeOptions() map[string]any {
+	now := time.Now().UTC()
+	return map[string]any{
+		"issuedAt":       now.Format(time.RFC3339),
+		"expirationTime": now.Add(DefaultTimeout).Format(time.RFC3339),
+		"statement":      DefaultStatement,
+	}
+}
+
+func makeNonce() (string, error) {
```
smsunarto, line 270: nit: this should be `generateNonce` so as to not be confused with a regular constructor

relay/nakama/siwe/siwe.go
```diff
@@ -0,0 +1,276 @@
...
+
+	var nonceObj nonceStorageObj
+	if err := json.Unmarshal([]byte(objs[0].GetValue()), &nonceObj); err != nil {
+		return nil, eris.Wrap(err, "failed to unmarshal nonces")
+	}
+	nonceObj.Version = objs[0].GetVersion()
+	return &nonceObj, nil
+}
+
+func makeOptions() map[string]any {
```
smsunarto, line 261: nit: seems like people like to use `new<...>` prefix for constructor

relay/nakama/siwe/siwe.go
smsunarto: i would also note here that this is not to be confused with the cardinal tx nonce

relay/nakama/siwe/siwe.go
```diff
@@ -0,0 +1,276 @@
...
+	issuedAt, err := time.Parse(time.RFC3339, msg.GetIssuedAt())
+	if err != nil {
+		return nil, eris.New("failed to parse issued at time")
+	}
+	if time.Now().Before(issuedAt) {
+		return nil, eris.New("message must be issued in the past")
+	}
+
+	delta := expTime.Sub(issuedAt)
+	if delta > DefaultTimeout {
```
smsunarto, line 128: why do we need this when we already have expiration time?

smsunarto (changes requested):

relay/nakama/custom_auth.go
jerargus: I'm not sure what you mean by cleaner here. Do you think the chaining of GetAccount().GetVars() isn't clear? 

Regarding validation, the validation code is essentially going to be a duplicate of the code below that identifies and calls the correct handler. The handler identification code will still need to end with a ErrBadCustomAuthType.

relay/nakama/siwe/siwe.go
jerargus: I added a triage ticket to consolidate the common storage object operations: https://linear.app/arguslabs/issue/WORLD-991/consolidate-nakama-storage-object-operations.

I'll add a note about the cardinal tx nonces.

relay/nakama/siwe/siwe.go
jerargus: This check means we're only going to accept SIWE messages that are valid for at most a 5 minute window. The messages we generate conform to this requirement, but there's nothing preventing a user from constructing their own SIWE Message with a much longer valid window. I want to avoid a situation where a user makes a SIWE Message that is valid from time 0 to the heat death of the universe.

relay/nakama/custom_auth.go
jerargus: done

relay/nakama/custom_auth.go
jerargus: I updated the comment to not include the siwe part, however we don't have the option of emitting a separate error for each custom authentication type. The RegisterBeforeAuthenticateCustom only supports registering a single function, so as we add on more custom authentication types, they will all have to go into the same handler.

relay/nakama/custom_auth.go
jerargus: I moved the SIWE validation into a dedicated validateSIWE helper. I'd rather not make a single validateAuthRequest that is meant to validate different custom auth types. I think the authentication for specific handlers should be called in the handlers themselves.

relay/nakama/custom_auth.go
jerargus: done

relay/nakama/custom_auth.go
jerargus: done

relay/nakama/custom_auth.go
jerargus: done. Error/siwe generation has been moved up. The last operation in this function is now the successful authentication of a user.

relay/nakama/siwe/siwe.go
jerargus: done

relay/nakama/siwe/siwe.go
jerargus: done

smsunarto (approved):

relay/nakama/siwe/siwe.go
```diff
@@ -0,0 +1,278 @@
...
+		Nonce:     msg.GetNonce(),
+		ExpiredAt: expTime,
+		IssuedAt:  issuedAt,
+	}, nil
+}
+
+// useNonce attempts to use the given nonce for the given signer address. If an identical nonce has been used, and
+// that nonce has not yet expired, useNonce will return false. A return value of (true, nil) means the nonce was
+// successfully used.
+func useNonce(
```
technicallyty, line 143: a little confused on the multiple nonces stuff. can you explain?

relay/nakama/siwe/siwe.go
```diff
@@ -0,0 +1,278 @@
...
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/heroiclabs/nakama-common/runtime"
+	"github.com/rotisserie/eris"
+	"github.com/spruceid/siwe-go"
+
+	"pkg.world.dev/world-engine/relay/nakama/utils"
+)
+
+const (
+	// TODO: This static domain and URI should be configurable via an environment variable.
```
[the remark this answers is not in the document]
jerargus: done

relay/nakama/siwe/siwe.go
jerargus: I tried to clarify this comment. If it's still not clear, let's jump on a call and talk about what's confusing. Then I'll take another swing and the phrasing.

technicallyty: that helps. all good here

PR #716, opened 2024-03-25
refactor(cardinal): move /query/debug/state from plugins to http handlers

jerargus:
Closes: WORLD-996

## Overview

Move the debug state handler out of plugins and back over to an HTTP handler.
Note, the cardinal/server/handler/debug.go was jus moved, and there are no actual logic changes.

The debug_test file was changed to 1) hit the HTTP endpoint instead of using the query directly and 2) use the structs that already exist in the test suite.

## Testing and Verifying

Existing tests were modified to hit the HTTP endpoint instead of using the query.


cardinal/server/handler/debug.go
```diff
@@ -0,0 +1,66 @@
...
+	ID         types.EntityID             `json:"id"`
+	Components map[string]json.RawMessage `json:"components" swaggertype:"object"`
+}
+
+type DebugStateResponse []*debugStateElement
+
+// GetDebugState godoc
+//
+//	@Summary		Get information on all entities and components in world-engine
+//	@Description	Displays the entire game state.
```
smsunarto, line 25: nit: formatting seems borked here

cardinal/server/handler/debug.go
```diff
@@ -0,0 +1,66 @@
...
+//	@Produce		application/json
+//	@Success		200	{object}	DebugStateResponse
```
smsunarto, line 27: nit: we should probably annotate possible internal server error (Error 500)

cardinal/server/handler/debug.go
```diff
@@ -0,0 +1,66 @@
...
+type debugStateElement struct {
+	ID         types.EntityID             `json:"id"`
+	Components map[string]json.RawMessage `json:"components" swaggertype:"object"`
+}
+
+type DebugStateResponse []*debugStateElement
+
+// GetDebugState godoc
+//
+//	@Summary		Get information on all entities and components in world-engine
```
smsunarto, line 24: nit: `Get all entity and component state in Cardinal`

cardinal/server/handler/debug.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+type DebugStateRequest struct{}
+
+type debugStateElement struct {
+	ID         types.EntityID             `json:"id"`
+	Components map[string]json.RawMessage `json:"components" swaggertype:"object"`
+}
+
+type DebugStateResponse []*debugStateElement
```
smsunarto, line 20: is there a reason why this is a pointer to struct? without pointer seems cleaner.

smsunarto (approved):
left some optional nits

cardinal/server/handler/debug.go
jerargus: done

cardinal/server/handler/debug.go
jerargus: done

cardinal/server/handler/debug.go
jerargus: No reason. It was copied from the plugin code. Changed to value.

technicallyty (approved):

PR #722, opened 2024-03-29
feat(cardinal): search v2

pyrofolium:
Closes: WORLD-1002

```
                                                                                                              
                                                                                                              
                                                                                     ┌─────────────────       
                                                                                     │                │       
                                                                                     ▼                │       
         ┌──────────────────┐              ┌─────────────────┐              ┌────────────────┐        │       
         │                  │              │                 │              │                │        │       
         │ NewSearch(...)   ├──────────────▶   Entity(...)   ├──────────────▶   Where(...)   │────────┘       
  ┌──────┴─────────┐        │    ┌─────────┴─────┐           │ ┌────────────┴──┐             │                
  │  Return Type:  ├────────┘    │ Return Type:  ├───────────┘ │ Return Type:  ├─────────────┘                
  │preSearchBuilder│             │ SearchBuilder │             │ SearchBuilder │     │                        
  └────────────────┘             └───────────────┘             └───────────────┘     │                        
                                                                       │             │                        
                                                                       │             └────┐                   
                                                                       │                  │                   
            ┌──────────────────────────────────────────────────────────┘                  │    ┌─────────────┐
            │                                                                             │    │   Methods   │
            │                                                                             ▼    │ return void │
            │                                                                  ┌───────────────┤ or integer  │
            ▼                                                                  │               └─────┬───────┘
     ┌──────────────┐              ┌───────────┐                               │     Each(...),      │        
     │              │              │           │                               │     First(...),     │        
     │              ◀──────────────┤           │          ┌───────────────────▶│   MustFirst(...),   │        
     │ Input type:  │              │  Return   │          │                    │     Count(...)      │        
     │SearchBuilder ├──────────────┤   Type:   │          │                    │                     │        
     │or Searchable │              │Searchable │          │                    └─────────────────────┘        
     │              │              │           │          │                                                   
     │              │              │           │          │                                                   
     └──────┬───────┘              └───┬───────┘          │                                                   
            │  And, Or, Not functions  │                  │                                                   
            │  that take search types  ├──────────────────┘                                                   
            │                          │                                                                      
            │                          │                                                                      
            │                          │                                                                      
            │                          │                                                                      
            └──────────────────────────┘                                                                      
```

There are 3 concepts needed to be understood. 

Set Operators - Composes search results via set operators, And, Or, Not
Filter Operators - Composes Contains and Exact filter operators via And, Or Not
Filter functions - functions that filter search results based on component properties. 

<!--
ELLIPSIS_HIDDEN
-->


----

| <a href="https://ellipsis.dev" target="_blank"><img src="https://avatars.githubusercontent.com/u/80834858?s=400&u=31e596315b0d8f7465b3ee670f25cea677299c96&v=4" alt="Ellipsis" width="30px" height="30px"/></a> | :rocket: This PR description was created by [Ellipsis](https://www.ellipsis.dev) for commit b56e8525076a2de66c63b9f651c2bae429ebbeea.  | 
|--------|--------|

### Summary:
This PR introduces a new, more flexible version of the search functionality in the Cardinal engine, including the concept of set operators and filter functions, and updates various files to use this new functionality.

**Key points**:
- Introduced a new version of the search functionality in the Cardinal engine.
- Added the concept of set operators (And, Or, Not) and filter functions for composing search results.
- Updated various files to use the new search functionality.
- Added tests for the new search functionality.


----
Generated with :heart: by [ellipsis.dev](https://www.ellipsis.dev)



<!--
ELLIPSIS_HIDDEN
-->


cardinal/benchmark/benchmark_test.go
```diff
@@ -36,7 +35,7 @@ func setupWorld(t testing.TB, numOfEntities int, enableHealthSystem bool) *testu
 		err := cardinal.RegisterSystems(
 			world,
 			func(wCtx engine.Context) error {
-				q := cardinal.NewSearch(wCtx, filter.Contains(Health{}))
+				q := cardinal.NewSearch(wCtx).Contains(cardinal.SearchComponent[Health]())
```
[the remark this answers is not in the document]
pyrofolium: let me see what I can do [](edited-with-Axolo)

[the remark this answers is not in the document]
pyrofolium: I did this, but here's a thing that's a bit weird about it. It makes it so that this feature can only be used by using things from two different packages. I don't mind too much though. [](edited-with-Axolo)

cardinal/search/filtercomponent.go
```diff
@@ -0,0 +1,103 @@
...
+		FilterFunc: f,
+	}
+}
+
+func (afc *andFilterComponent) Evaluate(wCtx engine.Context, id types.EntityID) (bool, error) {
+	var result = true
+	for _, filterComp := range afc.filterComponents {
+		otherResult, err := filterComp.Evaluate(wCtx, id)
+		if err != nil {
+			continue
```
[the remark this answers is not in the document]
pyrofolium: This was done with intent. See the comment on top. 

Basically some entities in the search result just won't contain component related to the property filter. In this case the evaluater will return an error rather then a true or a false. For this case the search algorithm just treats the filter as if it never existed.

cardinal/search/filtercomponent.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+	return !result, nil
+}
+
+func (ofc *orFilterComponent) Evaluate(wCtx engine.Context, id types.EntityID) (bool, error) {
+	var result = true
+	for _, filterComp := range ofc.filterComponents {
+		otherResult, err := filterComp.Evaluate(wCtx, id)
+		if err != nil {
+			continue
```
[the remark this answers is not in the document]
pyrofolium: Also same response as above. Open to discussion about this. lmk.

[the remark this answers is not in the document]
jerargus: I'll resolve this chain in favor of the one above.

cardinal/search/filtercomponent.go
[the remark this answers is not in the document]
jerargus: I don't know the exact right thing to do, but it looks inconsistent to have the notFilterComponent return an error and this continue.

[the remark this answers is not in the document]
pyrofolium: > I don't know the exact right thing to do, but it looks inconsistent to have the notFilterComponent return an error and this continue.

Yeah, I'll take a look at this.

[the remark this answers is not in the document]
pyrofolium: @jerargus Yeah, you're right there's a subtle thing going on here. 

the and, or filters will never return an error which is inconsistent with Not. The issue is subtle. What happens is the and, or filters default to true. There is a case where this shouldn't happen, it's when all evaluations error out. In this case the filter should also return an error. This makes it consistent with the Not filter.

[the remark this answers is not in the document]
pyrofolium: So in short this is how it works. 

Any filter that is composed of other filters, will only return an error if ALL internal filters also return an error. Otherwise if only some internal errors return an error, the filter will treat that individual filter as if it never existed.

pyrofolium:
https://argus-labs.slack.com/archives/D05MXTVLSUB/p1711741338073329

@jerargus caught a bug in the Not filter which was one of this filters without coverage. This bug has been fixed.

cardinal/search/filtercomponent.go
```diff
@@ -0,0 +1,118 @@
...
+	"pkg.world.dev/world-engine/cardinal/types/engine"
+)
+
+// This package involves primitives for search.
+// It involves creating and combining primitives that represent
+// filtering properties on components.
+
+type PredicateEvaluator interface {
+	Evaluate(wCtx engine.Context, id types.EntityID) (bool, error)
+}
```
technicallyty, line 16: godoc here as well, would help for those who are unfamiliar with this section of the code

cardinal/search/search.go
```diff
@@ -24,30 +24,95 @@ type cache struct {
...
+		reader:                  wCtx.StoreReader(),
+		wCtx:                    wCtx,
+		componentPropertyFilter: nil,
+	}
+}
+
+func (s *Search) Contains(component ...componentWrapper) *Search {
+	if s.filter != nil {
+		panic("Search already contains a filter.")
+	}
```
technicallyty, line 55: is this case ever possible? would it be bad if we just let it continue and overwrite the filter if there was one?

cardinal/search/searchcomponent.go
```diff
@@ -0,0 +1,25 @@
...
+
+// This file represents primitives for search
+// The primitive wraps the component and is used in search
+// the purpose of the wrapper is to prevent the user from ever
+// instantiating the component. These wrappers are used to check if entities
+// contain the specified component during the search.
+
+type componentWrapper struct {
+	Component types.Component
+}
```
technicallyty, line 13: any reason we can't just return a `types.Component` from the function below? not sure i understand why we need the wrapper

pyrofolium (commented):
responded.

cardinal/search/search.go
technicallyty: ahh i see. thanks!

cardinal/search/searchcomponent.go
technicallyty: you'll get a compile error if you try to pass in a struct value to the generic parameter though

this won't compile:
```
cardinal.NewSearch(wCtx).Contains(search.Component[EnergyComponent{Amt: 3}]())
```

technicallyty: oh wait i think i see why now

technicallyty: nevermind lol

cardinal/search/search.go
```diff
@@ -154,3 +226,59 @@ func (s *Search) evaluateSearch() []types.ArchetypeID {
 	cache.seen = s.reader.ArchetypeCount()
 	return cache.archetypes
 }
+
+func (s *Search) And(otherSearch *Search) *Search {
```
smsunarto, line 230: These logical operators (Or/And/Not) produce invalid behavior. The boolean algebra doesn't check out if you write up the truth table.

Here's an example: https://docs.google.com/spreadsheets/d/1LHyQPTGi3mwpJYWARaU_CCtk6PczTjfro79q3hpG_zw/edit?usp=sharing

smsunarto (changes requested):
Not done with the review yet, but want to flag what might be a significant flaw.

cardinal/search/search.go
pyrofolium: Hey @smsunarto I took a quick look at your spread sheet. I tried to implement line 2 and it worked. See cardinal/search_test.go in this PR. 

Not sure if that's how you did the query. Let's connect Monday to see what you're talking about.

pyrofolium: Additionally it's a bit wierd the way the spread sheet describes what's going on. 

Take a look at H and I. H filters for component types. I filters for component properties. 

while H can exist on it's own I can only exist and modify the value of the component type filter. It does not technically make sense to say (H && I). So if an AND operator was used to connect the two broken queries there might be problems. This is untested. It's likely better to force a giant panic so that these queries can be used correctly.

pyrofolium: nm about the earlier comments. I see what's going on here. I'll have to rethink the operators.

pyrofolium: I'm thinking the component property filters just have to be all permanently anded together under every search operator.

pyrofolium: Operators are now only allowed to work on Searches without the where clause. 

This makes it so that the Where clause can only be called at the tail end of the search. This prevents predicates from ever being anded/ored.

smsunarto: The solution for this is pretty straightforward, just and/or/not the resulting entity ids after executing the searches separately, i already wrote this down in the spreadsheet [](edited-with-Axolo)

pyrofolium:
I must have missed that. I'm out right now so can't conveniently view your
spreadsheet. Will look when available. I thought about this route though.
The reason why I didn't go with it is because it's not clear what it means
to NOT/AND/OR the entity id or a collection of entity ids. For example NOT
calls for an inverse value but what is an inverse of an id number? A
negative number? Those were my initial thoughts.

I will take a look at what you wrote when I can.

cardinal/search/search.go
pyrofolium: Moved this reply here from below: 

I must have missed that. I'm out right now so can't conveniently view your
spreadsheet. Will look when available. I thought about this route though.
The reason why I didn't go with it is because it's not clear what it means
to NOT/AND/OR the entity id or a collection of entity ids. For example NOT
calls for an inverse value but what is an inverse of an id number? A
negative number? Those were my initial thoughts.

I will take a look at what you wrote when I can.

pyrofolium: I took another look. The excel sheet is not entirely clear to me. 

Are you implying that the Boolean operators become set operators at the search level?

Or = union
And = intersection 
Not = complement

I'm a bit against this. It involves additional processing (looping).

smsunarto: It's clear to me and I'm pretty sure anyone else on what it means. It just means that the entity is in search 1 AND/OR in search 2. The reason why your not example doesn't make sense is because it's on the same class as AND/OR operators. NOT is a negation, so you need to use it with AND/OR and not by itself.

And/or boolean algebra is analagous to union/intersect, this is just how it works in math.

I don't see any other way to do it other than this; the way you implemented produces results that doesn't make intuitive sense [](edited-with-Axolo)

smsunarto: That aside, I don't think And/Or should be implemented as methods of search. 

It's impossible to get the two searches aligned using indentation, so it's not as readable. It should probably just be a top level function. [](edited-with-Axolo)

pyrofolium: > It's clear to me and I'm pretty sure anyone else on what it means. It just means that the entity is in search 1 AND/OR in search 2. The reason why your not example doesn't make sense is because it's on the same class as AND/OR operators. NOT is a negation, so you need to use it with AND/OR and not by itself.
> 
> And/or boolean algebra is analagous to union/intersect, this is just how it works in math.
> 
> I don't see any other way to do it other than this; the way you implemented produces results that doesn't make intuitive sense

Apologies, when I said it's not entirely clear to me I meant that in a completely neutral way. Not saying anything about the quality of your explanation or about me or about whether others understand it or not. More in the sense of "I don't get it, I'll have to talk to you about it for more clarity on my end." That's it. Honestly I wasn't even thinking about anything else, but apologies it didn't come off this way. 

The isomorphism between SET and boolean actually exists beyond just SET and Bool. It works for number theory and even graph theory. It even exists for cryptography primitives which you would likely know better than me. It's just abstract algebra and it's everywhere. Personally, I didn't fully understand what was going on with your sheets because there were expressions in that sheet where using the same symbols are used for boolean and set operators in a single expression. That's just me though, don't worry about it. 

Additionally note that NOT does work as a Set operator independently of other operators. If you have 3 entities total in the universe: [1,2,3], then NOT(1) = [2,3], there isn't a special case here. Basically if there's a NOT operator the search has to go into a secondary loop to remove relevant entities from a copy of the totality of entities to generate the complement. 

But this is besides the point. All I'm saying is that introducing a new type of operator that operates on SETs leads to a additional algorithmic complexity overhead. You are initializing a second loop for NOT and OR. Only AND does not need a secondary loop. It's also redundant on top of filters too, that is unless you want to get rid of the algebra library on filters all together? 

My personal opinion is the way I designed it now (post your post with the sheets) where the ordering of the methods makes it so operators have to be used before the WHERE clauses makes it exactly identical to how it's used before search V2 is more efficient as we won't need to launch a new loop. 

We can go whatever. I'm cool with whatever like always. This is your call, I'm just offering my perspective. 

> That aside, I don't think And/Or should be implemented as methods of search.
> 
> It's impossible to get the two searches aligned using indentation, so it's not as readable. It should probably just be a top level function.

Yeah this is just aesthetic. I agree stylistically with this, i prefer functional programming over OOP anyway.

pyrofolium: How about I just implement the functional operators on entity ids and I kill the filter library in subsequent PR because it becomes redundant.

smsunarto: All of this is unclear to me. I need to see a properly written down spec/reference implementation if you want me to consider any alternates [](edited-with-Axolo)

pyrofolium: @smsunarto design doc:

https://www.notion.so/arguslabs/Alternative-Operations-for-Search-9359dd288e06471795dc1771bda30fcc?pvs=4

pyrofolium: Changes in the discussion in the notion doc have been implemented.

pyrofolium: @smsunarto I didn't realize you introduced ANOTHER proposal within my proposal. I took a quick glance and I thought it was just an example of the existing proposal. I'm getting pulled in multiple directions here. I'll take a deeper look. 

I would recommend next time if there's a new proposal, preface it with a description rather then just a code example. I need some context here on why and what are the new changes.

smsunarto: Well you didn't ask any questions, you just assumed what you wanted to believe and ran with it. [](edited-with-Axolo)

pyrofolium: Yeah I did, but a random example in my proposal that's actually an entirely NEW design wasn't expected either [](edited-with-Axolo)

pyrofolium: I'm sort of getting pulled in multiple directions here [](edited-with-Axolo)

pyrofolium: there IS a design doc here btw and this isn't changing any of my issues [](edited-with-Axolo)

pyrofolium: there's two proposals, and a third proposal within my second proposal [](edited-with-Axolo)

pyrofolium: anyway no worries, I'll take a look and see what I can do [](edited-with-Axolo)

pyrofolium: just a little confused here with all the designs [](edited-with-Axolo)

cardinal/search/filtercomponent.go
```diff
@@ -0,0 +1,83 @@
...
+
+import (
+	"github.com/rotisserie/eris"
+
+	"pkg.world.dev/world-engine/cardinal/types"
+	"pkg.world.dev/world-engine/cardinal/types/engine"
+)
+
+// This package involves primitives for search.
+// It involves creating and combining primitives that represent
```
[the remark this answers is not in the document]
pyrofolium: Do this in a later PR.

cardinal/search/composedsearch.go
```diff
@@ -0,0 +1,269 @@
...
+		ids, err := search.collect(eCtx)
+		if err != nil {
+			return nil, err
+		}
+		for _, id := range ids {
+			resMap[id] = true
+		}
+	}
+	res := make([]types.EntityID, 0)
+	for id := range resMap {
```
jerargus, line 56: Looping over map keys is nondeterministic. res should be declared at the top of the function and appends should take place in the nested for loop near `resMap[id]=true`. The resMap should still be used to make sure duplicate ids are not added to the slice.

cardinal/search/composedsearch.go
```diff
@@ -0,0 +1,269 @@
...
+}
+
+func (andSearch *AndSearch) evaluateSearch(eCtx engine.Context) []types.ArchetypeID {
+	searchCounts := make(map[types.ArchetypeID]int)
+	for _, search := range andSearch.searches {
+		ids := search.evaluateSearch(eCtx)
+		for _, id := range ids {
+			v, ok := searchCounts[id]
+			if !ok {
+				searchCounts[id] = 0
```
jerargus, line 162: Same as above, this can be simplified to `searchCounts[id]++`. It also fixes this off-by-one error here.

cardinal/search/composedsearch.go
pyrofolium: I don't understand this request. Non-determinism doesn't have to do with anything here? 

It seems you're just requesting to re arrange the primitives to do the same thing? Please explain what you are thinking here.

jerargus: If I construct a Search that uses this OrSearch, and ends with a First(), a collection of entity IDs will put into a slice, and I'll be handed whatever entity ID happens to be at index 0. Given identical runs of the same tick, the set of items in the slice will always be the same, but the order of the items may change from run to run (because ranging over maps is non-deterministic). This means the entity ID returned from First() is nondeterministic.

My suggestion is to rearrange the existing primitives, but the end result will be a deterministic ordering of entity IDs that won't change between identical runs of the same tick.

If this explanation isn't clear, or if I'm missing something, let's hop on a call.

pyrofolium: This is correct behavior. Ordering isn't a factor in search [](edited-with-Axolo)

pyrofolium: I understand what you're saying [](edited-with-Axolo)

pyrofolium: First will be nondeterministic, we should absolutely never allow the user to rely on determinism here because it will be an arbitrary determinism/order with no rhyme or reason. [](edited-with-Axolo)

pyrofolium: If we want determinism that should be an explicit feature which isn't part of the spec. I would impliment it as some type of order [](edited-with-Axolo)

technicallyty: can we just sort it by entity ID? search will need to return the exact same results given the same state.

cardinal/search/search.go
```diff
@@ -167,7 +167,39 @@ func (s *Search) Each(eCtx engine.Context, callback CallbackFn) (err error) {
...
+	var acc types.EntityID = 0
+	for _, id := range ids {
+		if id > acc {
+			acc = id
+		}
+	}
+	return acc
+}
+
+func fastSortIds(ids []types.EntityID) {
```
technicallyty: we can also just use `slices.Sort` , it uses pdqs under the hood, so should be fast enough

cardinal/cardinal.go
```diff
@@ -31,8 +31,37 @@ var (
...
+var NewLegacySearch = search.NewLegacySearch
+
+type Search = search.Search
+
+// FilterFunction wrap your component filter function of func(comp T) bool inside FilterFunction to use
+// in search.
+//
+// Usage:
+//
+//	cardinal.NewSearch(worldCtx).
```
smsunarto, line 56: This documentation doesn't seem to hold true anymore since you are moving the wCtx to the `Each`/`First` instead of NewSearch

cardinal/cardinal.go
```diff
@@ -31,8 +31,37 @@ var (
...
+//
+// cardinal.NewSearch(worldCtx).Exact(search.Component[AlphaTest]())
 var NewSearch = search.NewSearch
 
+// NewLegacySearch allows users to create a Search object with a filter already provided
+// as a property.
+//
+// Usage:
+//
+// cardinal.NewSearchWithFilter(wCtx, filter.Exact(Alpha{}, Beta{}))
```
smsunarto, line 46: this needs to be updated

cardinal/search_test.go
```diff
@@ -0,0 +1,309 @@
...
+			if id <= ids[index+1] {
+				continue
+			}
+			return false
+		}
+	}
+	return true
+}
+
+func TestSetOperationsOnSearch(t *testing.T) {
```
smsunarto, line 132: Use table driven test

cardinal/search_test.go
smsunarto, line 1: These tests needs to be more granular and function names should clearly communicate what it is actually testing (https://ieftimov.com/posts/testing-in-go-naming-conventions/).

Many tests in the codebase are often just jumbles of happy path integration test jammed into a single function that makes it unclear how good the testing coverage is. For example, it's unclear to me here how `TestFullSearch` is different with `TestSearchExample`. The goal could be the same or it could be different, but it's impossible to tell because there are no comments nor is the function name descriptive enough.

Cases that I would like to see better coverage on:
- `Exact` strictly only returns entities that have exactly the same set of component
- `Contains` returns all entities that contains the component requested.
- What happens if the search targets a component that is not registered?
- What happens if in a set operator (i.e. search.And) one of the queries have a component that is not registered? How about other errors from the individual searches?
- Error handling of `.Where(...)` what is the expected behavior when an error is returned?

smsunarto (changes requested):

cardinal/cardinal.go
pyrofolium: Yeah I'll update that.

cardinal/search/composedsearch.go
pyrofolium: Yeah this is done.

cardinal/search_test.go
pyrofolium: Most of these tests existed before. I can modify them accordingly.

pyrofolium: @smsunarto added some tests that illustrates the behavior requested above. Let me know if you want some adjustments to the behavior here.

cardinal/search_test.go
```diff
@@ -305,3 +305,197 @@ func TestSearchExample(t *testing.T) {
 	assert.NilError(t, err)
 	assert.Equal(t, amount, 5)
 }
+
+func TestExactFilterOnSearch(t *testing.T) {
```
smsunarto, line 309:
```suggestion
func TestSearch_Exact_ReturnsExactComponentMatch(t *testing.T) {
```

The idea of this convention is `Test<relevant package/type/function>_<method or context>_<tested behavior>`

cardinal/search_test.go
```diff
@@ -305,3 +305,197 @@ func TestSearchExample(t *testing.T) {
...
+	assert.NilError(t, err)
+	_, err = cardinal.CreateMany(worldCtx, 10, AlphaTest{}, BetaTest{}, GammaTest{})
+	assert.NilError(t, err)
+
+	amt, err := search.NewSearch().Entity(filter.Exact(filter.Component[BetaTest]())).Count(worldCtx)
+	assert.NilError(t, err)
+	assert.Equal(t, amt, 12)
+}
+
+func TestContainsFilterOnSearch(t *testing.T) {
```
smsunarto, line 342:
```suggestion
func TestSearch_Contains_ReturnsEntityThatContainsComponents`(t *testing.T) {
```

cardinal/search_test.go
```diff
@@ -305,3 +305,197 @@ func TestSearchExample(t *testing.T) {
...
+	assert.NilError(t, err)
+	_, err = cardinal.CreateMany(worldCtx, 10, AlphaTest{}, BetaTest{}, GammaTest{})
+	assert.NilError(t, err)
+
+	amt, err := search.NewSearch().Entity(filter.Contains(filter.Component[BetaTest]())).Count(worldCtx)
+	assert.NilError(t, err)
+	assert.Equal(t, amt, 42)
+}
+
+func TestUnregisteredComponentOnSearch(t *testing.T) {
```
smsunarto, line 375:
```suggestion
func TestSearch_ComponentNotRegistered_ReturnsZeroEntityWithNoError(*testing.T) {
```

cardinal/search_test.go
smsunarto: There are cases where you have to bend the convention here, but the idea is to highlight the relevant function/methods/package and the behaviour that you want to pinpoint in the test.

cardinal/search_test.go
```diff
@@ -305,3 +305,197 @@ func TestSearchExample(t *testing.T) {
...
+	assert.NilError(t, err)
+	_, err = cardinal.CreateMany(worldCtx, 10, AlphaTest{}, BetaTest{}, GammaTest{})
+	assert.NilError(t, err)
+
+	amt, err := search.NewSearch().Entity(filter.Contains(filter.Component[HP]())).Count(worldCtx)
+	assert.NilError(t, err)
+	assert.Equal(t, amt, 0)
+}
+
+func TestUnregisteredComponentOnSetOperators(t *testing.T) {
```
smsunarto, line 408:
```suggestion
func TestSearch_SetOperator_ComponentNotRegistered_NoError(t *testing.T) {
```

cardinal/search_test.go
```diff
@@ -305,3 +305,197 @@ func TestSearchExample(t *testing.T) {
...
+		},
+	}
+	for _, searchStruct := range tests {
+		amt, err := searchStruct.search.Count(worldCtx)
+		assert.NilError(t, err)
+		assert.Equal(t, amt, searchStruct.count)
+	}
+}
+
+func TestWhereClauseOnSearch(t *testing.T) {
```
smsunarto, line 463:
```suggestion
func TestSearch_Where_FiltersEntityCorrectly(t *testing.T) {
```

cardinal/search_test.go
```diff
@@ -184,7 +184,7 @@ func TestSetOperationsOnSearch(t *testing.T) {
 	}
 }
 
-func TestSearchExample(t *testing.T) {
+func TestSearchEntityMethod(t *testing.T) {
```
smsunarto, line 187:
```sugestion
func TestSearch_Integration(t *testing.T) {
```

I recommend following this convention: `Test<relevant package/type/function>_<method or context>_<tested behavior>`

There are cases where you have to bend the convention here, but the idea is to highlight the relevant function/methods/package and the behavior that you want to pinpoint in the test.

smsunarto (approved):
left test function name nits, but lgtm

cardinal/search/search.go
pyrofolium: https://pkg.go.dev/sort#Sort

The sort here is actually O(N). Technically, O(max_entity_id_value). It works because of the way we create entity ids.

jerargus (approved):

cardinal/search/search.go
pyrofolium: I switched this so slices.Sort. I was mistaken about the way we create entity ids. This will cause a problem if we use this style of sort.

PR #724, opened 2024-04-02
feat(nakama): Auto reclaim persona tags when a transaction fails

jerargus:
Closes: WORLD-1022

## Overview

Expand the Nakama method that forwards transactions from clients to automatically re-register persona tags if there seems to be a problem with an unrecognized persona tag.

The use-case for this is in development environments when an engineer will reset/clear the Cardinal DB, but does not clear the Nakama DB. Nakama thinks certain persona tags are still active, but Cardinal will complain about a missing signer address.

This PR recognizes this specific failure case and re-creates the persona tag on Cardinal before retrying the request. From the client's perspective, the re-create and re-try code path happens automatically.

Two unpleasant notes about this PR:

1. time.Sleep is used after re-creating the persona tag. Nakama needs to give Cardinal enough time to process a tick. Ideally, the process would block until the relevant receipt comes back from Cardinal, but since this is essentially a dev-only feature, 
2. There is no automatic test for this code path. [See this doc for some discussion on why this is hard to test](https://www.notion.so/arguslabs/Automatic-Persona-Tag-Fixing-6786bf1ec4c3451ba819c13091843141?pvs=4#1d2703724021465694545d29eb5b8164)

## Testing and Verifying

These steps turned into a wall of text; I apologize. In summary: These instructions describe setting up a testing environment by starting fresh with Docker, registering and verifying a persona tag, simulating a game transaction, and then observing the issue with persona tags after clearing redis. It concludes with switching branches to apply this fix and verifying that transactions succeed post-fix.

1. Start off in the `main` branch.
2. (probably optional) Delete all docker containers and volumes to start fresh.
3. From world-engine, run `docker compose up nakama --build`
4. Navigate to http://localhost:7351/#/apiexplorer
5. Send Request to the `AuthenticateDevice` endpoint with the following body:
    * ` {"account": {"id": "1234567890"},"create": true,"username": "some-cool-user"}`
6. Navigate to http://localhost:7351/#/accounts and copy the User ID of your newly authenticated user
7. Navigate back to http://localhost:7351/#/apiexplorer and paste the User ID into the `set user ID as request context` field.
8. Send Request to the `nakama/claim-persona` endpoint with the following body:
    * `{"personaTag": "sometag"}`
9. Send Request to the `nakama/show-persona` endpoint (no body required)
    * You should see a `status: accepted` field, indicating you've registered a persona tag successfully.
    * You're now ready to send actual game transactions to Cardinal.
10. Send Request to the `tx/game/join` endpoint with the following (empty) body:
    * `{}`
    * The response should contain a `TxHash` and `Tick` field, indicating a successfully submitted transaction.
11. From the docker UI, stop ALL running containers.
12. Delete the `redis` docker container.
13. Re-run `docker compose up nakama --build`
14. Re-do step 10. THIS SHOULD FAIL
    * The error message should contain `persona tag does not have a signer`
    * Cardinal's state was re-set, but Nakama still has a person tag associated with your User ID.
15. Stop all docker contaienrs.
16. Checkout this branch with: `git checkout jer/world-1022-reregister-persona-tags`
17. Re-run `docker compose up nakama --build`
18. Re-do step 10. THIS SHOULD SUCCEED
    * There will be a brief pause as Nakama waits for Cardinal to complete a tick.
    * The response body should contain `TxHash` and `Tick` fields.


smsunarto:
@jerargus is there a reason why we are not waiting for the receipts instead?

jerargus:
> @jerargus is there a reason why we are not waiting for the receipts instead?
@smsunarto 

[I made a ticket to use receipts instead of time.Sleep](https://linear.app/arguslabs/issue/WORLD-1027/reclaim-persona-tag-should-not-use-timesleep). I opted to not include it in the initial PR because hooking this into the receipt system might get complicated; I wanted to keep the initial feature targeted. Also, this is a dev-only feature, so the occasional delayed request isn't a probem.

smsunarto (approved):

PR #726, opened 2024-04-02
feat: proxy method to query historical transactions from the base shard

technicallyty:
## Overview

implements a query transactions on `Router` that forwards requests to the evm's `x/shard` query transaction handler. also fixes a bug with recovering txs.

goal of this PR was to get rid of the dependency on the EVM in cardinal. previously, we used cosmos SDK's built in module grpc query system to query data from the chain. this is problematic because we have to import cosmos sdk stuff into cardinal, which then requires us to use replace statements.

this removes that dependency by implementing the transaction query through rift instead. rift acts like a Nakama (cardinal == game client, evm == backend client). 

you might suggest "why not just slap a go.mod in the grpc types that cosmos SDK uses", but unfortunately they use a modified code generator that imports cosmos sdk, so we'd have the same problem again. thus, proxy query!

## Testing and Verifying

`make e2e-evm` - i added a case that checks the query against the rift endpoint works as does the cosmos SDK endpoint. i also shut down the cardinal instance and restarted it to see that the recovery system works and the state is rebuilt! i'm going to add a ticket to make a more robust recovery test.


technicallyty:
ive opened https://linear.app/arguslabs/issue/WORLD-1038/remove-base-shard-query-address-from-config to make sure we remove the no longer needed configuraiton from world-cli.

smsunarto (approved):

PR #739, opened 2024-04-19
fix(nakama): Jer/world 1011 string events should be wrapped in json

jerargus:
Closes: WORLD-1011

## Overview

Previously, a non-json encoded string event would result in a json.Unmarshal error, which was then ignored.

Now when we encounter an error, the string is wrapped in a `{"message": "<the event>"}` json blob.

## Testing and Verifying

The e2e game has been adjusted to emit both a map event and a string event. The relevant test has been updated to check for both strings.

### DIFF
--- a/relay/nakama/main.go
+++ b/relay/nakama/main.go
@@ -167,8 +167,13 @@ func initEventHub(
 		ch := eventHub.SubscribeToEvents("main")
 		for event := range ch {
 			content := make(map[string]any)
-			err = json.Unmarshal(event, &content)
-			err := eris.Wrap(nk.NotificationSendAll(ctx, "event", content, 1, false), "")
+			err := json.Unmarshal(event, &content)
+			if err != nil {
+				// The event content isn't in JSON format. Wrap whatever it is in a JSON blob.
+				content["message"] = string(event)
+			}
+
+			err = eris.Wrap(nk.NotificationSendAll(ctx, "event", content, 1, false), "")
 			if err != nil {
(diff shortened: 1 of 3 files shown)


relay/nakama/main.go
```diff
@@ -167,8 +167,13 @@ func initEventHub(
...
 			content := make(map[string]any)
-			err = json.Unmarshal(event, &content)
-			err := eris.Wrap(nk.NotificationSendAll(ctx, "event", content, 1, false), "")
+			err := json.Unmarshal(event, &content)
+			if err != nil {
+				// The event content isn't in JSON format. Wrap whatever it is in a JSON blob.
+				content["message"] = string(event)
+			}
+
+			err = eris.Wrap(nk.NotificationSendAll(ctx, "event", content, 1, false), "")
```
technicallyty, line 176: not sure if this is meant to be a patch fix or not, but what if we just didn't unmarshal at all? IMO we should just be transferring the bytes received. not sure we gain anything by adding the generic unmarshal step

technicallyty: like we could just do `content["message"]=event`?

technicallyty: the `event` here should just be JSON marshalled `TickResults`, so i'm not actually sure whats causing the failure tbh

technicallyty: nvm talked with jeremy, this might be ugly

technicallyty (approved):

PR #744, opened 2024-04-22
fix(cardinal): Ensure components to delete are cleared between ticks

jerargus:
Closes: WORLD-1010

## Overview

compValuesToDelete keeps track of what entity/components need to be deleted from redis. Previously, this store was never cleared out between ticks. Consequently, if component X is delete on tick 10, every tick after 10 would also get a redis command to delete that component X. 

This set of components to delete would grow without bounds and cause the finalize tick portion of the tick to increase over time.

## Brief Changelog

Ensure compValuesToDelete is cleaned up between ticks. 
Add a test to ensure the struct is empty between ticks.

## Testing and Verifying

Unit test added.

### DIFF
--- a/cardinal/gamestate/redis.go
+++ b/cardinal/gamestate/redis.go
@@ -230,6 +230,9 @@ func (m *EntityCommandBuffer) addComponentChangesToPipe(ctx context.Context, pip
 			return eris.Wrap(err, "")
 		}
 	}
+	if err = m.compValuesToDelete.Clear(); err != nil {
+		return eris.Wrap(err, "failed to clear to-be-deleted component values store")
+	}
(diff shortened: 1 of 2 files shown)


smsunarto (approved):
LGTM. What do you think the post-mortem of this would be? How can we avoid it going forward? Any other places that we should be looking at?

pyrofolium:
> LGTM. What do you think the post-mortem of this would be? How can we avoid it going forward? Any other places that we should be looking at?

There was a similar bug a while back with a memory leak in the ECB. We can use pprof on a long running game test to maybe detect this? But it won't be real time.

pyrofolium (approved):
lgtm

PR #745, opened 2024-04-23
add agar game to benchmark.

pyrofolium:
Closes: WORLD-1082

This is a draft for now. 

Adds benchmarking for the agar game. So you can view the profile 

In the root run `make agar-benchmark`

Let it run for a bit

Then run `go tool pprof -http 127.0.0.1:8080 agar.cpu.prof` in the profile directory.


pyrofolium:
@smsunarto need your input here as well on whether we should have agar code used for an e2e benchmark in world-engine.

smsunarto:
I prefer a much simpler and robust game to use as a benchmark. 

We can simply turn it into a 2D grid where the player can move WASD in discrete steps and then use that for continuous profiling + chaos monkey test with the synthetic continuous workload. 

Having a full complex game makes upgrading the E2E test a chore on breaking changes. Also, agar shooter have so many things that I would consider an anti-pattern which would just create red herrings.

pyrofolium:
> I prefer a much simpler and robust game to use as a benchmark.
> 
> We can simply turn it into a 2D grid where the player can move WASD in discrete steps and then use that for continuous profiling + chaos monkey test with the synthetic continuous workload.
> 
> Having a full complex game makes upgrading the E2E test a chore on breaking changes. Also, agar shooter have so many things that I would consider an anti-pattern which would just create red herrings.

Good point on the upgrade part. I agree with this. I'll just leave this here as a branch then in case on the small chance people want to test anything with this.

smsunarto:
going to close this, can revisit at a later time

PR #750, opened 2024-04-29
perf(cardinal): Only rebuild the persona tag index when needed

jerargus:
## Overview

Previously, these persona-tag related systems would build an index of every persona-tag -> signer address every single tick. If no persona-tag related messages were present, this index would go unused.

With this modification, these persona-tag -> signer address map will only be built if it's actually going to be used.

## Testing and Verifying

Unit test added. It ensures the persona-tag related systems don't slow down significantly when no persona-tag messages need to be processed.

### DIFF
--- a/cardinal/plugin_persona.go
+++ b/cardinal/plugin_persona.go
@@ -19,7 +19,17 @@ import (
 	"pkg.world.dev/world-engine/cardinal/types/engine"
 )
 
-var _ Plugin = (*personaPlugin)(nil)
+var (
+	_ Plugin = (*personaPlugin)(nil)
+
+	// globalPersonaTagToAddressIndex keeps track of the mapping of persona-tags->signer-address so it doesn't need to
+	// be recomputed each tick.
+	globalPersonaTagToAddressIndex personaIndex
+	// The tick that the globalPersonaTagToAddressIndex was built on. In normal usage, wCtx.CurrentTick should always
+	// be greater than this number, but during tests the currentTick will be reset. Tracking this number at the global
+	// is easier than updating each test to reset this global value.
+	tickOfPersonaTagToAddressIndex uint64
+)
 
 type personaIndex = map[string]personaIndexEntry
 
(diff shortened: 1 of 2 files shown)


cardinal/plugin_persona.go
```diff
@@ -102,15 +102,25 @@ func (p *personaPlugin) RegisterMessages(world *World) error {
 // users who want to interact with the game via smart contract can link their EVM address to their persona tag, enabling
 // them to mutate their owned state from the context of the EVM.
 func AuthorizePersonaAddressSystem(wCtx engine.Context) error {
-	personaTagToAddress, err := buildPersonaIndex(wCtx)
-	if err != nil {
-		return err
-	}
+	// The persona tag index will be lazily built for the first AuthorizePersonaAddress message. If no such messages
+	// are present, the index will not be built.
+	var personaTagToAddress personaIndex
```
smsunarto, line 107: Wouldn't it be better if we have an index that persist across ticks?

smsunarto: I'd say that we have to do this because otherwise we would have a tick drift everytime someone interacts with Persona [](edited-with-Axolo)

smsunarto:
I'd say that we have to do this because otherwise we would have a tick drift everytime someone interacts with Persona

cardinal/plugin_persona.go
smsunarto: At least until we have a proper way to do indexing [](edited-with-Axolo)

smsunarto:
At least until we have a proper way to do indexing

cardinal/plugin_persona.go
pyrofolium: the concept of an index will always be "out of ecs" until we change storage. That's where an index would typically go. I agree with scott here, a bit of a temp hack is ok [](edited-with-Axolo)

pyrofolium:
the concept of an index will always be "out of ecs" until we change storage. That's where an index would typically go. I agree with scott here, a bit of a temp hack is ok

cardinal/plugin_persona.go
jerargus: OK, persona tag tracking has been moved to an in-memory var. 

the `tickOfPersonaTagToAddressIndex` is used to tell if we're in a test so the index can be rebuilt. This seemed more straight forward than creating a "clear index" public function or trying to use an init function. If either of you have an alternative suggestion I'm all ears.

cardinal/plugin_persona.go
```diff
@@ -19,7 +19,17 @@ import (
...
+	_ Plugin = (*personaPlugin)(nil)
+
+	// globalPersonaTagToAddressIndex keeps track of the mapping of persona-tags->signer-address so it doesn't need to
+	// be recomputed each tick.
+	globalPersonaTagToAddressIndex personaIndex
+	// The tick that the globalPersonaTagToAddressIndex was built on. In normal usage, wCtx.CurrentTick should always
+	// be greater than this number, but during tests the currentTick will be reset. Tracking this number at the global
+	// is easier than updating each test to reset this global value.
+	tickOfPersonaTagToAddressIndex uint64
+)
```
pyrofolium, line 32: Globals are bad, but this is a temp hack. It should be mentioned in the comments.

cardinal/plugin_persona.go
```diff
@@ -221,18 +234,18 @@ func buildPersonaIndex(wCtx engine.Context) (personaIndex, error) {
 				return true
 			}
 			lowerPersona := strings.ToLower(sc.PersonaTag)
-			personaTagToAddress[lowerPersona] = personaIndexEntry{
+			globalPersonaTagToAddressIndex[lowerPersona] = personaIndexEntry{
```
pyrofolium, line 237: Is index the right word for this? This looks like memoization. 

A general memoization module owned by eCtx or world would be more palatable for reviewers. 

In this case it can live independent the concept of an index as it's a much more simpler concept. 

I personally don't care. It can live as a global, but in this case it must be marked very loudly as temporary. If not then a general memoization module surrounding certain logic and owned by world/eCtx works too.

cardinal/plugin_persona.go
jerargus: I added a TODO and linked to the linear ticket related to indexing.

cardinal/plugin_persona.go
jerargus: The name matches the terminology for the already-existing struts.

A more general memoization module in wCtx would be more involved and requires a design pass. This PR is meant to fix a specific performance problem.

pyrofolium: Highly disagree. This PR itself fails a design pass like a red flag, so even a application specific memoization module or even moving these variables out of the global scope is better than what you did here. Idc to much though if this will be removed. I'll leave it to Scott to do the final approval.

PR #752, opened 2024-04-30
fix: router fixes

technicallyty:
## Overview

some various fixes for router stuff.

## Brief Changelog

- updates some docker compose values from recent **breaking** changes in cardinal
- moves sequencer code to router directory
- updates namespace and shard storage module to all use the same module account address.
- updates the start-sequencer.sh script with a hard coded address. the env variable piping didn't work. this should be updated later on.


evm/scripts/start-sequencer.sh
```diff
@@ -50,9 +50,10 @@ cp app.toml /root/.world-evm/config/app.toml
 # CometBFT API
 sed -i'.bak' 's#"tcp://127.0.0.1:26657"#"tcp://0.0.0.0:26657"#g' /root/.world-evm/config/config.toml
 
-# Faucet setup. changes the default account to the provided account, then supplies it with tokens.
-sed -i'.bak' "s#'20f33ce90a13a4b5e7697e3544c3083b8f8a51d4'#'$FAUCET_ADDR'#g" /root/.world-evm/config/genesis.json
```
technicallyty, line 54: for some reason this env variable injection into the sed command wasn't working. not sure how to fix. so just gonna hard code it with the default for now.

all this address does is give account EVM funds. this way you can test cross-shard messaging.

PR #753, opened 2024-05-01
feat(cardinal): Jer/remove recovery tick and recovery transactions

jerargus:
## Overview

Remove the code related to recovering from an incomplete system run. Pending transactions are no longer saved to the DB and there's no attempt to rerun the last failing tick on startup. If an error is encountered during a tick, any queued up transactions will be discarded.

As a result, we only need to keep track of 1 tick number value. This value is increment in the redis transaction that also updates the state of all components.

## Testing and Verifying

Some tests that verify the old behavior have been removed.

Other tests that relied on the initial recovery tick have been modified to manually run a tick after startup.


jerargus:
I'm closing this PR since I won't be around to merge it in the future. It's not a work-in-progress PR; It's currently ready to be reviewed and merged.

PR #749, opened 2024-04-27
refactor(cardinal): move system manager to world package

smsunarto:
Closes: WORLD-XXX

## Overview

Move system manager to the world package

## Brief Changelog

- Move system manager to the world package
- Remove error/log references to "engine" that is no longer a concept that exist
- Create an interface over system manager to prevent direct access to the manager struct fields.
- Remove unnecessary methods in world that were previously simply invoking the managers
- Fix docs where system and query is accidentally capitalized, likely because of the Jetbrains auto refactor.

## Testing and Verifying

- All existing test passes.

PR #754, opened 2024-05-01
refactor(cardinal): move message and query into cardinal.

pyrofolium:
Closes: WORLD-1110

Move query and message into cardinal. Very mechanical. 

There are some name changes too. Some things like `messages.NewManager` have been changed to `cardinal.NewMessageManager` where relevant.


jerargus (approved):

cardinal/message_manager.go
smsunarto, line 1: The managers needs to be thrown under an interface like my example (https://github.com/Argus-Labs/world-engine/pull/749/files#diff-490f654845fb6b98fad55dbdaf2a4cbe5920862f802ca6bc8c6c423ff1ce4753R27-R39, https://github.com/Argus-Labs/world-engine/pull/749/files#diff-a347e172c71021c8563a1e76f2c8814a8d97f2ae01991104af1475c56fe7d4c9R49) here so that users can't accidentally modify the manager struct fields that are now directly accessible. 

Also notice that the manager struct is now private to prevent people from directly accessing it.

cardinal/world.go
```diff
@@ -60,10 +58,10 @@ type World struct {
 
 	// Core modules
 	worldStage       *worldstage.Manager
-	msgManager       *message.Manager
+	msgManager       *MessageManager
```
smsunarto, line 61: See my comment above w/r/t to interfaces

cardinal/world.go
smsunarto, line 1: There are functions that currently route method calls to the manager, such as `GetComponentByName` and `GetMessageByName`, that could be removed since are using interface composition to provide that functionality directly to World. see example here: https://github.com/Argus-Labs/world-engine/pull/749/files#diff-a347e172c71021c8563a1e76f2c8814a8d97f2ae01991104af1475c56fe7d4c9

smsunarto (changes requested):

cardinal/world.go
pyrofolium: I'll create another ticket for the other managers too.

smsunarto: my bad, yeah i meant the message and query [](edited-with-Axolo)

pyrofolium: There's problems doing this for MessageManager due to the fact that World needs to implement the Provider interface in router.go. I'm going to leave it out for now and resolve. lmk if you want me to try to attack this or if you have any ideas on it.

cardinal/cardinal_test.go
```diff
@@ -52,11 +51,11 @@ type Qux struct{}
 
 func (Qux) Name() string { return "qux" }
 
-type Health struct {
```
smsunarto, line 55: If there is already a component with the same name in the package then just reuse it

cardinal/message_manager.go
```diff
@@ -9,22 +9,30 @@ import (
...
-type Manager struct {
+type MessageManager interface {
+	RegisterMessage(msgType types.Message, msgReflectType reflect.Type) error
+	GetRegisteredMessages() []types.Message
+	GetMessageByID(id types.MessageID) types.Message
+	GetMessageByFullName(fullName string) (types.Message, bool)
+	GetMessageByType(mType reflect.Type) (types.Message, bool)
+}
+
+type MessageManagerImpl struct {
```
smsunarto, line 20: this struct doesn't need to be public. just follow the naming convention from the system.go (https://github.com/Argus-Labs/world-engine/blob/main/cardinal/system.go)

cardinal/message_manager.go
```diff
@@ -9,22 +9,30 @@ import (
...
 	// registeredMessages maps message FullNames to a types.Message.
 	registeredMessages       map[string]types.Message
 	registeredMessagesByType map[reflect.Type]types.Message
 	nextMessageID            types.MessageID
 }
 
-func NewManager() *Manager {
-	return &Manager{
+func NewMessageManager() MessageManager {
```
smsunarto, line 27: also doesnt need to be public

cardinal/query_manager.go
smsunarto, line 1: same comments as the message manager

cardinal/world.go
```diff
@@ -61,11 +59,11 @@ type World struct {
 
 	// Core modules
 	worldStage       *worldstage.Manager
-	msgManager       *message.Manager
+	msgManager       MessageManager
```
smsunarto, line 62: This should just be `MessageManager` similar to SystemManager. Move this and the QueryManager up to below the SystemManager

smsunarto (changes requested):

smsunarto (approved):

PR #757, opened 2024-05-07
refactor(cardinal): move context into cardinal

pyrofolium:
closes WORLD-1112

Very large PR. 
- context moved into cardinal package
- internal methods on cardinal package privatized
- search moved into cardinal package (filters remain in the search package)
- context dependency removed from server
- GetTransactionReceiptsForTick created to replace privatized method on ctx (for tests only)
- GetComponentByNameFromContext created to replace privatized method on ctx (for tests only)
- GetStoreManagerFromContext created to replace privatized method on ctx (for tests only)
- Query Interface moved to cardinal

For testing just make sure original tests work as no new features are created. This is just reorganization.


pyrofolium:
@smsunarto this is rough but the methods are now privatized. There's some issues with swagger. will deal with it tmrw.

cardinal/server/handler/debug.go
```diff
@@ -1,23 +1,13 @@
...
 
-type DebugStateRequest struct{}
-
-type debugStateElement struct {
-	ID         types.EntityID             `json:"id"`
-	Components map[string]json.RawMessage `json:"components" swaggertype:"object"`
-}
-
-type DebugStateResponse []debugStateElement
+type DebugStateResponse = types.DebugStateResponse
```
pyrofolium, line 10: This is done to satisfy swagger.

cardinal/server/handler/world.go
```diff
@@ -5,23 +5,13 @@ import (
...
-	Messages   []FieldDetail `json:"messages"`
-	Queries    []FieldDetail `json:"queries"`
-}
-
-type FieldDetail struct {
-	Name   string         `json:"name"`   // name of the message or query
-	Fields map[string]any `json:"fields"` // variable name and type
-	URL    string         `json:"url,omitempty"`
-}
+type GetWorldResponse = engine.GetWorldResponse
```
pyrofolium, line 14: This is done to satisfy swagger

cardinal/types/debug.go
```diff
@@ -0,0 +1,12 @@
+package types
+
+import "encoding/json"
+
+type DebugStateRequest struct{}
```
smsunarto, line 5: Request and Response type is something that is specific to server. It's not supposed to be a common type that is passed around through different packages, so I don't see a reason why this should be here.

cardinal/types/engine/info.go
```diff
@@ -0,0 +1,14 @@
+package engine
+
+type GetWorldResponse struct {
```
smsunarto, line 3: Same feedback as above, the concept of "Response" struct type is something that is specific to server, you are not supposed to put in an omnibus types package.

cardinal/types/cql.go
```diff
@@ -0,0 +1,8 @@
+package types
+
+import "encoding/json"
+
+type CqlData struct {
```
smsunarto, line 5: comment what this is for

cardinal/types/debug.go
```diff
@@ -0,0 +1,12 @@
...
+
+type DebugStateElement struct {
```
smsunarto, line 7: this seems identical to `CqlData` can we consolidate this into something we can use for both?

cardinal/types/engine/info.go
```diff
@@ -0,0 +1,14 @@
...
+	Namespace  string        `json:"namespace"`
+	Components []FieldDetail `json:"components"` // list of component names
+	Messages   []FieldDetail `json:"messages"`
+	Queries    []FieldDetail `json:"queries"`
+}
+
+type FieldDetail struct {
```
smsunarto, line 10: 1. Move this out of the `engine package. We don't want anything there anymore.
2. Comment what this is for

cardinal/types/engine/query.go
```diff
@@ -1,25 +1,3 @@
...
-	// DecodeEVMReply decodes EVM reply bytes, into the concrete go reply type.
-	DecodeEVMReply([]byte) (any, error)
-	// EncodeAsABI encodes a go struct in abi format. This is mostly used for testing.
-	EncodeAsABI(any) ([]byte, error)
-	// IsEVMCompatible reports if the query is able to be sent from the EVM.
-	IsEVMCompatible() bool
-	// GetRequestFieldInformation returns a map of the fields of the query's request type and their types.
-	GetRequestFieldInformation() map[string]any
-}
+type QueryHandler = func(name string, group string, bz []byte) ([]byte, error)
```
smsunarto, line 3: Why cant this be in the cardinal package?

cardinal/util.go
```diff
@@ -48,3 +52,34 @@ func isFatalError(err error) bool {
...
+		return nil, eris.New("error in test type assertion.")
+	}
+	return ctx.world.GetTransactionReceiptsForTick(tick)
+}
+
+func GetStoreManagerFromContext(wCtx Context) gamestate.Manager {
+	return wCtx.storeManager()
+}
+
+func GetComponentByNameFromContext(wCtx Context, name string) (types.ComponentMetadata, error) {
```
smsunarto, line 83: I saw this is only used in `search_test.go` and I don't understand why you need this and `GetStoreManagerFromContext` instead of just doing `cardinal.SetComponent`

cardinal/util.go
```diff
@@ -48,3 +52,34 @@ func isFatalError(err error) bool {
...
+	}
+	var _ types.Message = &msg
+	res, ok := tempRes.(*MessageType[In, Out])
+	if !ok {
+		return &msg, eris.New("wrong type")
+	}
+	return res, nil
+}
+
+func GetTransactionReceiptsForTick(wCtx Context, tick uint64) ([]receipt.Receipt, error) {
```
smsunarto, line 71: Unclear why you can't use `cardinal.EachMessage` instead

cardinal/util.go
```diff
@@ -48,3 +52,34 @@ func isFatalError(err error) bool {
 	}
 	return true
 }
+
+func GetMessage[In any, Out any](wCtx Context) (*MessageType[In, Out], error) {
```
smsunarto, line 56: This seems to only be used in test and is used to access the `message.In` iterator method which is already exposed via `cardinal.EachMessage`. I don't think we need this

cardinal/world.go
```diff
@@ -182,7 +181,7 @@ func NewWorld(opts ...WorldOption) (*World, error) {
 			)
 		}
 	}
-
+	world.QueryManager.BuildQueryIndex(world)
```
smsunarto, line 184: See my comment on the query manager, I don't think a query index is necessary and just makes things more clunky because you can easily forget to rebuild the query index.

cardinal/world.go
```diff
@@ -382,6 +379,21 @@ func (w *World) startServer() {
 	}()
 }
 
+func (w *World) BuildQueryFields() []engine.FieldDetail {
```
smsunarto, line 382: This should be in QueryManager

cardinal/world.go
```diff
@@ -553,6 +565,37 @@ func (w *World) UseNonce(signerAddress string, nonce uint64) error {
 	return w.redisStorage.UseNonce(signerAddress, nonce)
 }
 
+func (w *World) GetDebugState() (types.DebugStateResponse, error, error) {
```
smsunarto, line 568: just call this `GetState`. The return type shouldn't be a "Response" because "Response" is a server specific concept.

cardinal/server/handler/cql.go
```diff
@@ -34,17 +28,18 @@ type CQLQueryResponse struct {
...
+func PostCQL(
+	world servertypes.ProviderWorld) func(*fiber.Ctx) error {
 	return func(ctx *fiber.Ctx) error {
 		req := new(CQLQueryRequest)
 		if err := ctx.BodyParser(req); err != nil {
 			return err
 		}
 
-		// getComponentByName is a wrapper function that casts component.ComponentMetadata from ctx.GetComponentByName
+		// getComponentByName is a wrapper function that casts component.ComponentMetadata from ctx.getComponentByName
```
smsunarto, line 39: You can move line 39 to 53 into the `world.RunCQLSearch` method

cardinal/world.go
```diff
@@ -628,3 +671,38 @@ func (w *World) populateAndBroadcastTickResults() {
 		log.Err(err).Msgf("failed to broadcast tick results")
 	}
 }
+
+func (w *World) ReceiptHistorySize() uint64 {
+	return w.receiptHistory.Size()
+}
+
+func (w *World) RunCQLSearch(filter filter.ComponentFilter) ([]types.CqlData, error, error) {
```
smsunarto, line 679: nit: rename this `EvaluateCQL`

cardinal/world_context.go
```diff
@@ -9,14 +9,49 @@ import (
...
 	"pkg.world.dev/world-engine/cardinal/worldstage"
 	"pkg.world.dev/world-engine/sign"
 )
 
 // interface guard
-var _ engine.Context = (*worldContext)(nil)
+var _ Context = (*worldContext)(nil)
+
+//go:generate mockgen -source=context.go -package mocks -destination=mocks/context.go
+type Context interface {
```
smsunarto, line 21: this should be renamed `WorldContext` to avoid confusion with `context.Context`

smsunarto (changes requested):

cardinal/query_manager.go
```diff
@@ -2,51 +2,82 @@ package cardinal
...
+	GetRegisteredQueries() []Query
+	GetQueryByName(name string) (Query, error)
+	BuildQueryIndex(world *World)
+	GetQueryIndex() map[string]map[string]Query
 }
 
 type queryManager struct {
-	registeredQueries map[string]engine.Query
+	registeredQueries map[string]Query
+	queryIndex        map[string]map[string]Query
```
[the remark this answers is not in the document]
pyrofolium: I'm just moving things around, an index did live separate from registered queries. but yes this is better.

cardinal/world.go
pyrofolium: If we move it onto manager the manager interface it loses access to world. 

We can pass it in as `w.BuildQueryFields(w)`  but this breaks the Provider pattern on server.

smsunarto: It doesn't need access to world. GetRegisteredQueries is just calling the QueryManager anyway, you can just make a direct call to it from within querymanager [](edited-with-Axolo)

pyrofolium: missed that.

cardinal/server/handler/debug.go
```diff
@@ -26,41 +16,18 @@ type DebugStateResponse []debugStateElement
...
-						return false
-					}
-					resultElement.Components[c.Name()] = data
-				}
-				result = append(result, resultElement)
-				return true
-			},
-		)
+		var searchEachErr error
+		result, eachClosureErr, searchEachErr = world.GetDebugState()
```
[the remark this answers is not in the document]
pyrofolium: This was copied from the original code and placed into a function. I was not messing or changing the logic outside of that block of code. I can just change it to return 1 error by grabbing some additional logic from the surrounding context.

cardinal/server/handler/debug.go
pyrofolium: @smsunarto I messed with this a bit. --parseDependency and --parseInternal don't work. Using --dir didn't work for me either. 

I think type aliases work fine here even without swagger. We did follow this pattern in cardinal where we would alias other types that were coming from other packages and create a new alias in cardinal. I would recommend leaving this as is unless you have some other ideas.

cardinal/server/handler/query.go
```diff
@@ -18,35 +18,13 @@ import (
...
-		if !ok {
-			return fiber.NewError(fiber.StatusNotFound, "query name not found")
-		}
-
 		ctx.Set("Content-Type", "application/json")
-		resBz, err := query.HandleQueryRaw(wCtx, ctx.Body())
+		resBz, err := world.QueryHandler(ctx.Params("name"), ctx.Params("group"), ctx.Body())
 		if err != nil {
-			return fiber.NewError(fiber.StatusBadRequest, "encountered an error in query: "+err.Error())
+			return fiber.NewError(fiber.StatusNotFound, "query not found")
```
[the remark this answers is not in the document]
pyrofolium: I can make some specific error types here so that we can differentiate here between what error actually occured.

cardinal/types/debug.go
pyrofolium: Yeah possibly. I'll look into it.

cardinal/types/debug.go
pyrofolium: Right now it's used in cardinal and server. I'll look into consolidating it.

cardinal/types/engine/info.go
pyrofolium: I'll see what I can do.

cardinal/types/engine/info.go
pyrofolium: done

cardinal/types/engine/info.go
pyrofolium: Moved GetWorldResponse to server.

cardinal/types/debug.go
pyrofolium: nm, just moving it ended up working.

cardinal/query_manager.go
```diff
@@ -2,51 +2,82 @@ package cardinal
...
+	// Create query index
+	for _, query := range queries {
+		// Initialize inner map if it doesn't exist
+		if _, ok := m.queryIndex[query.Group()]; !ok {
+			m.queryIndex[query.Group()] = make(map[string]Query)
+		}
+		m.queryIndex[query.Group()][query.Name()] = query
+	}
+}
+func (w *World) QueryHandler(name string, group string, bz []byte) ([]byte, error) {
```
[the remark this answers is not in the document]
pyrofolium: group is no longer needed for this call. I'll make a ticket to change the url signature too.

cardinal/search.go
```diff
@@ -139,11 +147,11 @@ func (s *Search) Where(componentFilter filterFn) EntitySearch {
 
 // Each iterates over all entities that match the search.
 // If you would like to stop the iteration, return false to the callback. To continue iterating, return true.
-func (s *Search) Each(eCtx engine.Context, callback CallbackFn) (err error) {
+func (s *Search) Each(eCtx Context, callback CallbackFn) (err error) {
```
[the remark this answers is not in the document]
pyrofolium: done.

cardinal/types/search.go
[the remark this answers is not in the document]
pyrofolium: artifact from multiple refactorings. removed

cardinal/world.go
pyrofolium: removed.

cardinal/util.go
pyrofolium: When a method gets privatized and is used in a test I just dumbly wrapped it in a public function so the functionality remained the same. That's what's going on here. The test was previously using an internal method so I kept the functionality the same. That's what your seeing here with some of these public functions. 

Of course the method you mentioned here is better. I'll switch it.

cardinal/util.go
pyrofolium: As stated above. Tests that already used internal methods, I just wrapped those methods in public functions as they got privatized. I'll change these to use the alternative patterns you mentioned tmrw.

cardinal/util.go
pyrofolium: Same as above. I'll switch these tomorrow.

cardinal/types/engine/query.go
pyrofolium: Actually looks like that type isn't even used anywhere. Will remove.

cardinal/server/types/provider.go
```diff
@@ -2,20 +2,32 @@ package types
...
+	ReceiptHistorySize() uint64
+	GetTransactionReceiptsForTick(tick uint64) ([]receipt.Receipt, error)
+	EvaluateCQL(cql string) ([]types.EntityStateElement, error, error)
+	GetDebugState() (types.EntityStateResponse, error)
+	BuildQueryFields() []types.FieldDetail
+}
+
+// GetWorldResponse is a type representing the json super structure that contains
+// all info about the world.
+type GetWorldResponse struct {
```
smsunarto, line 28: What's the point of this being here instead of just being in the handler file?

cardinal/types/query.go
smsunarto, line 1: This is an overcomplicate way to have custom errors. You can just do 

```
var ErrQueryNotFound = fmt.Errorf("...")
```

If you want to provide context (i.e. query name) just use error wrapping.

```
eris.Wrap(types.ErrQueryNotFound, fmt.Sprintf("could not find query %q", e.name)
```

cardinal/types/entity.go
```diff
@@ -1,3 +1,14 @@
 package types
 
+import "encoding/json"
+
 type EntityID uint64
+
+type EntityStateResponse []EntityStateElement
+
+type EntityStateRequest struct{}
```
smsunarto, line 9: These types should just exist in the handler package. You already have `EntityStateElement` to use across packages, no?

cardinal/util.go
```diff
@@ -48,3 +52,34 @@ func isFatalError(err error) bool {
 	}
 	return true
 }
+
+func GetMessage[In any, Out any](wCtx WorldContext) (*MessageType[In, Out], error) {
```
smsunarto, line 56: Keeping a note that this should be removed. Not sure why the previous comment is marked as resolved.

cardinal/util.go
```diff
@@ -48,3 +52,34 @@ func isFatalError(err error) bool {
...
+		return nil, eris.New("error in test type assertion.")
+	}
+	return ctx.world.GetTransactionReceiptsForTick(tick)
+}
+
+func GetStoreManagerFromContext(wCtx WorldContext) gamestate.Manager {
+	return wCtx.storeManager()
+}
+
+func HandleQuery(wCtx WorldContext, query query, a any) (any, error) {
```
smsunarto, line 83: If you really need to use these methods for test, then the test should be made as an internal test with the same package name so it has access to private methods, not by rewrapping a method that should be private into a public method.

cardinal/world.go
```diff
@@ -553,6 +548,43 @@ func (w *World) UseNonce(signerAddress string, nonce uint64) error {
 	return w.redisStorage.UseNonce(signerAddress, nonce)
 }
 
+func (w *World) GetDebugState() (types.EntityStateResponse, error) {
```
smsunarto, line 551: See comments above re: types. Return type here should just be similar to EvaluateCQL
```
func (w *World) GetDebugState() ([]EntityStateElement, error) {
```

smsunarto (changes requested):

cardinal/gamestate/doc.go
```diff
@@ -69,15 +69,15 @@ number.
 
 The in-memory data model roughly matches the model that is stored in redis, but there are some differences:
 
-Components are stored as generic interfaces and not as serialized JSON.
+Data are stored as generic interfaces and not as serialized JSON.
```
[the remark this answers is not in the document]
pyrofolium: Hmm no, this is the goland refactor tool. Yeah I guess it hit some comments here.

cardinal/gamestate/manager.go
```diff
@@ -15,10 +15,10 @@ type Reader interface {
 	GetComponentForEntity(cType types.ComponentMetadata, id types.EntityID) (any, error)
 	GetComponentForEntityInRawJSON(cType types.ComponentMetadata, id types.EntityID) (json.RawMessage, error)
 
-	// Many Components One Entity
+	// Many Data One Entity
```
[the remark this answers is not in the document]
pyrofolium: Nope. The refactor tool hit some comments. I believe I changed a property on a struct from Components to Data and it hit the comment, I'll change it back.

cardinal/server/handler/query.go
```diff
@@ -18,35 +20,15 @@ import (
...
 	return func(ctx *fiber.Ctx) error {
-		query, ok := queries[ctx.Params("group")][ctx.Params("name")]
-		if !ok {
-			return fiber.NewError(fiber.StatusNotFound, "query name not found")
-		}
-
 		ctx.Set("Content-Type", "application/json")
-		resBz, err := query.HandleQueryRaw(wCtx, ctx.Body())
-		if err != nil {
+		resBz, err := world.HandleQuery(ctx.Params("name"), ctx.Body())
```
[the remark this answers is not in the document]
pyrofolium: Ah I see, ok I'll flip it back.

cardinal/search/filter/helper.go
```diff
@@ -5,7 +5,7 @@ import (
 )
 
 // MatchComponentMetadata returns true if the given slice of components contains the given component.
-// Components are the same if they have the same Name.
+// Data are the same if they have the same Name.
```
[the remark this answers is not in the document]
pyrofolium: nope, see above. I'll fix this tomorrow.

cardinal/types/entity.go
pyrofolium: Yeah, I can organize it that way. Will do this tomorrow.

cardinal/util.go
pyrofolium: Must have clicked it on accident. Will change this tomorrow.

cardinal/util.go
pyrofolium: Doing this can cause unintentional side effects. Circular dependencies for one. It's lower risk to wrap it, that's my reasoning. But I can just go with what you want here.

cardinal/world.go
pyrofolium: Got it, will change tomorrow.

cardinal/types/query.go
pyrofolium: Sure, I'll change it. tomorrow.

cardinal/cardinal_test.go
```diff
@@ -51,6 +50,12 @@ type Qux struct{}
 
 func (Qux) Name() string { return "qux" }
 
+type CardinalTestHealth struct {
```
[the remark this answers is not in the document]
pyrofolium: Got it, I'll take a look at this.

cardinal/util.go
smsunarto: The risk of 3rd party user misusing a method that is supposed to be private is a much more significant problem [](edited-with-Axolo)

pyrofolium: agree to disagree. I think it's better to be private but not absolutely critical. But this is ultimately your engine, so I will following your philosophy. [](edited-with-Axolo)

cardinal/cardinal_test.go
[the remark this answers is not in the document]
pyrofolium: Ok this was addressed. CardinalTestHealth is just unused. Removing it.

cardinal/server/types/provider.go
pyrofolium: no point. It must've been required to be here at one point, but with all the dependencies shifting around the path is open to put it into server.

pyrofolium: No wait a minute. This looks appropriate. It's under `server.types` I can just put it under it's own file or something if you want to segregate it away from provider stuff. But the package looks organizationally appropriate to me. 

Why do you want this under server.handler?  @smsunarto

smsunarto: You are not supposed to put types in its own package unless necessary (when it's used by multiple packages) in Go [](edited-with-Axolo)

pyrofolium: Ok, got it I'll move it to handler. [](edited-with-Axolo)

cardinal/types/entity.go
pyrofolium: This can't because Provider relies on EntityStateResponse. I can just kill the EntityStateResponse type type alltogether if needed. But for now I moved it to types/entity.go

cardinal/server/handler/query.go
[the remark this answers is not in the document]
pyrofolium: I think it's the other way around. WithCustomQueryGroup allows identical groups but not identical names. Registered queries are stored in a map by name and the group exists as a property on the query itself. 

See: https://github.com/Argus-Labs/world-engine/blob/8ad754dec16e084fad40db77b253d45cea8dc3d8/cardinal/query_manager.go#L27-L37

and: https://github.com/Argus-Labs/world-engine/blob/8ad754dec16e084fad40db77b253d45cea8dc3d8/cardinal/query.go#L35-L43

I could be mistaken here. Let me know how you want to proceed.

cardinal/util.go
pyrofolium: Hey @smsunarto moving relevant tests into internal causes circular dependencies with testutils and NewFixture which is a nightmare to resolve. Also the tester doesn't automatically bring the new test into the fold, it looks like it expects a specialized package cardinal_test. or some package name for _test for this to work. 

How about this, we punt this problem for now, I give HandleQuery some dangerous name like InternalHandleQuery or TestHandleQuery or something of that nature that will stop any user in his tracks from using it. I'll make a ticket for it because untangling this is looking to be a new small problem but sufficiently large enough to be it's own small ticket. 

I think to solve this problem we will have to bring NewTestFixture down into cardinal. 

Up to you. Let me know how you want to proceed.

cardinal/util.go
pyrofolium: GetMessage gets a message type which provides a message id that can be used to add a transaction. The tests are doing this. 

EachMessage does not provide access to this id or the message type. 

```
	// Entities at index 5, 10 and 50 will be updated with some values
	modifyScoreMsg, err := cardinal.GetMessage[*ModifyScoreMsg, *EmptyMsgResult](wCtx)
	assert.NilError(t, err)
	tf.AddTransaction(
		modifyScoreMsg.ID(), &ModifyScoreMsg{
			PlayerID: ids[5],
			Amount:   105,
		},
	)
	tf.AddTransaction(
		modifyScoreMsg.ID(), &ModifyScoreMsg{
			PlayerID: ids[10],
			Amount:   110,
		},
	)
	tf.AddTransaction(
		modifyScoreMsg.ID(), &ModifyScoreMsg{
			PlayerID: ids[50],
			Amount:   150,
		},
	)
```

cardinal/util.go
smsunarto: sure [](edited-with-Axolo)

cardinal/server/handler/query.go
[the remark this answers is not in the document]
smsunarto: Thats just an invalid behavior with isQueryNameUnique. As I previously mentioned, just make it such that there is making of group -> name -> query. That's the intended behavior [](edited-with-Axolo)

cardinal/types/entity.go
smsunarto: The provider relies on EntityStateResponse because you made `GetDebugState` returns `types.EntityStateResponse` when it could have just returned `[]EntityStateElement`. You are just getting confused by your own footguns. [](edited-with-Axolo)

cardinal/server/handler/query.go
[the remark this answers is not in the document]
pyrofolium: I'm more referring to this:

```
m.registeredQueries[name] = query 
```

the storage for queries makes it so all queries are indexed by a unique name. If we want to change this I recommend  moving this to another ticket as it will change how things work under the hood. This context thing is more of an aesthetic api change and doing this imo will make the changes in this PR too big and less focused. 

I think this could be a good plan: I'll change the url and method signatures back to include group, and I'll create a ticket for changing the whole thing to `group->name->query` for the long term. 

Let me know if you're cool with the above.

cardinal/util.go
pyrofolium: The idiomatic way to handle this is to not have the test rely on GetMessage. Changing it to EachMessage doesn't fully work yet. So this isn't that straightforward of an edit. Again I suggest punting this to another ticket. This PR is getting far to large already. lmk if this works for you @smsunarto.

smsunarto: fine [](edited-with-Axolo)

cardinal/types/entity.go
pyrofolium: Seems like your stretched. I'll make a decision here to kill the type and merge it in, lmk if you want to reverse it.

smsunarto: I already pushed the code and cleaned it myself.

Again, you are not actually reading the comment I left or asking clarification questions about it. You just ended up confusing yourself when what you did ends up being exactly what I'm describing. [](edited-with-Axolo)

cardinal/server/handler/query.go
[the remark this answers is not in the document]
pyrofolium: scott said this is ok in an in person discussion.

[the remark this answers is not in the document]
pyrofolium: This is done.

cardinal/gamestate/doc.go
smsunarto, line 1: Accidental change?

cardinal/gamestate/manager.go
smsunarto, line 1: Accidental change?

cardinal/testutils/world.go
smsunarto, line 1: Accidental change?

smsunarto (changes requested):

cardinal/gamestate/manager.go
pyrofolium: Oh more of this. Ok. I'll do a big search on that

smsunarto (approved):
lgtm, looks nice

pyrofolium:
All commits need to be signed with a GPG key which did not occur here in this PR. Rather then modify the commits and potentially mangle things or potentially mess with the conversations that occured in this PR, I am creating a new PR where all the commits are squashed into a single commit that is signed.

smsunarto:
@pyrofolium can we close this?

pyrofolium:
Yea

PR #758, opened 2024-05-18
feat: add ability for cardinal to read toml config file

zulkhair:
Closes: WORLD-1129

## Overview
Add ability for cardinal to read toml config file, the file can be read from :
- flag `--CARDINAL_CONFIG={filepath}`
- env var `CARDINAL_CONFIG`
- search `world.toml` file in current dir & parrent dir

## Brief Changelog
- Add func to read toml file in config.go and put it to env var

## Testing and Verifying

- Added unit test
- Test manually using e2e package


zulkhair:
* **#758** <a href="https://app.graphite.dev/github/pr/Argus-Labs/world-engine/758?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @zulkhair and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

smsunarto (changes requested):
hmm, this looks kinda complicated. we should look into https://github.com/spf13/viper to abstract this for us

cardinal/config.go
```diff
@@ -43,46 +51,67 @@ var (
...
+		// Unmarshal the [cardinal] section into the WorldConfig struct
+		if err := viper.Sub("cardinal").Unmarshal(&cfg); err != nil {
+			log.Warn().Err(err).Msg("failed to unmarshal config file")
+			log.Info().Msg("using default config values")
+		}
+	}
+
+	// Override config values with environment variables
+	// This is done after reading the config file to allow for environment variable overrides
 	if err := config.FromEnv().To(&cfg); err != nil {
```
smsunarto, line 115: We should be able to drop JeremyLoy/config altogether since viper can also read from environment variable no?

cardinal/config.go
```diff
@@ -19,6 +23,10 @@ const (
 	DefaultCardinalLogLevel          = "info"
 	DefaultRedisAddress              = "localhost:6379"
 	DefaultBaseShardSequencerAddress = "localhost:9601"
+
+	// Toml config file related
+	tomlConfigFileEnvVariable = "TOML_CONFIG_FILE"
```
smsunarto, line 28: rename the env to `CARDINAL_CONFIG`

cardinal/config.go
```diff
@@ -160,3 +189,52 @@ func (w *WorldConfig) setLogger() error {
 
 	return nil
 }
+
+func setupViper() {
+	once.Do(func() {
+		// Bind the command-line flags to Viper
+		pflag.String(tomlConfigFileEnvVariable, "", "path to the toml config file")
```
smsunarto, line 196: Make sure to properly capitalize the message here since it's a user facing message

cardinal/config.go
```diff
@@ -160,3 +189,52 @@ func (w *WorldConfig) setLogger() error {
...
+		pflag.Parse()
+		err := viper.BindPFlags(pflag.CommandLine)
```
smsunarto, line 198: nit: if the return value is just an error, you can inline this into the if block

```
if err := viper.BindPFlags(pflag.CommandLine); if err != nil {
	log.Debug().Err(err).Msg("...")
}
```

cardinal/config.go
```diff
@@ -160,3 +189,52 @@ func (w *WorldConfig) setLogger() error {
...
+	viper.SetConfigType("toml")  // REQUIRED if the config file does not have the extension in the name
+
+	// Find the toml file from the flag and env variable
+	tomlFilePath := viper.GetString(tomlConfigFileEnvVariable)
+	if tomlFilePath == "" {
+		// Find the toml file in the current and parent directories
+		viper.AddConfigPath(".") // look for config in the working directory
+
+		// Add parent directory to the search path
+		parentDir, err := os.Getwd()
```
smsunarto, line 219: you need to run a check first whether you've already found a world.toml file in the current directory. if there is already a config file in the current directory, you don't need to check the parent directory.

cardinal/config.go
```diff
@@ -160,3 +189,52 @@ func (w *WorldConfig) setLogger() error {
...
+		if err != nil {
+			log.Warn().Err(err).Msg("failed to get current directory for toml file search")
+		} else {
+			parentDir = filepath.Dir(parentDir)
+			viper.AddConfigPath(parentDir)
+		}
+	} else {
+		// Use the toml file from the flag or env variable
```
smsunarto, line 227: the priority for the config path should be flag -> env -> default path (world.toml in current directory or parent's directory)

cardinal/config.go
smsunarto, line 1: nit: make sure all logs are capitalized

smsunarto (changes requested):

cardinal/config.go
zulkhair: updated, replaced by viper

cardinal/config.go
zulkhair: updated

cardinal/config.go
zulkhair: updated

cardinal/config.go
zulkhair: updated

cardinal/config.go
zulkhair: updated

cardinal/config.go
zulkhair: flag and env precedence is handled by viper. it should be already like flag > env > default (reference : https://github.com/spf13/viper#why-viper).
if flag and env is not set then we search in current dir and parrent dir.

cardinal/config.go
zulkhair: updated

cardinal/config_test.go
```diff
@@ -266,3 +270,84 @@ func defaultConfigWithOverrides(overrideCfg WorldConfig) WorldConfig {
...
+	t.Setenv(tomlConfigFileEnvVariable, tomlFilePath)
+
+	cfg, err := loadWorldConfig()
+	assert.NilError(t, err)
+	assert.Equal(t, "my-world-e2e", cfg.CardinalNamespace)
+}
+
+func TestWorldConfig_loadWorldConfigUsingTomlFlag(t *testing.T) {
+	// Set up command-line arguments
+	os.Args = []string{"cmd", "--TOML_CONFIG_FILE=../e2e/testgames/world.toml"}
```
smsunarto, line 285: didn't we change this to `CARDINAL_CONFIG`? i wonder why this test still passes

cardinal/config.go
```diff
@@ -19,6 +23,10 @@ const (
 	DefaultCardinalLogLevel          = "info"
 	DefaultRedisAddress              = "localhost:6379"
 	DefaultBaseShardSequencerAddress = "localhost:9601"
+
+	// Toml config file related
+	tomlConfigFileEnvVariable = "CARDINAL_CONFIG"
```
smsunarto, line 28: nit: `configFilePathEnvVariable`

cardinal/config.go
```diff
@@ -19,6 +23,10 @@ const (
...
+	defaultTomlConfigFilename = "world.toml"
```
smsunarto, line 29: `defaultConfigFileName`

cardinal/config.go
```diff
@@ -43,47 +51,68 @@ var (
...
 	cfg := defaultConfig
 
-	if err := config.FromEnv().To(&cfg); err != nil {
+	// Setup Viper for world toml config file
+	setupViper()
+
+	// Read the config file
+	if err := viper.ReadInConfig(); err != nil {
+		log.Warn().Err(err).Msg("Failed to read config file")
+		log.Info().Msg("Using default config values")
```
smsunarto, line 104: you can just collapse this message into the warn log

cardinal/config.go
```diff
@@ -160,3 +189,73 @@ func (w *WorldConfig) setLogger() error {
...
+		tag := field.Tag.Get("mapstructure")
+		if tag != "" {
+			if err := viper.BindEnv(tag); err != nil {
+				log.Warn().Err(err).Str("field", field.Name).Msg("Failed to bind env variable")
+			}
+		}
+	}
+}
+
+func searchTomlInDirectories() {
```
smsunarto, line 236: imo, for the next 3 functions you can just inline them. 

this is too "clean code" that it ended up just being more confusing because the program flow jumps around.

cardinal/config.go
```diff
@@ -43,47 +51,68 @@ var (
...
+	// Setup Viper for world toml config file
+	setupViper()
+
+	// Read the config file
+	if err := viper.ReadInConfig(); err != nil {
+		log.Warn().Err(err).Msg("Failed to read config file")
+		log.Info().Msg("Using default config values")
+	} else {
+		// Unmarshal the [cardinal] section into the WorldConfig struct
+		if err := viper.Sub("cardinal").Unmarshal(&cfg); err != nil {
```
smsunarto, line 107: Do you really need this? This looks identical with the unmarshal below. How do you know the one below doesn't also read from the config file?

smsunarto (changes requested):

cardinal/config.go
```diff
@@ -160,3 +189,73 @@ func (w *WorldConfig) setLogger() error {
...
+		pflag.String(tomlConfigFileEnvVariable, "", "Path to the TOML config file")
+		pflag.Parse()
+		if err := viper.BindPFlags(pflag.CommandLine); err != nil {
+			log.Debug().Err(err).Msg("Failed to bind command-line flags to Viper")
+			// Continue even if the binding fails
+		}
+	})
+
+	// Bind the environment variables to Viper
+	viper.AutomaticEnv()
```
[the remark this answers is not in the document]
zulkhair: this is for `CARDINAL_CONFIG` env var.
i'll remove this and replace with `viper.BindEnv` in the bottom

cardinal/config_test.go
zulkhair: ahh, it's because viper value still there after other test case execution. need to reset the viper value after running the test cases.

cardinal/config.go
zulkhair: updated

cardinal/config.go
zulkhair: updated

cardinal/config.go
zulkhair: updated

cardinal/config.go
zulkhair: yes, the one above `viper.Sub("cardinal").Unmarshal(&cfg)` is for marshaling from toml file with `cardinal` section/table in toml file.

and the bottom one `viper.Unmarshal(&cfg)` is read from the env var, but also read from the outer section/table in toml file, if the env var not set and the key=value is exist on the config file.

cardinal/config.go
zulkhair: just got blamed by linter for nested if, so i'm separate it to the 3 func. 😁
i'll make another approach, or maybe i'll just use //nolint for this

smsunarto: just use nolint if the linter is being stupid [](edited-with-Axolo)

cardinal/config.go
smsunarto: this feels clunky, is there a way to make this cleaner? [](edited-with-Axolo)

cardinal/config.go
zulkhair: updated using nolint

cardinal/config.go
zulkhair: Tidied up the code.

Still need two unmarshal because values from env is located at the parent of viper config, and the values from config file (toml) is located at sub/child path (cardinal) of viper config.

If we want to exec unmashal only once, we need to move the value from the sub/child to the parent of the config or vice versa. But i think it's inefficient.

smsunarto (approved):

PR #769, opened 2024-05-31
refactor(cardinal): remove InternalHandleQuery

pyrofolium:
Closes: WORLD-1123

The overall goal is to remove `InternalHandleQuery`which is a helper function used in tests. 

InternalHandleQuery is part of cardinal and takes in a query and uses the private method query.handleQuery

To get rid of InternalHandleQuery, all tests that use that function are moved to the cardinal package. Once moved the tests can directly use query.handleQuery without the need of calling InternalHandleQuery. 

When tests are moved to cardinal, it causes a circular dependency with the testutils/world.go which contains the Fixture stuff. 

I just moved all the Fixture stuff into cardinal as well.


zulkhair (approved):
The changes made are quite simple and clear. and the functionality should already checked by GH action (test & lint). no concern.

smsunarto (approved):

PR #771, opened 2024-05-31
refactor(cardinal): create interfaces for component manager

pyrofolium:
Closes: WORLD-861

Create an manager interface on top of the component manager. 

If I call it ComponentManager, the linter will complain about stuttering. Hence the "component manager" is actually called "component.Manager" and instantiating a new Manager on world looks something like this:

```
World {
     Manager: NewComponentManager.... 
}
```

It's a bit awkward. Let me know if there's any better ways to do it. Or if you want to go at it the other way and just stutter the name.


cardinal/world.go
```diff
@@ -123,13 +123,13 @@ func NewWorld(opts ...WorldOption) (*World, error) {
...
-		MessageManager:   newMessageManager(),
-		SystemManager:    newSystemManager(),
-		componentManager: component.NewManager(&redisMetaStore),
-		QueryManager:     newQueryManager(),
-		router:           nil, // Will be set if run mode is production or its injected via options
-		txPool:           txpool.New(),
+		worldStage:     worldstage.NewManager(),
+		MessageManager: newMessageManager(),
+		SystemManager:  newSystemManager(),
+		Manager:        component.NewManager(&redisMetaStore),
```
smsunarto, line 129: this should be ComponentManager

smsunarto (approved):

smsunarto (changes requested):
see above

cardinal/world.go
pyrofolium: It's changed. and I added a linter exception so that the linter will ignore the stuttering.

smsunarto (approved):

PR #782, opened 2024-06-22
feat(cardinal): deterministic wrapper around math/rand so users can use it. 

pyrofolium:
changes:

- new 'Rand' method on world context that returns 'math/rand/Rand'
- new property on world context that holds the Rand struct
- A new seeded rand is generated on every tick to ensure determinism. 
- BIG change: timestamp is changed from milliseconds to nanoseconds because tests indicate that many ticks can actually execute faster than a millisecond. When this happens an equivalent random number is generated across ticks.


pyrofolium:
I'm converting this back to draft. There are potential issues with recovery. When transactions are re-executed the timestamp is not deterministic and it's not in nano either.

smsunarto:
> •  BIG change: timestamp is changed from milliseconds to nanoseconds because tests indicate that many ticks can actually execute faster than a millisecond. When this happens an equivalent random number is generated across ticks.

This is an issue with your test code. Cardinal ticks have an interval of 1 second by default and can be configured using `cardinal.WithTickChannel`. Sub-millisecond ticks in Cardinal are not realistic in a production environment, so your test code just needs to incorporate this interval instead.

pyrofolium:
> > •  BIG change: timestamp is changed from milliseconds to nanoseconds because tests indicate that many ticks can actually execute faster than a millisecond. When this happens an equivalent random number is generated across ticks.
> 
> This is an issue with your test code. Cardinal ticks have an interval of 1 second by default and can be configured using `cardinal.WithTickChannel`. Sub-millisecond ticks in Cardinal are not realistic in a production environment, so your test code just needs to incorporate this interval instead.

It's not realistic, but it could happen in some edgecase, I think it's better to code for safety here. Do you see any potential downsides for switching this to nano? If it's not a harmless change I will switch it backl.

smsunarto:
Nano isn't legible and makes using timestamps for other purposes a chore. It's also not reliable because not all OS have nanosecond resolution on timestamp (which will lead us back to collision). It's always better for the test fixture to just emulate the real-world use case as much as possible.

pyrofolium:
milli isn't that legible either. Also realistically anything running world engine should have nano. Ime unit tests should test the "unit" and the unit should be functionally robust independent of the context because the unit can be used in other contexts.

That's just my 2 cents. I'll change it back, and I'll change the unit test to account for a bigger internval.

smsunarto:
Everyone knows how many milliseconds a second is on top of their head; it's also widely used to measure network latency.

The moment you use Cardinal's test fixture, then it's an integration test, and it should take into account how Cardinal is used. If you want to write a proper isolated unit test then go for it, but then it should have no side effect from cardinal and no changes to cardinal should be needed.

pyrofolium:
Can't do the unit test, because the code isn't segregated enough that way. I guess my point is that you should write modules as close as you can as if they were going to be tested against actual unit tests. But practically speaking, ime no one is actually that strict for testing.

As for the millisecond thing, personally I can't read a unix timestamp whether it's milli or nano or even seconds. I can agree though if milli is a commonly expected unit of time for the use case. I'll just change it back to milli.

smsunarto:
You have to take into consideration how it is used; it's used extensively to do duration/interval math in the game systems (want an action to take 1.5s? that's 1500ms; if it's nanosecond then it would be 1500000000ns, which is significantly more unpleasant to work with degrees of precision that will just be pointless since there's no sub-millisecond ticks possible)

pyrofolium:
I'm switching to milliseconds. There's going to be some other changes on top of this to handle recovery too.

smsunarto:
What recovery? The test fixtures don't enable router recovery, so I'm not sure where you are seeing this recovery issue. Determinism on the timestamp only matters when you are recovering a finalized tick and rebuilding the cardinal state, not if your program crashes and you rerun it.

pyrofolium:
Recovery has nothing to do with the tests. During recovery the recovery functionality creates a new timestamp that is different from the timestamp the previous txs were executed on.

smsunarto:
When you say recovery, which method are you talking about?

pyrofolium:
```
recoverAndExecutePendingTxs
```

pyrofolium:
not a method, a function

pyrofolium:
my bad it's a method

smsunarto:
This is exactly what i'm telling you above, this method is not the same as the router recovery aka RecoverFromChain. This does not need to be deterministic because the tick was never finalized and sequenced to begin with

pyrofolium:
fine I'll take it out

pyrofolium:
it's ready

cardinal/world_context.go
```diff
@@ -102,6 +114,10 @@ func (ctx *worldContext) Logger() *zerolog.Logger {
 	return ctx.logger
 }
 
+func (ctx *worldContext) Rand() *rand.Rand {
```
smsunarto, line 117: We need to add a conditional check here to ensure that this is only used with `newWorldContextForTick` similar to how there is a read-only check. 

This is to prevent people from using this via `NewWorldContext` and `NewReadOnlyWorldContext` (used by queries) that will produce invalid behaviors for the timestamp since it's not executing within a tick.

cardinal/world_context.go
```diff
@@ -60,6 +63,12 @@ type worldContext struct {
 	txPool   *txpool.TxPool
 	logger   *zerolog.Logger
 	readOnly bool
+	rand     *rand.Rand
+}
+
+func createRngFromWorld(world *World) *rand.Rand {
+	//nolint:gosec // we require manual in the rng which crypto/rand doesn't have, but math/rand does.
+	return rand.New(rand.NewSource(int64(world.timestamp.Load())))
```
smsunarto, line 71: nit: i would just inline this into `newWorldContextForTick`. The other world context should not have a rand object, see my previous comment.

smsunarto (changes requested):

smsunarto (approved):

PR #806, opened 2024-11-01
refactor: separate validator from rest of server code

ezavada:
Closes: WORLD-1237

## Overview

Separates out the validation code into SignatureValidator object in a server/validator subpackage to reduce complexity and improve separation of concerns, and makes it independently testable.

## Brief Changelog

- Separates out the validation code into SignatureValidator object in a server/validator subpackage
- ValidateTransactionTTL, which checks the timestamps and the hash
- ValidateTransactionSignature, which checks the signature
- The validation package handles all the management of the cache and greatly simplifies the PostTransaction calls in server
- There is now a validator_test.go file that tests all the validation functionality without any need for setting up Cardinal.

## Testing and Verifying

All CI tests have been run and passed, including the new cardinal/server/validator/validator_test.go suite

## Release Notes

- **New Features**
	- Introduced a new `SignatureValidator` for improved transaction validation, enhancing security against replay attacks.
	- Added a method to retrieve signer addresses based on persona tags.
	- Added utility functions for timestamp management in the signing process.
	- Enhanced server configuration options, including message expiration and hash cache size management.
	- Introduced documentation for enabling monitoring in Nakama using Jaeger and Prometheus.
	- Added a new document detailing monitoring setup for Nakama.

- **Bug Fixes**
	- Enhanced error handling for transaction validation, ensuring more robust feedback for various failure scenarios.

- **Documentation**
	- Updated documentation to reflect changes in transaction processing and validation mechanisms.
	- Added a hyperlink for further information regarding the `DA_AUTH_TOKEN` configuration parameter.
	- Changed `ENABLE_ALLOWLIST` setting from a string to a boolean in the Nakama configuration documentation.
	- Added a new page for monitoring Nakama in the documentation structure.

- **Tests**
	- Added comprehensive tests for the new validation logic and timestamp handling to ensure reliability and accuracy.


cardinal/server/handler/tx.go
```diff
@@ -1,48 +1,24 @@
...
 
-type SignatureVerification struct {
-	IsDisabled               bool
-	MessageExpirationSeconds int
-	HashCacheSizeKB          int
-	Cache                    *freecache.Cache
-}
-
 type Transaction = sign.Transaction
+type SignatureValidator = validator.SignatureValidator
```
smsunarto, line 21: I generally prefer to not do type alias like this; it makes it less obvious where a certain object/struct is imported from.

cardinal/server/handler/tx.go
```diff
@@ -69,15 +45,15 @@ func PostTransaction(
 		}
 
 		// extract the transaction from the fiber context
-		tx, fiberErr := extractTx(ctx, verify)
+		tx, fiberErr := extractTx(ctx, validator)
```
smsunarto, line 48: It's idiomatic in Go to just reuse the err variable. You don't have to provide a different name for each error.

In this case, you can just do
`tx, err := extractTx(ctx, validator)`

cardinal/server/handler/tx.go
```diff
@@ -69,15 +45,15 @@ func PostTransaction(
...
 		if fiberErr != nil {
 			return fiberErr
 		}
 
-		// Validate the transaction
-		if err := validateTx(tx); err != nil {
-			log.Errorf("message %s has invalid transaction payload: %v", tx.Hash.String(), err)
-			return fiber.NewError(fiber.StatusBadRequest, "Bad Request - invalid payload")
+		// make sure the transaction hasn't expired
+		if validationErr := validator.ValidateTransactionTTL(tx); validationErr != nil {
```
smsunarto, line 54: Same here, just use `err`

cardinal/server/handler/tx.go
```diff
@@ -87,31 +63,18 @@ func PostTransaction(
...
-				log.Errorf("Signature validation failed for message %s: %v", tx.Hash.String(), err)
-				return fiber.NewError(fiber.StatusUnauthorized, "Unauthorized - invalid signature")
-			}
+		// there's a special case for the CreatePersona message
+		var signerAddress string
+		if msgType.Name() == personaMsg.CreatePersonaMessageName {
+			// don't need to check the cast bc we already validated this above
+			createPersonaMsg, _ := msg.(personaMsg.CreatePersona)
+			signerAddress = createPersonaMsg.SignerAddress
+		}
```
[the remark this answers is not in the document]
smsunarto: +1

cardinal/server/option.go
```diff
@@ -33,7 +33,7 @@ func DisableSignatureVerification() Option {
 // must be synchronized
 func WithMessageExpiration(seconds int) Option {
 	return func(s *Server) {
-		s.verify.MessageExpirationSeconds = seconds
+		s.config.messageExpirationSeconds = seconds
```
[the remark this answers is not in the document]
smsunarto: +1, FWIW, we should just make it 0 if the number is less than 0 for whatever reason. Or can just make it uint.

cardinal/server/option.go
```diff
@@ -42,6 +42,6 @@ func WithMessageExpiration(seconds int) Option {
 // This setting is ignored if the DisableSignatureVerification option is used
 func WithHashCacheSize(sizeKB int) Option {
 	return func(s *Server) {
-		s.verify.HashCacheSizeKB = sizeKB
+		s.config.messageHashCacheSizeKB = sizeKB
```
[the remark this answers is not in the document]
smsunarto: +1, should probably enforce a minimum + log a warn if the cache size option is too small.

cardinal/server/validator/validator.go
```diff
@@ -0,0 +1,180 @@
...
+const ttlMaxFutureSeconds = 2 // this is how many seconds in the future a message is allowed to be timestamped
+// this allows for some degree of clock drift. It's safe enought to accept a message that's stamped from the near
+// future because we will still keep it in the hash cache and prevent it from being a replay attack vector. However,
+// we don't want to take messages from an unlimited amount of time into the future since they could cause our hash
+// cache to overflow
+
+const bytesPerKb = 1024
+
+var (
+	ErrNoPersonaTag     = errors.New("persona tag is required")
```
smsunarto, line 34: Use eris.New so the error shows up with stacktrace

cardinal/server/validator/validator.go
```diff
@@ -0,0 +1,180 @@
...
+
+	"github.com/coocood/freecache"
+	"github.com/ethereum/go-ethereum/common" // for hash
+	"github.com/rotisserie/eris"
+
+	"pkg.world.dev/world-engine/sign"
+)
+
+type SignerAddressProvider interface {
+	GetSignerAddressForPersonaTag(personaTag string) (addr string, err error)
```
smsunarto, line 17: Just use WorldProvider directly imo.

cardinal/server/validator/validator.go
smsunarto, line 1: We don't really need this to be a separate package. It can go under the `server` package: `cardinal/server/validator.go` -- Go will punish you with circular dependency issues when you try to split off packages that are tightly coupled.

cardinal/server/validator/validator.go
```diff
@@ -0,0 +1,180 @@
...
+		cache:                    nil,
+		signerAddressProvider:    provider,
+	}
+	if !disabled {
+		validator.cache = freecache.NewCache(validator.HashCacheSizeKB * bytesPerKb)
+	}
+	return &validator
+}
+
+type Transaction = sign.Transaction
```
smsunarto, line 87: Don't type alias

cardinal/server/validator/validator.go
```diff
@@ -0,0 +1,180 @@
...
+	cache                    *freecache.Cache
+	signerAddressProvider    SignerAddressProvider
+}
+
+type ValidationError interface {
+	error
+	GetLogMessage() string
+	GetStatusCode() int
+}
+type validationError struct {
```
smsunarto, line 58: This custom error type is not necessary and we generally try to avoid since the basic error primitive is more than sufficient.

1. I'd prefer not to control the status code from the validator because this breaks separation of concern; the validator doesn't necessarily have to be consumed by an HTTP server. Instead, in the server handler, use `errors.Is`/`eris.Is` to check for sentinel errors (the one you declared above) and handle the HTTP response accordingly there.

2. Instead of using `LogMsg`, you can just `eris.Wrap` your sentinel error to provide additional context to the error which you can log/print out.

By doing these 2 things, you should be able to eliminate this custom error type.

cardinal/world_persona.go
```diff
@@ -11,6 +11,12 @@ import (
 	"pkg.world.dev/world-engine/cardinal/types"
 )
 
+// implements the validator.SignerAddressProvider interface
+func (w *World) GetSignerAddressForPersonaTag(personaTag string) (addr string, err error) {
```
smsunarto, line 15: I'd prefer not to add this extra method; it's ambiguous with `w.GetSignerForPersonaTag` and doesn't provide much abstraction.

sign/sign.go
```diff
@@ -47,14 +47,21 @@ type Transaction struct {
 	Body       json.RawMessage `json:"body" swaggertype:"object"` // json string
 }
 
+// returns a sign compatible timestamp for the current wall time
```
smsunarto, line 50: Go convention is to prefix function/method comment with the function/method name. Otherwise, godoc won't recognize it.

`TimestampNow ...`

sign/sign.go
```diff
@@ -47,14 +47,21 @@ type Transaction struct {
...
 func TimestampNow() int64 {
 	return time.Now().UnixMilli() // millisecond accuracy on timestamps, easily available on all platforms
 }
 
+// returns a sign compatible timestamp for the time passed in
```
smsunarto, line 55: same here.

sign/sign.go
```diff
@@ -47,14 +47,21 @@ type Transaction struct {
...
 func TimestampAt(t time.Time) int64 {
 	return t.UnixMilli()
 }
 
+// returns a GoLang time from a sign compatible timestamp
```
smsunarto, line 60: same here.

smsunarto (changes requested):

cardinal/server/option.go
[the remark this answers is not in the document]
ezavada: FreeCache itself has a minimum cache size of 512k. I added a comment to that effect, let me know if you still think this needs to issue a warning.

sign/sign.go
```diff
@@ -47,14 +47,21 @@ type Transaction struct {
...
+func Timestamp(t int64) time.Time {
+	return time.UnixMilli(t)
+}
+
```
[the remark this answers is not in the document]
ezavada: validator depends on sign, and the sign package needs to generate the timestamp in order to sign it. Therefore, moving the timestamp functions into validator would create a circular dependency,

cardinal/server/validator/validator.go
```diff
@@ -0,0 +1,180 @@
...
+	return false, err
+}
+
+// validateSignature validates that the signature of transaction is valid
+func (validator *SignatureValidator) validateSignature(tx *Transaction, signerAddr string) error {
+	if tx.Namespace != validator.namespace {
+		return eris.Wrap(ErrWrongNamespace, fmt.Sprintf("expected %q got %q", validator.namespace, tx.Namespace))
+	}
+	return eris.Wrap(tx.Verify(signerAddr), "")
+}
```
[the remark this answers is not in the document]
ezavada: Great suggestion!

cardinal/server/validator/validator.go
ezavada: I tried that and ended up with circular references because then server/handler (for tx.go) needs to depend on server, but server needs to depend on handler. By moving it into it's own package I was able to have both server and handler depend on validator without any conflict.

sign/sign.go
ezavada: This would be a good thing for CodeRabbit to pick up on. CodeRabbit, are you listening?

ezavada: @coderabbitai [](edited-with-Axolo)

smsunarto (approved):

PR #807, opened 2024-11-06
fix: Change context to match what Dockerfile is expecting

ezavada:
Change context to match what Dockerfile is expecting


rmrt1n (approved):
Saw the successful action, lgtm

PR #880, opened 2026-03-31
feat: add Box2D physics2d plugin with CGO

AnthonyWadham:
# Add 2D Physics Plugin with Box2D Integration

This pull request introduces a comprehensive 2D physics plugin for Cardinal using Box2D as the simulation backend.

## Components

- **Transform2D**: World-space position and rotation
- **Velocity2D**: Linear and angular velocity 
- **Rigidbody2D**: Body type (static/dynamic/kinematic) and simulation parameters
- **Collider2D**: Compound collider with multiple shapes (circle, box, polygon, chain)
- **ActiveContacts**: Singleton component tracking active contact pairs for event deduplication

## Core Systems

- **InitPhysicsSystem**: Creates Box2D world and performs initial ECS→Box2D rebuild
- **ReconcilePhysicsSystem**: Incrementally syncs ECS changes to Box2D using shadow-copy diffing
- **PhysicsStepSystem**: Steps the simulation and emits contact/trigger events

## Contact Events

- **ContactBeginEvent/ContactEndEvent**: Non-sensor collision events
- **TriggerBeginEvent/TriggerEndEvent**: Sensor overlap events
- Proper event deduplication after world rebuilds using persisted contact state

## Query API

- **Raycast**: Cast rays through the physics world
- **OverlapAABB**: Find fixtures overlapping an axis-aligned bounding box
- **CircleSweep**: Sweep a circle along a path for continuous collision detection

## Runtime Management

- Shadow-copy reconciliation for efficient incremental updates
- Structural vs mutable change detection (fixture recreation vs in-place updates)
- Deterministic rebuild ordering and contact event handling
- Proper cleanup of orphaned bodies and contact pairs

The plugin integrates seamlessly with Cardinal's ECS architecture while providing a complete 2D physics simulation with collision detection, rigid body dynamics, and comprehensive query capabilities.


AnthonyWadham:
* **#880** <a href="https://app.graphite.com/github/pr/Argus-Labs/world-engine/880?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#882](https://github.com/Argus-Labs/world-engine/pull/882) <a href="https://app.graphite.com/github/pr/Argus-Labs/world-engine/882?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#883](https://github.com/Argus-Labs/world-engine/pull/883) <a href="https://app.graphite.com/github/pr/Argus-Labs/world-engine/883?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/Argus-Labs/world-engine/880?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/Argus-Labs?ref=merge-queue-instructions-in-stack-comment&prId=880).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

pkg/plugin/physics2d/system/init.go
```diff
@@ -0,0 +1,60 @@
...
+type physicsSingletonArchetype = cardinal.Exact[struct {
+	Tag            cardinal.Ref[physicscomp.PhysicsSingletonTag]
+	ActiveContacts cardinal.Ref[physicscomp.ActiveContacts]
+}]
+
+// InitPhysicsSystemState runs once at world init: FullRebuildFromECS from current ECS entities.
+type InitPhysicsSystemState struct {
+	cardinal.BaseSystemState
+	cardinal.Contains[physicsArchetype]
+	Singleton physicsSingletonArchetype
```
rmrt1n: suggestion:
- we probably shouldn't use the embedded struct shorthand for public examples as they're less clear (unnamed `cardinal.Contains[physicsArchetype]`)
- we should decide on a naming convention. here "archetype" is used for different concepts. "archetype" fits the type inside the searches (`Exact`/`Contains`) more, and in rampage search type aliases are usually postfixed w `Search`, tho we don't have to follow their convention.

pkg/plugin/physics2d/system/singleton.go
```diff
@@ -0,0 +1,19 @@
...
+import physicscomp "github.com/argus-labs/world-engine/pkg/plugin/physics2d/component"
+
+// ensurePhysicsSingleton creates the plugin singleton entity if none exists. Call from Init
+// and PreUpdate reconcile so snapshot restore (which may skip Init) still has persisted
+// ActiveContacts storage before the first physics step.
+//
+// singleton must be a pointer: Exact embeds search, whose fields slice points at Refs inside
+// the system's result struct. Passing Exact by value makes Create attach those original Refs
+// but return a copy of result whose Refs were never attached (nil world → panic on Set).
+func ensurePhysicsSingleton(singleton *physicsSingletonArchetype) {
```
rmrt1n: nit: inline this in `InitPhysicsSystem` as it's only used there. also:
- replace the iter loop with `Iter().Single()`.
- `Create` initializes the entity with zero values for its components, so you don't need to set them again.

pkg/plugin/physics2d/system/init.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+// InitPhysicsSystem creates the singleton entity (if absent), then builds the Box2D world
+// and bodies from ECS. Runs on cardinal.Init.
+func InitPhysicsSystem(state *InitPhysicsSystemState) {
+	ensurePhysicsSingleton(&state.Singleton)
```
rmrt1n, line 50: issue: ok I just realized a big problem w init systems and snapshots. they won't be run after a snapshot restore, so this will never run and recreate the in-memory box2d runtime...

suggestion: in `cardinal.go`, run the `w.world.Init()` after the restore. the downside (to me at least), is that now, init systems must be idempotent, i.e. rerunning it produces the same result, e.g.:

from this:

```go
func MyInitSystem(state *MyInitSystemState) {
    _, _, err := state.Singleton.Iter().Single()
    assert(err == nil)

    // Create singleton.

    // Setup box2d (or other side effects).
}
```

to this:
```go
func MyInitSystem(state *MyInitSystemState) {
    _, _, err := state.Singleton.Iter().Single()
    if err != nil {
        // Returned err must indicate there's no singleton.
        assert(eris.Is(err, cardinal.ErrSingleNoResult))

        // Create singleton.
    }

    // Singleton already exists (from create above or snapshot restore).

    // Setup box2d (or other side effects).
}
```

chore: after merging we should tag a release and update the init systems in the templates, lobby, and rampage.

pkg/plugin/physics2d/system/init.go
```diff
@@ -0,0 +1,60 @@
...
+
+	cfg := stepConfig()
+	g := box2d.MakeB2Vec2(cfg.Gravity.X, cfg.Gravity.Y)
+	entries := gatherRebuildEntries(state.Iter())
+	if err := internal.FullRebuildFromECS(g, entries); err != nil {
```
rmrt1n: suggestion: crash if this fails.

pkg/plugin/physics2d/system/reconcile.go
```diff
@@ -0,0 +1,36 @@
...
+	ensurePhysicsSingleton(&state.Singleton)
+	entries := gatherRebuildEntries(state.Iter())
+	cfg := stepConfig()
+	g := box2d.MakeB2Vec2(cfg.Gravity.X, cfg.Gravity.Y)
+	if rt.World == nil {
+		if err := internal.FullRebuildFromECS(g, entries); err != nil {
+			state.Logger().Error().Err(err).Msg("physics2d: FullRebuildFromECS failed (nil world recovery)")
+		}
+		return
+	}
```
rmrt1n: suggestion: this system should just assume the runtime is already setup by the init system, which must have already been run. if the init system crashes if it fails to set up the runtime, this system (and others that depend on it) doesn't have to handle such cases, resulting in simpler code.

pkg/plugin/physics2d/system/pipeline.go
rmrt1n, line 1: yeah not sure how this'll fit with user systems ATM. I'll pause the review here until we get a better idea

rmrt1n (changes requested):

pkg/plugin/physics2d/system/init.go
AnthonyWadham: So I did a deep dive into this to make sure.   
My pipeline file has:  


```
if rt.World == nil {
		if err := internal.FullRebuildFromECS(g, entries); err != nil {
			state.Logger().Error().Err(err).Msg("physics2d: FullRebuildFromECS failed (nil world recovery)")
		}
		return
	}
```

  
This should catch the case you are stating.  
  
I also looked into a full crash case and to my knowledge it reruns the register so everything like the config and runtime would be recreated?

pkg/plugin/physics2d/system/init.go
AnthonyWadham: do you think insted of panic here it should panic within fullrebuild so in pipline.go if world == nil and tries to rebuild and fails also would panic?  
  
I think if this ever fails we cannmot run game anymore correct?

pkg/plugin/physics2d/system/singleton.go
AnthonyWadham: This is used at the start of pipeline.go.  
  
its use case is if a user deletes this entity it will be recreated.    
Say they have a function to clean the entire shard and wipe all entities,  then this would never be recreated and break the game.

darkmavis (approved):
As was just disccussed, I'm fine with accepting that for a change with this level of complexity, we simply need to merge and test in the real-world.  I'd be shocked if we didn't encounter issues, but that's ok.  We'll find out what they are and fix them.

In some ways, this carries less risk than the initial decision to use an unofficial port of Box2D from C to Go, and for that, we had to real-world test and find out if it was stable and deterministic, and what to avoid if anything.  We did and it worked.  We can do the same here, but with more confidence that we're picking up the latest changes and nothing in the core library was lost in translation.

fwiw - I reviewed code with help from Claude.  The gameplay test footage is enough of a sanity check for me.  I'm not expecting perfection, and if we get it, great.

Also stoked for the performance and stability improvements provided by v3.  Thanks for thinking to do this and making it happen.

I'm fine with having this merged in once Ryan's requested changes are resolved.

pkg/plugin/physics2d/internal/writeback.go
```diff
@@ -0,0 +1,69 @@
...
+// their position. Since both BodyTypeKinematic and BodyTypeManual map to Box2D kinematic
+// bodies, the ECS body type is checked rather than the bridge body type.
+func WritebackFromStepResults(states []cbridge.BodyState, entries []WritebackEntry) {
+	rt := Runtime()
+	if !cbridge.WorldExists() {
+		return
+	}
+
+	// Build a lookup from entity ID to writeback entry for efficient matching.
+	entryMap := make(map[cardinal.EntityID]*WritebackEntry, len(entries))
```
winton-library, line 33: I notice that this map usage is correct, but I think we do not need to always reallocate it.  
  
My understanding is that every tick, writeback does:
entryByID := make(map[cardinal.EntityID]*WritebackEntry, len(entries))

In long term, Maps aren't free to allocate. For 1000 entities, that's roughly 80KB of heap allocation per tick. At 60Hz that's +-4.8MB/second of garbage for the GC to clean up. Since this is one of the hot path, you always want to fight for zero GC pressure.

Possible fix depends on whether if states and entries have the same element order in the slice. If yes you can: Iterate EntityID order, skip the map entirely.  
  
If not, then keep one map on PhysicsRuntime and clear() it between ticks (Go 1.21+ clear builtin is O(n) but doesn't re-allocate). Map memory gets reused.

winton-library: Again this is more like long term performance enhancer rather than bugs.

PR #887, opened 2026-04-13
feat: replace concurrent scheduler with sequential scheduler

rmrt1n:
## TL;DR

The PR title sums it up basically.

## Rationale

Entity operations (create, destroy, etc.) aren't concurrent safe and require locks to synchronize. The concurrent scheduler brings better performance when systems touch different components, but when it comes to entity operations, all systems effectively become sequential. Tested on Rampage, a sequential scheduler cuts down tick execution time by roughly half just from the lack of lock contention.


rmrt1n:
* **#887** <a href="https://app.graphite.com/github/pr/Argus-Labs/world-engine/887?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Argus-Labs/world-engine/887?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/Argus-Labs?ref=merge-queue-instructions-in-stack-comment&prId=887).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

pkg/cardinal/system.go
winton-library: I think this is no longer relevant, right? I think we can remove this.

pkg/cardinal/debug.go
winton-library, line 391: I was thinking about this ToProto function. I check the caller and it boils down to snapshot use and debug.

Since the snapshot path runs on the tick goroutine itself (after Tick() returns), there's no data-integrity issue there because nothing is mutating state while it serializes. But GetState runs on the HTTP handler goroutine with no lock. Doesn't that mean it can return an inconsistent state for the engineer's debugging session?

winton-library (commented):
I left some comments, I think it's only worth to take a look if the new architecture for cardinal is still a long way to go.

pkg/cardinal/debug.go
rmrt1n: yeah ur right, I'm reworking snapshots, will take a look at this again in the future. added to backlog for now.

PR #893, opened 2026-05-14
feat: add data plugin for JSON config loading with snapshot reconciliation

AnthonyWadham:
## TL;DR

Adds a `data` Cardinal plugin that loads JSON configuration into a process-local in-memory catalog at world init, with snapshot-aware reconciliation so a restarted world resumes on the exact config it was running before the crash.

## What changed?

A new `pkg/plugin/data` package, split into a thin facade plus two sub-packages (matching the `lobby` / `physics2d` convention):

**`Plugin`** — the Cardinal plugin facade. Callers do `data.Register`[`T`](plugin) for each config kind, then `cardinal.RegisterPlugin(world, plugin)`. The catalog loads eagerly during registration, so `data.Get[T]()` is valid from anywhere in the game immediately after — no plugin handle is threaded through systems; `Get` resolves against the registered plugin internally.

**`Definition`** **/** **`Resolver`** **/** **`Validator`** — the interfaces a config kind implements.

- `Definition` (required): `Name()` (catalog key) + `JSONFile()` (source-relative path).
- `Resolver` (optional): fetches additional files the kind's JSON references — e.g. Tiled `.tmj` tilemaps that `map_levels.json` points at. **Resolver fetches always go through the local embedded filesystem, never the operator**, because these are heavy designer-bundled build-time assets that don't fit a "edit a row in a web UI" workflow.
- `Validator` (optional): enforces post-load invariants (duplicate IDs, required fields); a non-nil error panics at boot.

**`Source`** **/** **`EmbedSource`** — the abstraction over where bytes come from. `Fetch(ctx, file, hash)` returns the bytes plus the content hash the source attaches to them. `EmbedSource` serves `go:embed` files and computes a sha256 of the bytes. `PickSource` selects the implementation by environment, with a hook point for a future `OperatorSource` when operator/forge integration ships.

**`State`** — owns the catalog, the per-file manifest (`jsonFile → content hash`), and the reconcile logic. `LoadAll` fetches and assembles every registered kind at startup. `Reconcile` runs every `PreUpdate` tick and covers the full lifecycle: fresh boot, steady-state restart (no-op fast path), snapshot-restored restart whose config differs — re-fetching each changed file at the snapshot's exact hash from a versioned source, or logging a warning and keeping current config when a single-version source (the embed) can't serve history — `cardinal.Reset()` debug reload, and restore from a pre-feature snapshot.

**`component.ConfigManifest`** — the only snapshot-resident artifact the plugin owns: a `map[string]string` of JSON file path → content hash for whatever config the world is currently running. The reconcile system reads it back after a snapshot restore to detect and correct drift. Static config itself no longer lives in snapshots at all.

## How to test?

```
go test ./pkg/plugin/data/...

```

All cases are in `pkg/plugin/data/data_test.go`, using an embedded `testdata/` fixture plus in-process fake `Source` implementations — no external dependencies. Coverage: eager load at `RegisterPlugin`; `ConfigManifest` creation on a fresh world; re-fetch of a historical version on snapshot restore; warn-and-keep-current when a single-version source can't serve the snapshot's hash; panic on versioned-source fetch failure; panic on `Register`\-time fetch error; duplicate `Name()`/`JSONFile()` registration panics; custom `UnmarshalJSON`; `Validator`; and a `Resolver` test that proves Resolver fetches go through the local embed and never touch the primary source.

## Why make this change?

Game shards previously loaded JSON config through ad-hoc per-kind loader systems, each writing onto its own Cardinal singleton component. This plugin centralizes that into one reusable abstraction that is:

- **Snapshot-lean** — static config (byte-identical across replicas, never mutated at runtime) no longer gets re-serialized into every snapshot. The only resident artifact is a few-KB hash manifest.
- **Snapshot-safe** — a crash and restart never silently changes the rules of an in-flight game. The `ConfigManifest` + reconcile system guarantee the catalog matches whatever version the snapshot was running, re-fetching historical versions by content hash when needed.
- **Dev/prod portable** — `Source` decouples the load path from the storage backend. Local dev uses the embedded FS; operator/forge integration flips automatically in `PickSource` with no change to any shard's `main.go`.
- **Fail-loud** — fetch errors, unmarshal errors, validation errors, and unresolvable snapshot hashes all panic immediately rather than producing a quietly-broken runtime.

### DIFF
--- a/pkg/plugin/data/system/state.go
+++ b/pkg/plugin/data/system/state.go
@@ -1,0 +1,223 @@
+package system
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"maps"
+	"slices"
+
+	"github.com/argus-labs/world-engine/pkg/cardinal"
+	"github.com/argus-labs/world-engine/pkg/plugin/data/component"
+	"github.com/goccy/go-json"
+	"github.com/rotisserie/eris"
+)
+
+// State owns the data plugin's per-instance load state: registered kinds, the in-memory catalog,
+// and the per-file manifest hashes currently reflected in that catalog.
+//
+// The plugin facade (data.Plugin) holds a pointer to a State and forwards every operation here.
+// Splitting it out keeps plugin.go a thin facade (matching the lobby / physics2d convention) and
+// keeps all the load-and-reconcile logic colocated with the components it operates on.
+type State struct {
+	loaders  map[string]kindLoader // jsonFile → loader
+	catalog  map[string]Definition // Name() → loaded value
+	manifest map[string]string     // jsonFile → hash currently reflected in catalog
+}
+
+// NewState builds an empty State. Call AddKind for each registered kind before LoadAll.
+func NewState() *State {
+	return &State{
+		loaders:  map[string]kindLoader{},
+		catalog:  map[string]Definition{},
+		manifest: map[string]string{},
+	}
+}
+
+// AssembleFunc materializes raw bytes into a fully-loaded Definition: json.Unmarshal → Resolve (if
+// implemented) → Validate (if implemented). Used both at initial load and at reconcile re-fetches
+// so a snapshot-pinned version goes through the same hooks as a fresh load.
+//
+// resolverSource is the Source the Resolver hook (if any) uses to fetch additional files. By
+// convention this is always the local embed (never the operator), because Resolver-fetched files
+// are heavy, designer-bundled, build-time assets that don't fit a "designer edits a row in a web
+// UI" workflow.
+type AssembleFunc func(ctx context.Context, resolverSource Source, raw []byte) (Definition, error)
+
+// kindLoader is one entry in State.loaders, keyed by jsonFile: enough to assemble bytes into a
+// loaded Definition and key the result by Name().
+type kindLoader struct {
+	name     string
+	assemble AssembleFunc
+}
+
+// MakeAssemble returns the standard assemble function for kind T: json.Unmarshal into a fresh T,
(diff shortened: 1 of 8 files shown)


AnthonyWadham:
* **#893** <a href="https://app.graphite.com/github/pr/Argus-Labs/world-engine/893?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Argus-Labs/world-engine/893?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/Argus-Labs?ref=merge-queue-instructions-in-stack-comment&prId=893).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

pkg/plugin/data/system/state.go
```diff
@@ -0,0 +1,223 @@
...
+		var def T
+		if err := json.Unmarshal(raw, &def); err != nil {
+			return nil, err
+		}
+		if r, ok := any(&def).(Resolver); ok {
+			if err := r.Resolve(ctx, resolverSource); err != nil {
+				return nil, err
+			}
+		}
+		if v, ok := any(def).(Validator); ok {
```
zulkhair: should it use (&def) like on line 63?

AnthonyWadham: yes good catch

zulkhair (approved):

PR #896, opened 2026-05-21
feat: handle client connections directly in cardinal

rmrt1n:
This is an intermediary PR that moves the gateway's client handling logic to Cardinal. At this state, we can't migrate the infra yet since we still need to figure out service discovery for clients and shards.

> [!NOTE]
> Cross-shard commands are still using the old NATS path until we figure out service discovery.

**For reviewers**:

The new behavior is in `service2.go`, just so it's easier to read. There's also a `cmd/client` to test this PR:

```bash
# start NATS
docker network create world-engine-network >/dev/null 2>&1 || true; docker rm -f world-engine-nats >/dev/null 2>&1 || true; docker run -d --name world-engine-nats --network world-engine-network --restart unless-stopped -p 4222:4222 -p 8222:8222 -p 6222:6222 -p 4443:4443 -v world-engine-nats:/data/jetstream nats:2.12.2 -js -sd /data/jetstream

# start cardinal, u can use any template
go run ./shards/game/main.go

# start the client (the instructions will be printed)
go run ./cmd/client/main.go
```


rmrt1n:
* **#896** <a href="https://app.graphite.com/github/pr/Argus-Labs/world-engine/896?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Argus-Labs/world-engine/896?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/Argus-Labs?ref=merge-queue-instructions-in-stack-comment&prId=896).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

pkg/cardinal/cardinal.go
```diff
@@ -150,7 +149,8 @@ func (w *World) StartGame() {
 	// Start the NATS connection and handler. Failures here panic; observers
 	// above are already running, so a goroutine/stack profile is reachable
 	// during the panic window via the deferred shutdown chain.
-	if err := w.service.init(); err != nil {
+	// Start the ConnectRPC client-facing service.
+	if err := w.service.init(":5000"); err != nil {
```
winton-library: Can we move this to other port, maybe :5050? :5000 collides with macOS Control Center's AirPlay Receiver. The airplay feature can be disabled if needed, but changing the port is friendlier for other macOS devs.

pkg/cardinal/service2.go
```diff
@@ -0,0 +1,854 @@
...
+		Addr:              address,
+		Handler:           h2c.NewHandler(mux, &http2.Server{}),
+		ReadHeaderTimeout: 10 * time.Second,
+	}
+
+	go func() {
+		if err := s.server.ListenAndServe(); err != nil && !eris.Is(err, http.ErrServerClosed) {
+			s.log.Error().Err(err).Msg("service2 server error")
+		}
+	}()
```
winton-library: This will fail silently if the port was in used. Maybe we can utilize something like this (before the go func) to check whether port can be used or not.

`
 listener, err := net.Listen("tcp", address)
    if err != nil {
        return eris.Wrapf(err, "failed to listen on %s", address)
    }
`

pkg/cardinal/service2.go
```diff
@@ -0,0 +1,854 @@
...
+		}
+
+		subscriber.mu.Lock()
+		err := subscriber.stream.Send(&cardinalv1.StartEventStreamResponse{
+			Address: s.world.address,
+			Event:   eventPb,
+		})
+		subscriber.mu.Unlock()
+		if err != nil {
+			return eris.Wrap(err, "failed to send event to subscriber")
```
winton-library: Since this is planned to become an actor model later, just noting the return behaviour here.

If one client consistently errors on Send, the loop returns here, so every subscriber ordered after it that tick (the order comes from the randomized s.subscribers map range above) misses the update, and since the order re-randomizes, it's a different subset of clients each tick until the bad one is evicted from said map.

proto/worldengine/cardinal/v1/cardinal.proto
```diff
@@ -0,0 +1,128 @@
...
+        pattern: "^[a-zA-Z0-9.*>_-]+$"
+      }
+    }
+  }];
+}
+
+// StartEventStreamRequest represents a request to establish a stream of events from Cardinal.
+// This enables real-time monitoring of system events through a server-side streaming connection.
+message StartEventStreamRequest {
+  // TODO: remove this.
```
winton-library: I think this comment needs more detail on why it will be removed.

winton-library (commented):
I left some comments. I think the port issue are the only major one (the 5000 port and how it fails silently).

winton-library (approved):

winton-library (changes requested):
The linter fails, I copied the wrong code example that doesn't use context.

winton-library (approved):
LGTM

PR #917, opened 2026-07-25
refactor: publish immutable state view each tick so GetState is race-free and always current

AnthonyWadham:
### TL;DR

Fixes a data race in `GetState` and eliminates redundant world serialization by unifying snapshot and debug state publishing into a single `persistState` path.

### What changed?

`GetState` previously called `ToProto` directly on live ECS state from an HTTP handler goroutine, racing with the tick goroutine. It now reads from an `atomic.Pointer[stateView]` that the tick goroutine publishes at each tick boundary.

The old `snapshot` function has been replaced by `persistState`, which serializes the world once per tick and fans the result out to both the durable snapshot store and the debug view. When debug is disabled and no snapshot is due, serialization is skipped entirely, preserving the production fast path.

`ComponentsBitmap` in archetypes aliases live bitmap memory via a zero-copy `ToBytes` call. Since the published view outlives the tick, `publishState` now clones each bitmap before storing it, preventing later tick mutations from corrupting the view seen by `GetState`.

`setPaused` now mirrors the pause flag onto the published view immediately, so `GetState` reflects pause/resume transitions without waiting for the next tick. An initial state view is published at world startup (before the first tick) and after reset, so `GetState` is never unavailable once the world is running.

`debugModule` is now a pointer type throughout, enabling nil-receiver methods (`publishState`, `captureState`, `setPaused`, etc.) that are no-ops when debug is disabled, removing scattered nil checks at call sites.

The unused `stepReady` channel has been removed from `tickControl`.

### How to test?

New internal tests cover the changed behavior directly:

- `TestGetStateUnavailableBeforeFirstPublish` — `GetState` returns `Unavailable` before the first publish.
- `TestPublishStateRoundtrip` — `GetState` serves exactly what `publishState` stored.
- `TestPublishStateClonesArchetypeBitmaps` — the published view does not alias live bitmap memory.
- `TestSetPausedUpdatesView` — `setPaused` updates both the control flag and the view atomically.
- `TestDebugStateNilSafety` — nil-receiver methods do not panic.
- `TestGetStateLifecycle` — drives the real `run()` loop through pause, step, reset, and resume, asserting `GetState` reflects each transition immediately.
- `TestPersistStateFeedsSnapshotAndView` — a single tick serialization feeds both the snapshot store and the debug view at the same height.
- `TestPersistStateProdFastPath` — with debug disabled, serialization only occurs on snapshot ticks.

### Why make this change?

The previous `GetState` implementation had an acknowledged data race (marked with a `TODO` comment) and called `ToProto` on every request, regardless of whether the world state had changed. Separately, snapshot ticks and debug state were produced by independent serialization calls, doubling the cost on ticks where both were needed. This change fixes the race, removes the duplicate work, and makes the debug view's staleness bounded to one tick.

### DIFF
--- a/pkg/cardinal/cardinal.go
+++ b/pkg/cardinal/cardinal.go
@@ -123,8 +123,7 @@ func NewWorld(opts WorldOptions) (*World, error) {
 
 	// Create the debug module only if debug is on.
 	if *options.Debug {
-		debug := newDebugModule(world)
-		world.debug = &debug
+		world.debug = newDebugModule(world)
 	}
 
 	// Create the pprof module only if pprof is on.
@@ -176,6 +175,9 @@ func (w *World) run(ctx context.Context) error {
 	if err := w.restore(ctx); err != nil {
 		return eris.Wrap(err, "failed to restore state from snapshot")
 	}
+
+	// Publish an initial view so GetState works before the first tick. Nil-safe: no-op when debug is off.
+	w.debug.captureState(w.currentTick.height, w.currentTick.timestamp)
 
 	logger := w.tel.GetLogger("shard")
 	logger.Info().Msg("starting core shard loop")
@@ -229,27 +231,41 @@ func (w *World) Tick(ctx context.Context, timestamp time.Time) {
 		w.tel.Logger.Warn().Err(err).Msg("errors encountered dispatching events")
 	}
 
-	// Publish snapshot.
-	if w.currentTick.height%uint64(w.options.SnapshotRate) == 0 {
-		snapshotCtx, cancel := context.WithTimeout(ctx, 2*time.Second)
-		w.snapshot(snapshotCtx, timestamp)
-		cancel()
-	}
+	// Publish state to snapshot and debug module.
+	w.persistState(ctx, timestamp)
 
 	// Increment tick height.
 	w.currentTick.height++
 }
 
-// snapshot persists the world state as a best-effort operation. Snapshots are best effort only, and
-// we just log errors instead of returning it, which would cause the world to stop and restart,
-// effectively losing unsaved state. If a snapshot fails, the main loop still continues and we retry
-// in the next snapshot call.
-func (w *World) snapshot(ctx context.Context, timestamp time.Time) {
-	worldState, err := w.world.ToProto()
-	if err != nil {
-		w.tel.Logger.Warn().Err(err).Msg("failed to serialize world for snapshot")
+// persistState serializes world state and does best effort to publish snapshot and debug state view.
+// We just log errors instead of returning it, which would cause the world to stop and restart,
+// effectively losing unsaved state. If a state serialization fails, the main loop still
+// continues and we retry in the next persistState call.
+func (w *World) persistState(ctx context.Context, timestamp time.Time) {
+	snapshotDue := w.currentTick.height%uint64(w.options.SnapshotRate) == 0
+	if !snapshotDue && w.debug == nil {
 		return
 	}
+
+	worldState, err := w.world.ToProto()
+	if err != nil {
+		w.tel.Logger.Warn().Err(err).Msg("failed to serialize world state")
+		return
+	}
+
+	w.debug.publishState(worldState, w.currentTick.height, timestamp)
+	if snapshotDue {
+		w.snapshot(ctx, timestamp, worldState)
+	}
+}
+
+// snapshot writes an already-serialized world state to storage, best-effort: errors are logged, not
+// returned, so a failed write doesn't stop the world and lose unsaved state — the next snapshot retries.
+func (w *World) snapshot(ctx context.Context, timestamp time.Time, worldState *cardinalv1.WorldState) {
+	ctx, cancel := context.WithTimeout(ctx, 2*time.Second)
+	defer cancel()
+
 	data, err := proto.MarshalOptions{Deterministic: true}.Marshal(worldState)
 	if err != nil {
 		w.tel.Logger.Warn().Err(err).Msg("failed to marshal world state to bytes")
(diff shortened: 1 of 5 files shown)


AnthonyWadham:
* **#917** <a href="https://app.graphite.com/github/pr/Argus-Labs/world-engine/917?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Argus-Labs/world-engine/917?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/Argus-Labs?ref=merge-queue-instructions-in-stack-comment&prId=917).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

pkg/cardinal/debug.go
```diff
@@ -267,22 +279,20 @@ func (d *debugModule) buildTypeSchemas(cache map[string]*structpb.Struct) []*car
 
 // tickControl coordinates pause, resume, step, and reset signaling for the tick loop.
 type tickControl struct {
-	pauseCh   chan chan uint64   // Request pause, receives tick height when paused
-	resumeCh  chan struct{}      // Signal to resume
-	stepCh    chan chan uint64   // Request step, receives tick height after step
-	resetCh   chan chan struct{} // Request reset
-	isPaused  atomic.Bool        // Current pause state
-	stepReady chan struct{}      // Signals that step result is ready to be read
```
AnthonyWadham, line 275: was found to be dead code so just removed

pkg/cardinal/cardinal.go
```diff
@@ -229,27 +231,41 @@ func (w *World) Tick(ctx context.Context, timestamp time.Time) {
...
+	w.debug.publishState(worldState, w.currentTick.height, timestamp)
+	if snapshotDue {
+		w.snapshot(ctx, timestamp, worldState)
+	}
+}
+
+// snapshot writes an already-serialized world state to storage, best-effort: errors are logged, not
+// returned, so a failed write doesn't stop the world and lose unsaved state — the next snapshot retries.
+func (w *World) snapshot(ctx context.Context, timestamp time.Time, worldState *cardinalv1.WorldState) {
+	ctx, cancel := context.WithTimeout(ctx, 2*time.Second)
```
AnthonyWadham, line 273: context cancel was duplicated in tick and shutdown so I merged. Highlightning incase there was a reason we specifically had todo this outside the snapshot function

rmrt1n (changes requested):
Instead of the world pushing the state to debug, just unconditionally serialize (proto) and store in the world. Snapshot and introspect should just read from the stored state. This means serializing is done just once per tick. 

I'd use a RWLock instead of the atomic pointer. I don't think your current code works as intended. The atomic guards access to the pointer itself, not the pointee, i.e. you atomically acquire the pointer while some other thread is writing to the pointee.

The final snapshot during shutdown should also use the pre-serialized state, not serialize right there, because shutdown may be called in the middle of a tick and the state may be half-written.

The rest of my comments are more nits/style guides:

pkg/cardinal/cardinal.go
```diff
@@ -177,6 +176,9 @@ func (w *World) run(ctx context.Context) error {
 		return eris.Wrap(err, "failed to restore state from snapshot")
 	}
 
+	// Sync debug state (perf + initial view) before the first tick. Nil-safe: no-op when debug is off.
+	w.debug.refreshState()
```
rmrt1n: We no longer need this if debug only pulls from world.

pkg/cardinal/debug.go
```diff
@@ -352,27 +362,65 @@ func (d *debugModule) Reset(
...
+//
+// ComponentsBitmap aliases live ECS memory (bitmap.ToBytes is zero-copy; see archetype.toProto).
+// ConnectRPC marshals this view off-tick while later ticks mutate that memory, so clone it before
+// publishing.
+func (d *debugModule) publishState(ws *cardinalv1.WorldState, height uint64, timestamp time.Time) {
+	if d == nil {
+		return
+	}
+	for _, arch := range ws.GetArchetypes() {
+		arch.ComponentsBitmap = slices.Clone(arch.GetComponentsBitmap())
```
rmrt1n: Since the serialized state is now used in more that one tick, we should copy in the `arch.toProto` instead of here.

pkg/cardinal/cardinal.go
```diff
@@ -229,27 +231,41 @@ func (w *World) Tick(ctx context.Context, timestamp time.Time) {
...
-// effectively losing unsaved state. If a snapshot fails, the main loop still continues and we retry
-// in the next snapshot call.
-func (w *World) snapshot(ctx context.Context, timestamp time.Time) {
+// persistState serializes world state and does best effort to publish snapshot and debug state view.
+// We just log errors instead of returning it, which would cause the world to stop and restart,
+// effectively losing unsaved state. If a state serialization fails, the main loop still
+// continues and we retry in the next persistState call.
+func (w *World) persistState(ctx context.Context, timestamp time.Time) {
+	snapshotDue := w.currentTick.height%uint64(w.options.SnapshotRate) == 0
+	if !snapshotDue && w.debug == nil {
```
rmrt1n, line 250: This code is no longer relevant, but cardinal doesnt' need to check if debug is nil since we already made the effort to make sure a nil `*debug` doesn't crash if we call its methods.

AnthonyWadham: The point of the nil debug check here is to make sure we are no serializing world on every tick in prod.  This statement is saying serialize if snapshot is due or in debug mode.  without this debug would only get state serialize on snapshots

pkg/cardinal/cardinal.go
AnthonyWadham: This is needed because its guarding trying to pull from the world before first tick runs.  If for instance you have a low tick rate and pause b4 first tick you will get a unavailable return error.  
  
Or if MCP starts a world and pauses instantly and tries to read state.  Its a niche bug but its protecting vs that bug.  We also need to reset performance regardless here so we aren't saving a line of code either way.

rmrt1n (approved):
lgtm. get rid of the tests

PR #922, opened 2026-07-29
feat(cardinal): add WatchSystemsTiming and ProfileSystems RPCs

ryanditjia:
## Summary

- Add `WatchSystemsTiming` for lightweight total system timing.
- Add `ProfileSystems` for per-system spans captured only while a profiler is connected.
- Keep profiling decisions fixed for each tick, so profiles are never partial.
- Let each RPC stream own its batching; the collector publishes completed ticks through bounded, latest-wins queues.
- Prevent pre-reset and post-reset ticks from appearing in the same batch.
- Keep `StreamPerf` as a deprecated compatibility adapter.
- Regenerate the Go, TypeScript, and C# clients.

## Stack

Built on #937.

## Verification

- `go test ./pkg/... -count=1`
- `go test -race ./pkg/cardinal/internal/performance -count=1`
- `task lint`
- `buf lint`
- `buf breaking --against '.git#branch=origin/main'`

> [!NOTE]
> ### Add `WatchSystemsTiming` and `ProfileSystems` RPCs to `DebugService`
> - Adds `WatchSystemsTiming` and `ProfileSystems` server-streaming RPCs to [debug.proto](https://github.com/Argus-Labs/world-engine/pull/922/files#diff-8e0932d0a304daec1332c310f9df7ebad7868e0069a158d70036066cc7341c50) for streaming per-tick timing and profiling data.
> - Reworks the performance [collector.go](https://github.com/Argus-Labs/world-engine/pull/922/files#diff-ed0bd1e94305f8d46e100e81defc88db79407feb2fcecde5c0c2b4f3072b9042) to publish per-tick instead of batching, tracking generations to drop pre-reset ticks, and using a fast path when idle.
> - System span recording in `registerSystem` is now conditional and only happens when a `ProfileSystems` subscriber is active, reducing overhead for timing-only streams.
> - Risk: `StreamPerf` is marked deprecated and now sources data from the new profile subscription. The collector API changed (`NewCollector` takes buffer size, `RecordTick` requires a `TickCapture`).
>
> 
>
> <details>
> <summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 1da9980. 5 files reviewed, 1 issue evaluated, 1 issue filtered, 0 comments posted</summary>
>
> ### 🗂️ Filtered Issues
> <details>
> <summary>pkg/cardinal/internal/performance/collector.go — 0 comments posted, 1 evaluated, 1 filtered</summary>
>
> - [line 44](https://github.com/Argus-Labs/world-engine/blob/1da998042604692b4aa37be93b4c47688924fb83/pkg/cardinal/internal/performance/collector.go#L44): The new `subscriberCount` fast path lets `StartTick` return before clearing `currentSpans` whenever no client is connected, while `RecordSpan` still appends every system execution unconditionally. Consequently, ordinary ticks with no profiling subscriber retain all spans indefinitely, growing `currentSpans` until the process exhausts memory. <b>[ Out of scope (triage) ]</b>
> </details>
>
>
> </details>
>


pkg/cardinal/cardinal.go
```diff
@@ -222,57 +217,50 @@ func (w *World) Tick(ctx context.Context, timestamp time.Time) {
...
 
 	// Tick ECS world.
+	var systemPhaseStart time.Time
+	if w.debug != nil {
+		systemPhaseStart = time.Now()
+	}
 	w.world.Tick()
 
-	w.debug.recordTick(w.currentTick.height, timestamp)
+	if w.debug != nil {
```
AnthonyWadham: we dont need to check there because the nil check happens on the debug function

pkg/cardinal/cardinal.go
```diff
@@ -222,57 +217,50 @@ func (w *World) Tick(ctx context.Context, timestamp time.Time) {
 	_ = w.commands.Drain()
 
 	w.currentTick.timestamp = timestamp
-	w.debug.startPerfTick()
+	w.currentTick.captureSystemSpans = w.debug.startSystemSpanCapture()
 
 	// Tick ECS world.
+	var systemPhaseStart time.Time
+	if w.debug != nil {
+		systemPhaseStart = time.Now()
```
AnthonyWadham: Can we just use timestamp?
I get that it would include the drain time but its very small amount of time and then we are not polluting so much of the tick with debug code?

Not 100% of this just leaving it for you to think about

pkg/cardinal/cardinal.go
ryanditjia: Addressed in [b8fe654](https://github.com/Argus-Labs/world-engine/commit/b8fe654a524654efb95f7c3cb0ff002aa4b8efaf). This avoids calling `time.Since(systemPhaseStartedAt)` before the nil-safe `recordTick` when `debug == nil` and `systemPhaseStartedAt` is zero-valued.

pkg/cardinal/cardinal.go
ryanditjia: I thought of this too, went back and forth until I settled to keep `systemPhaseStart`, and later renaming it to `systemPhaseStartedAt`.

Reason: it might seem harmless now (only drain time as the pollutant), but it would be dishonest later if, however unlikely, we add something more other than drain. I’d rather we keep it extremely honest: purely systems.

PR #941, opened 2026-08-14
refactor: extract world-free runtime services

winton-library:
## Decompose Cardinal so a world-free runtime can serve the same protocol

Everything a client can observe about the CardinalService protocol lived inside `pkg/cardinal`, welded to a world. ADR-057 needs a tickless shard to serve that same protocol, which would have meant reimplementing it — and every disagreement between the two copies would be a client-visible bug.

Four extractions, no behavior change for existing worlds:

- **`pkg/plugin/data`** — `Plugin.Load()`. Loading config never needed a world, only snapshot reconciliation did.
- **`pkg/auth`** — `User`, `Mode`, `NewMiddleware`, both authenticators.
- **`pkg/micro`** — `Service.ServeCommands` / `Client.SendCommand`. Inbound ISC was ~30 lines of validation ending in one line of runtime work.
- **`pkg/shard`** — the CardinalService handler, behind `CommandSink` + `EventBus`.`pkg/cardinal/service.go`: 759 → 248 lines. Cardinal's implementation of the seam is four
lines, and the same method now serves the ISC path, so client and inter-shard commands
converge on one implementation instead of two that happen to agree.

### Fixes a latent reply-waiter race

`SendCommandWithReply` registered its waiter *after* submitting the command. Harmless under a tick — the command waits in a queue, so the waiter always wins. A runtime that executes inside `Submit` publishes the reply before returning, so the event lands with nobody listening and the client blocks until timeout. The shared server registers first; regression test confirmed by restoring the old order (fails with `context deadline exceeded`).

### Behavior changes

- Reply-waiter ordering, above — closes a race, can't regress an existing world.
- ISC enqueue failure now reports `Internal` rather than `InvalidArgument`. Only fires for an unregistered command name, which is unreachable through a registered endpoint.

### Compatibility

`cardinal.User`, `AuthMode`, `ParseAuthMode`, `UserFromContext` remain as aliases — existing game code compiles unchanged. `newService` now returns an error (3 internal call sites updated).

---

### DIFF
--- a/pkg/auth/auth.go
+++ b/pkg/auth/auth.go
@@ -1,0 +1,217 @@
+// Package auth authenticates client requests to a shard's ConnectRPC service.
+//
+// Two modes ship: ARGUS validates a bearer JWT against the Argus Auth service's JWKS, and DEV
+// trusts an X-Email header. Both resolve a request to a *User, which downstream handlers read
+// with UserFromContext — a command's persona comes from there, never from the request body.
+//
+// Callers wire it in one step:
+//
+//	middleware, err := auth.NewMiddleware(auth.ModeDev, "")
+//	if err != nil {
+//	    return err
+//	}
+//	mux.Handle(path, middleware.Wrap(handler))
+//
+// The package holds no world or tick state, so any runtime serving the client-facing protocol can
+// use it, whether or not it has a simulation loop.
+package auth
+
+import (
+	"context"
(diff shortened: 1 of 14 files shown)


winton-library:
Closed due to the changes here will be applied locally into rampage-backend because we plan to have a new design for shards that doesn't use cardinal tick.

PR #947, opened 2026-08-31
refactor: snapshot format to use bytes, not serialized data

AnthonyWadham:
## Snapshots: encode straight to bytes

The snapshot path built a full protobuf message graph of the entire world every time it ran: one
message per entity, per component, per column, then `proto.Marshal` over the lot. This replaces that
with a two-pass encoder that writes the world directly from ECS memory into a single
exactly-sized buffer. No intermediate graph exists at any point.

Storage and the writers now carry opaque bytes rather than a message, so nothing downstream
re-serializes what the tick already produced.

### Cost

Storage write path, 1000 physics bodies:

| Path | Time | Allocations |
|---|---|---|
| Before (marshal, hand down, re-marshal) | 615 µs | 9,038 |
| Now | 108 µs | 1 |

The one allocation is the output buffer. At 5000 bodies it is 3,000 µs and 45,043 allocations
before, against 506 µs and 1 after.

### How it works

`StateWireSize` computes the exact encoded size, `AppendStateWire` writes exactly that many bytes.
Protobuf needs lengths before contents, so the size pass is unavoidable; it is pure arithmetic over
component fields, so nothing is cached between the two passes.

Components carry generated `SizeWire`/`AppendWire` methods, now part of `schema.Serializable`.
`MarshalWire` remains as a convenience but is derived from those two, so there is one encoder rather
than two that must agree byte for byte. The generator emits the pair for every message, nested types
included, because a parent needs a child's size before its bytes.

The output is byte-identical to `proto.Marshal` of the equivalent message, verified across every
shape the engine allows: scalars and named scalars, `immutable.Slice`, fixed arrays at any depth,
nested and mirrored messages, and timestamps. Decode is unchanged and still goes through protobuf.

### Format

`WorldState` now describes entities and a component name table instead of mirroring the runtime's
archetype storage. Archetypes and index tables are rebuilt on restore, so the file no longer depends
on internal layout. Entities ascend, the name table is in registration order, and restore resolves
components by name, so identical worlds produce identical bytes and old files survive component
changes.

Restore builds into a scratch state and commits only once the whole file is accepted, so a malformed
snapshot leaves the running world untouched. It bounds the free list a file may imply, since free ids
are derived from gaps and a few bytes could otherwise ask for billions.

### Also in here

- Entity id reuse is now order-independent. The free list is a min-heap, so a restored world hands
  out the same ids as one that kept running. It was FIFO by deletion order, which a snapshot cannot
  reconstruct.
- `SizeWire` rejects strings that are not valid UTF-8, which `proto.Marshal` used to catch. It runs
  over the whole world before any byte is written, so a bad value fails the write rather than the
  restore.
- The debug service stores the encoded bytes and decodes only when `GetState` is called. It was
  decoding the whole snapshot every tick, which handed back the entire allocation win whenever
  debug was enabled.
- physics2d, lobby, data and the three shipped templates are regenerated.

### Not in here

Old snapshots are not migrated. `WorldState` reuses field tags 2 and 3 for new types while the format
version stays at 1, which is safe only while nothing has persisted a snapshot yet. The proto records
this and says to bump the version before that stops being true.


pkg/cardinal/internal/ecs/archetype.go
```diff
@@ -24,18 +20,33 @@ type archetype struct {
 	entities   []EntityID       // List of entities of this archetype
 	columns    []abstractColumn // List of columns containing component data
 	compCount  int              // Number of component types in the archetype
+
+	// wireCIDs is columns' component IDs as a flat slice: wireCIDs[i] is the component held by
+	// columns[i]. The bitmap already stores this, but the snapshot encoder reads it per entity and
+	// a slice walk beats a bitmap iteration. Component IDs are also the snapshot's name-table
+	// indices (both follow registration order), so this is everything the encoder needs.
+	wireCIDs []ComponentID
```
rmrt1n: remove this, just use the components bitmap.

pkg/cardinal/internal/ecs/world_state_wire.go
rmrt1n, line 1: move this to world_state.go in its own section, cut down on the comments, use asd-ste100 or anything to write only necessary comments.

pkg/cardinal/internal/ecs/archetype.go
rmrt1n: remove this, just use the components bitmap

rmrt1n (approved):
some nits

PR #948, opened 2026-09-01
refactor: box2d fixes

AnthonyWadham:
# fix(physics2d): correct sleep-state tracking, contact point space, capsule diffs, and AABB overlap

## Summary

Five defects in the physics2d wrapper layer, found by a scenario suite run against the
pure-Go Box2D port, plus an allocation fix for the query path one of them touched.
All changes are in this plugin — `pkg/box2d` is untouched. In each case the engine
behaves per upstream Box2D v3 and the wrapper mistranslated it.

| # | Defect | Symptom | Fix |
|---|---|---|---|
| 1 | `PhysicsBody2D.Awake` never mirrored back | Snapshot restore froze a moving body mid-air (~57 m off) | Mirror solver sleep state after each step |
| 2 | Teleport didn't wake the body | Moving a body onto a sleeping one raised no contact events | Wake on game-written transform |
| 3 | `ContactEventPayload.Point` was body-relative | Contact points wrong for any body not at the origin | Use world-space `ClipPoint` |
| 4 | Capsule end centers absent from shape diffs | Resizing a capsule silently did nothing | Add to deep copy and both comparators |
| 5 | `OverlapAABB` was broad-phase only | False positives; shapes whose geometry misses the box | Route through `World.OverlapShape` |

Two commits: `e929bada` (the five fixes) and `466a6bf9` (query allocations).

## Sleep-state model (changes 1, 2 + two follow-ons)

`Awake` was write-only: pushed into Box2D at creation, never read back. The component
froze at its creation value while the solver's real state diverged. Because a rebuild
recreates bodies from components, restore resurrected bodies in the wrong sleep state.

Mirroring the value back exposed three latent bugs that the stale `true` had been masking.
The resulting policy — **an explicit `Awake` write always wins; implicit cases wake**:

- **After each step**: mirror solver sleep into the component, on change only. Disabled
  bodies are skipped (`IsBodyAwake` is false for them, but that's disablement, not sleep).
- **Game writes a transform**: wake. `b2Body_SetTransform` deliberately does not wake, and
  a sleeping body's contacts are never narrow-phased, so a teleport would otherwise raise
  no begin/end events. Skipped when the same tick explicitly wrote `Awake=false`.
- **Game changes a body param without touching `Awake`**: wake, so the change takes effect
  now rather than at the next impact. (Previously "worked" only via the stale `true`.)
- **Full rebuild** (restore / `Plugin.Reset`): wake bodies listed in the persisted
  `ActiveContacts` baseline. The post-rebuild diff can't distinguish "still touching but
  asleep" from "gone", so restored sleepers produced spurious `ContactEnd`. Bodies with no
  persisted contacts keep their value — a body spawned asleep stays asleep.

## Notes on the other three

**Contact point** was `Manifold.Points[0].AnchorA`, which is relative to the body's origin,
surfaced as a world coordinate. Fixed at **two** sites: the live path (`applyManifold`) and
the persisted-contact path (`collectBodyContacts`), which had the same bug.

**Capsule centers** were missing from `deepCopyColliderShape`, `colliderShapeDeepEqual`, and
`colliderShapeStructuralEqual` — so a capsule geometry change never reached Box2D.

**`OverlapAABB`** used `World.OverlapAABB`, which reports every fat-AABB broad-phase hit. The
documented contract is narrow-phase, so it now builds a proxy from the query box and calls
`World.OverlapShape`, which distance-tests each candidate.

## Query performance

Narrow-phasing costs real time — that is the price of not returning wrong answers. The
second commit pays some of it back by fixing how the result is built.

`BenchmarkOverlapAABB/Bodies_1000` (~49-hit query), `-benchtime=1s -count=3`, median:

| Version | ns/op | B/op | allocs/op |
|---|---|---|---|
| Before this PR (broad-phase, **incorrect results**) | 2,485 | 2,032 | 7 |
| After correctness fix alone (`e929bada`) | 6,701 | 2,176 | 8 |
| After allocation fix (`466a6bf9`, HEAD) | 6,350 | **896** | **1** |

Net versus the old behaviour: ~2.5x the time, but correct, and with **7x fewer allocations
and 56% less memory** per call.

Two things caused the allocation churn, both fixed:

- The result slice grew from zero capacity on every call — 7 reallocations for a 49-hit
  query, 11 for an 800-hit one. Hits now gather into a runtime-owned buffer (growth paid
  once, ever) and the caller-owned result is allocated once at the exact final size.
- `World.OverlapShape` parks the proxy pointer in the engine's query context, so a stack
  local escaped and cost an extra allocation per call. The proxy moved onto the Runtime,
  matching the scratch pattern already used for the callback contexts.

The win grows with result size. At 784 hits it is 161 us / 29,936 B / 11 allocs before,
135 us / 13,568 B / 1 alloc after — 16% faster, 55% less memory.

A high-water-mark capacity hint was tried first and rejected: it produced the same timings
but made every subsequent small query allocate at the largest size ever seen (13 KB for a
4-hit result after one 800-hit query). The gather buffer has no such cliff.

Raycast and CircleSweep are unchanged and remain zero-allocation — they return a single
closest hit, so they have no variable-length result to build.

## Testing

- `go test ./pkg/plugin/physics2d/...` passes.
- New `TestBodyFlag_TeleportWithExplicitSleepStaysAsleep` pins the explicit-sleep rule.
  Verified it fails without the guard (body falls 4.86 m instead of holding position).
- External scenario suite: **397/0**, up from 389/5.
- Crash-restore driver: **10/0** in both modes (with and without `Plugin.Reset`), up from 8/2.
- Determinism unchanged: identical digest across worker counts 0-16.
- Broadphase scaling verified healthy: holding result size fixed, 200x more bodies costs
  1.4x on overlap and nothing measurable on raycast.

## Review notes

- `shadow.go` shows ~15 lines of whitespace-only churn: `CapsuleCenter1` is longer than any
  existing field name, so gofmt re-pads the struct literal. `git diff -w` shows the 2 real lines.
- No `pkg/box2d` changes. The port itself reproduced the reference implementation correctly
  on everything these scenarios reach.


pkg/plugin/physics2d/internal/contact_listener.go
```diff
@@ -52,10 +52,29 @@ func (rt *Runtime) Step() {
...
+		return
+	}
+	for key := range rt.ActiveContacts {
+		if bodyID, ok := rt.Bodies[key.EntityA]; ok {
+			rt.World.SetBodyAwake(bodyID, true)
+		}
+		if bodyID, ok := rt.Bodies[key.EntityB]; ok {
+			rt.World.SetBodyAwake(bodyID, true)
+		}
+	}
```
winton-library, line 86: I have concern about this loops over a Go map. Map iteration in Go is unordered and explicitly not deterministic, nothing here defines what sequence we call SetBodyAwake in.

I had Claude build a test case around the restore path. It constructs a snapshot the way ECS would hand it to us, a ground plane plus 14 crates settled against each other with Awake: false, then runs the real restore, here's the report:
- FullRebuildFromECS rebuilds the bodies from those components, LoadActiveContactsFromComponent loads the persisted contact baseline into rt.ActiveContacts, and then rt.Step(), which is the call that runs wakePersistedContactEntities.
- It steps 900 ticks so the crates topple and form new contact pairs, then fingerprints every crate's final position into a string. That whole sequence runs 30 times with nothing changed between runs, and the fingerprints go into a map so we can count how many distinct worlds came out of 30 identical restores.
- There's a second arm that wakes the bodies in sorted EntityID order before Step, the map loop then finds everything already awake and does nothing, which is the proposed fix acting as the control.
- As-is it produces 5-6 distinct worlds. We want only 1.

I've attached the analysis + the unit test.

[physics2d-wake-order-restore.md](https://github.com/user-attachments/files/31761831/physics2d-wake-order-restore.md)

winton-library (commented):
I left a comment that needs confirmation from your side @AnthonyWadham. I assume the Scott's underlying branch is correct and only based the review on yours.

pkg/plugin/physics2d/internal/contact_listener.go
AnthonyWadham: good catch and agreed

winton-library (approved):

PR #958, opened 2026-09-10
refactor(physics2d): convert slices to immutable slice type

AnthonyWadham:
active contacts, chain points and shapes are now immutable slices
polygons are array size 8
regenerated sdk
fixed tests

quick to unblock snapshots but have a bigger PR refactor not ready so just did the small stuff here fast

---

---
## Summary by cubic
Converts physics2d slice fields to `immutable.Slice` so component copies handed out by `Ref.Get` can't write into the world's storage, and makes polygon vertices a fixed 8-slot array with a live count. This fixes snapshot determinism for in-place shape edits.

- `Shapes`, `ChainPoints`, and `ActiveContacts.Pairs` are now `immutable.Slice`; callers use `Len`, `At`, `All`, and `Values` instead of indexing, and mutate via `With`, `Append`, and `Sub` followed by `Set`.
- Polygon vertices live in a fixed `[MaxPolygonVertices]Vec2` array with `VertexCount`; build them with `WithVertices`, which reports over-8 polygons to `Validate` instead of truncating or panicking.
- `VertexCount` was added to the wire proto as field 18, so existing snapshots decode without renumbering.
- `ChainPoints` has no JSON form — JSON is for hand-written payloads, and snapshots go through protobuf.
- Removed the `polygon-too-many-vertices` e2e scenario because over-8 polygons now fail `Validate` and never reach Box2D.

<sup>Written for commit 9eae5b341fce06904eca4bc5d6f9e46c221770f3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Argus-Labs/world-engine/pull/958?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


pkg/plugin/physics2d/component/collider.go
```diff
@@ -42,25 +51,34 @@ const (
...
 //   - ShapeTypeStaticChain → ChainPoints (open polyline in local space)
 //   - ShapeTypeStaticChainLoop → ChainPoints (closed loop in local space)
 //   - ShapeTypeEdge → EdgeVertices (exactly 2 points in local space)
 //   - ShapeTypeCapsule → CapsuleCenter1, CapsuleCenter2, Radius (two semicircles connected by a rectangle)
+//
+// A polygon has a bound Box2D fixes at MaxPolygonVertices, so Vertices is a fixed array stored
+// inline and VertexCount says how many of its slots are live — the same shape box2d.Polygon uses.
+// A chain has no such bound, so ChainPoints is an immutable.Slice: it travels as the same repeated
+// field a []Vec2 would, but its backing array is unreachable, so the copy Ref.Get hands back cannot
+// be used to write into the world's own column.
```
winton-library: This isn't quite right: `immutable.Slice` hides the array from indexing, but its derivations write through to it. `pb.Shapes.With(0, sh)` changes the world's column before any `Set` runs (`shadow.go` in this PR says the same thing). Someone reading this comment could skip the `Set`.

```suggestion
// A chain has no such bound, so ChainPoints is an immutable.Slice: it travels as the same repeated
// field a []Vec2 would, and its backing array cannot be indexed or resliced. Its derivations (With,
// Filter, ...) still write through to the world's column, so always Set the component after one.
```

winton-library:
FYI, not a blocker. I'm not sure what the best move is here, so I'm leaving this for ack.

Since `Vertices` is now `[8]Vec2`, the generated encoder (`wire.gen.go`) writes all 8 vertex slots for every shape, including circles and boxes, which have no vertices. Each slot is a separate `ToProto()` allocation, so every snapshot does about 12 extra allocations per shape (8 slots plus slice growth).

Benchmarks, `e7c97c7a` against this PR (M2 Pro, n=6):

| Benchmark | time | bytes | allocs |
|---|---|---|---|
| SnapshotTick 1000 bodies, snapshot every tick | +36% | +52% | +60% |
| SnapshotTick 5000 bodies, snapshot every tick | +32% | +52% | +60% |
| SnapshotTick 1000 bodies, snapshot every 50 ticks | +10% | +52% | +58% |
| Step 1000 bodies (debug on, so serializes every tick) | +28% | +46% | +57% |

As far as I can tell, the physics2d change follows the rules (bounded list becomes array plus count). The cost comes from how `world sdk generate` encodes fixed arrays. One option is for the generator to skip trailing zero slots, which loses nothing because the decoder already zero-fills missing slots. Another is to accept the cost for now, since the bigger refactor is still coming.

winton-library (approved):
I pre-approve it so you can do comment changes and merge it. There is only one impact/improvement that I can think of but it is not for this PR to solve because of how we currently structure the sdkgen + physics2d.

PR #968, opened 2026-09-15
refactor(ecs): move entity, component, and event APIs onto World methods

smsunarto:
Moves the remaining ECS operations that take a World onto World methods, including component access, entity iteration, registration, and system events. Migrates Cardinal callers and the API examples together.

Validation: Cardinal tests and the full package build pass, including World isolation, entity lifecycle, and event-delivery tests.

---
## Summary by cubic
Moves the remaining ECS APIs that take a `*World` onto `World` methods, replacing package-level calls like `ecs.Create(world)` with `world.Create()`. This completes the migration with no runtime behavior changes.

- Updates Cardinal callers and internal ECS examples.
- Adds tests covering entity lifecycle, world isolation, and system events.

<sup>Written for commit 2af8a17768d46e4f2cf788aa154f6002916ce086. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Argus-Labs/world-engine/pull/968?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


smsunarto:
The Has behavior bug is fixed with regression coverage in stacked PR #976. The misleading cardinal.World.Get/Set/Remove references have been removed from Ref documentation, and the README now says to create an entity and attach a component.

The existing isolation assertion is intentional: the test first asserts w.Alive(eid) is true and then other.Alive(eid) is false. This would fail if Alive used shared entity state across worlds, so it is not vacuous.

